PR #30, opened 2013-12-09
Proof of concept: more terse syntax

fungusakafungus:
I am playing with an alternative syntax for templates.

I would like to implement following:
 - use python assignments to set names of resources/params/outputs:
`Ec2Instance = ec2.Instance(ImageId="some-ami")` instead of `Ec2Instance = ec2.Instance("Ec2Instance", ImageId="some-ami")`
 - if a resource/param is used somewhere in another resourse, make a Ref out of it:
`ec2.Instance(KeyName=KeyName)` instead of `ec2.Instance(KeyName=Ref(KeyName)`
 - write `resource.attribute` instead of `GetAtt(resource, 'attribute')`
 - use pythons `str.format` or `%` string operation instead of `Join` -> not yet implemented

The current syntax should still work.

The idea of resources without names is to construct the template out of `locals()` or `module.__dict__` and add everything suitable to the template.

The idea of implementation of lazy references is to generate strings containing _Ref() or _GetAtt() and the id of object being referenced, which get resolved after the resources have been added to the template.

What do you think of this approach? Is it useful? Too much magic? Conceptually impossible? A better way to implement this?

What is implemented:
 - EC2InstanceSampleNewSyntax.py produces the same output as EC2InstanceSample.py
 - shortcuts for Ref and GetAtt
 - introduced ro_attributes class constant, which lists read-only attributes, to check if GetAtt calls are valid.

What is missing/badly implemented:
 - resource references inside lists like `NetworkInterfaces=[eth0, eth1]` do not work (yet)
 - only Ec2::Instance has its read-only attributes listed
 - `str.format` instead of `Join` is not yet implemented
 - weird method/function names and placement
 - if a resource is bound to two variables in the same scope, bad things will happen. This is not yet checked.
 - name is not required anymore in BaseAWSObject, as I set it afterwards.

I've tried to implement some of this some time ago but got confused: https://github.com/fungusakafungus/python-cfn-templates. That thing had str.format to `Fn::Join` conversion working.

### DIFF
--- a/examples/EC2InstanceSampleNewSyntax.py
+++ b/examples/EC2InstanceSampleNewSyntax.py
@@ -1,0 +1,63 @@
+# Converted from EC2InstanceSample.template located at:
+# http://aws.amazon.com/cloudformation/aws-cloudformation-templates/
+
+from troposphere import Base64, FindInMap, Mapping
+from troposphere import Parameter, Output, Ref, Template
+import troposphere.ec2 as ec2
+
+
+template = Template()
+
+KeyName = Parameter(
+    Description="Name of an existing EC2 KeyPair to enable SSH "
+                "access to the instance",
+    Type="String",
+)
+
+RegionMap = Mapping({
+    "us-east-1": {"AMI": "ami-7f418316"},
+    "us-west-1": {"AMI": "ami-951945d0"},
+    "us-west-2": {"AMI": "ami-16fd7026"},
+    "eu-west-1": {"AMI": "ami-24506250"},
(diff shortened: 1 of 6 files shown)


phobologic:
@fungusakafungus this has been open forever (coming up on 3 years) and it doesn't look like there's been much activity lately.  I'm going to go ahead and close it - if you're still interested in the idea, please go ahead and re-open (there were some conflicts anyway, so those will need to be resolved).  Thanks for submitting this!

fungusakafungus:
:+1: for closing, was never gonna happen.

PR #70, opened 2014-03-29
Improved cfn2py

mbrossard:
Hi,

I improved cfn2py to the point where it can roundtrip the examples (except for the OpenStack examples). It also works with some private files that include AWS::CloudFormation::Init sections.

Though not perfect and not used every day, cfn2py was my gateway to using troposphere. I think these improvements might help the next future users to step onboard.

To test with the examples, I used the following script:
```bash
#!/bin/sh

if ! [ -d tmp ] ; then mkdir tmp ; fi

for i in examples/*.py ; do
    BASE=tmp/`basename $i .py`
    if python $i >$BASE.json1 2>&1 ; then
        echo $i;
        ./scripts/cfn2py $BASE.json1 > $BASE.py ;
        python $BASE.py > $BASE.json2 ;
        cmp $BASE.json1 $BASE.json2
    fi
done
```

Sincerely,
Mathias Brossard
PS: I haven't coded much in Python, so there is probably some room for improvement.

### DIFF
--- a/scripts/cfn2py
+++ b/scripts/cfn2py
@@ -165,7 +165,7 @@ def output_value(v):
     if isinstance(v, basestring):
         return '"%s"' % (v,)
     elif isinstance(v, bool):
-        return '"%s"' % (str(v).lower(),)
+        return '%s' % (str(v))
     elif isinstance(v, list):
         return "[" + ", ".join(map(output_value, v)) + "]"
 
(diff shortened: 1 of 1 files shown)


mbrossard:
I've rebased my commits on master.

phobologic:
Thanks @mbrossard for rebasing this and for the PR!

PR #136, opened 2014-10-20
Extensions to cfn tool

etuttle:
Hi, I made some extensions to the cfn tool to support my workflow with troposphere.

The only change to existing behavior is the --create operation now waits for the stack creation to complete.  It tails along the stack event log waiting for a completion event, printing out events along the way.  The exit code of the tool will be nonzero if the creation failed.

I also added --delete and --update operations that behave in a similar way.

I also made the tool automatically render .py templates into json before sending to AWS.

Finally, I added a --capability flag which is necessary to create stacks if they use IAM capabilities.

Is troposphere interested in any of this work?

Thanks for the cool project!

### DIFF
--- a/scripts/cfn
+++ b/scripts/cfn
@@ -195,6 +195,9 @@ if __name__ == "__main__":
                         help="tail event log")
     parser.add_argument("-u", "--update", help="Update an existing stack",
                         dest="update")
+    parser.add_argument("-w", "--wait", action="store_true",
+                        help="Wait for the operation to complete and return "
+                             "an appropriate error code")
     parser.add_argument("stack", nargs='?')
     values = parser.parse_args()
 
@@ -220,14 +223,15 @@ if __name__ == "__main__":
                           parameters=values.params,
                           capabilities=values.capabilities)
 
-        for event in events:
-            print_event(event)
-            if event.physical_resource_id == stack_id:
-                if event.resource_status in ['UPDATE_ROLLBACK_COMPLETE',
-                                             'UPDATE_ROLLBACK_FAILED']:
-                    sys.exit(1)
-                if event.resource_status == 'UPDATE_COMPLETE':
-                    break
+        if values.wait:
+            for event in events:
+                print_event(event)
+                if event.physical_resource_id == stack_id:
+                    if event.resource_status in ['UPDATE_ROLLBACK_COMPLETE',
+                                                 'UPDATE_ROLLBACK_FAILED']:
+                        sys.exit(1)
+                    if event.resource_status == 'UPDATE_COMPLETE':
+                        break
 
     if values.delete:
         stack_id = get_stack_id_if_exists(conn, values.stack)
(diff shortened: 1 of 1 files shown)


phobologic:
Hey @etuttle - we actually had someone ask about this change on the cloudtools-dev mailing list.  I realize this probably fell off your plate in all this time, but just wanted to check and see if you were working on this still or planned to?  Thanks again!

etuttle:
I'm not actively using it.  Moved on to other things and not even using AWS at the moment.  But let me take a look at how hard it will be to finish up...

phobologic:
NP Ethan - if it proves to be too much of a pain let me know and I'll see what I can do.  Thanks!

etuttle:
I rebased the existing commits and then added the --wait flag we discussed back in December.  Then force pushed.

Warning, this isn't very well tested!  I ran a `cfn --wait --capability iam --create <py file>` and it appeared to do the right thing, but that was the extent of my testing.

rcuza:
@phobologic, @etuttle,

I just spent a little time trying to work out the merge conflicts. It looks like the `cfn` script that is on master has a **capabilities** command line option like the version in this branch. It would be easy enough to have the changes in this branch squash the file on master.

But as I was getting my fingernails dirty learning the code, I realized there are a number of issues with the script in this branch and the one on master.  The version on master has some serious style issue in the `create_stack` function (e.g. line 36 in commit 3aad426). The version in this branch has at least three `pep8` errors. And, I have not even tried to use either version to create a real stack.

It looks like none of the tests touch this script. Which begs the question, should we have anything in this repo of this level of complexity that isn't tested? Also keep in mind that there are other projects, like [stacker](https://github.com/remind101/stacker), that seem to be solving the same problem. 

If the answer is yes, these scripts are useful, then I'll create a README to the directory to start the documentation ball rolling for the context of the scripts directory, which will in turn should lead to the next step in resolving this pull request.

jeanphix:
> serious style issue in the create_stack function (e.g. line 36 in commit 3aad426)

@rcuza Why?

rcuza:
@jeanphix - It was not consistent with other functions in the script in how it was laid out, but it looks like commit da7e7fc is consistent. 

I may of used the serious adjectives because I have a different preference when I align attributes for a function. In hind sight, my preference is not a good enough reason to seriously criticize the style. My apologies. To go into too much detail, 3aad426 commit aligned the closing parenthesis  with the 'd' in `def`and put the attributes at four spaces:

```
def create_stack(
    conn,
    stackname,
    template=None,
    url=None,
    params=None,
    capabilities=None,
):
```
I prefer to have the closing parenthesis be aligned with the attributes and the attributes aligned with the opening parenthesis:
```
def create_stack(conn,
                 stackname,
                 template=None,
                 url=None,
                 params=None,
                 capabilities=None,
                 ):
``` 

The most important thing is to be consistent throughout the script.

PR #173, opened 2015-01-03
Fixing UpdatePolicy validation

robzienert:
Fixes validation using UpdatePolicy introduced to 0.7.0, which had an API-breaking change, but didn't update validation. Also removed the root UpdatePolicy class so not to lead to confusion.

Without this, any use of the new UpdatePolicy will fail to serialize to JSON.

### DIFF
--- a/examples/Autoscaling.py
+++ b/examples/Autoscaling.py
@@ -1,9 +1,10 @@
 from troposphere import Base64, Join
-from troposphere import Parameter, Ref, Template, UpdatePolicy
+from troposphere import Parameter, Ref, Template
 from troposphere import cloudformation, autoscaling
 from troposphere.autoscaling import AutoScalingGroup, Tag
 from troposphere.autoscaling import LaunchConfiguration
 from troposphere.elasticloadbalancing import LoadBalancer
+from troposphere.policies import UpdatePolicy, AutoScalingRollingUpdate
 import troposphere.ec2 as ec2
 import troposphere.elasticloadbalancing as elb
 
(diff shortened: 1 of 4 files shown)


phobologic:
Thanks for this @robzienert - one thing I noticed is that this seems to remove validation of the old style of UpdatePolicies, which could break things for quite a few people.  We probably want to keep that in there as well for now.  That said, I may be missing something - since I changed jobs I haven't rolled out CloudFormation quite as heavily (yet, it's happening).

Adding @philtay  since I believe he made the original change.

robzienert:
That's correct; this PR does have a breaking change that would force people to update their templates. I went this route for two reasons:

1. The current UpdatePolicy class only supports a subset of the CFN UpdatePolicy attribute's capabilities. The new(er) `policies.UpdatePolicy` class supports the full breadth of the CFN attribute and is written in such a way to support future changes whereas the old UpdatePolicy cannot.
2. I didn't want to have two separate UpdatePolicy classes as that would only lead to confusion and more complex validation code since we would need to check what instance of UpdatePolicy someone is using. I think we should just have a single way to do things; the more flexible and complete of the two is the obvious choice.

I assumed making an API-breaking change would be acceptable due to the pre-1.0 version and the upgrade path is straightforward as well: https://github.com/cloudtools/troposphere/pull/173/files#diff-499e3b7a9d5227fa78ded8dcb7e1ee2aL228

markpeek:
I merged @robzienert code with master and added a full deprecation of the root UpdatePolicy class. This should alert people of the change and point them in the right direction.

robzienert:
Looks like this PR can be closed, then, if the commit is in gh-173 branch?

markpeek:
Thanks!

PR #176, opened 2015-01-05
Adding AWS::CloudFormation::Init ConfigSets support

robzienert:
Current CFN Init capabilities only allow for a single "config" property
on the `cloudformation.Init` class. This PR adds support for configSets
according to:

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-init.html

Took some of what @mbrossard had in #70 and ran with it.


markpeek:
Thanks!

PR #178, opened 2015-01-07
GH-159: Changing CloudWatch Alarm's Threshold prop to be basestring

### LINKED ISSUE #159: Cloudwatch Alarm Threshold Type
The parameter type 'Threshold' within Cloudwatch Alarms is currently of type 'integer' whereas the AWS documentations notes this should be a String.

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cw-alarm.html#cfn-cloudwatch-alarms-threshold

I am hitting an issue when using alarms to check instance health - to monitor StatusCheckFailed I have an implementation which sets Threshold to 0.5 to evaluate a healthcheck of sorts. This works in Cloudformation but fails when I try to use it in the troposphere code.

I think the line 'Threshold': (integer, True), should be 'Threshold': (basestring, True), within cloudwatch.py

Any thoughts?

robzienert:
Closes #159

### DIFF
--- a/troposphere/cloudwatch.py
+++ b/troposphere/cloudwatch.py
@@ -4,7 +4,7 @@
 # See LICENSE file for full license.
 
 from . import AWSObject, AWSProperty, Ref
-from .validators import integer, positive_integer, boolean
+from .validators import positive_integer, boolean
 
 
 class MetricDimension(AWSProperty):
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks!

PR #179, opened 2015-01-15
Support AWS helper functions in lists during validation

mbrossard:
Hi,

I found some places were a list of basestring is expected but where a Ref function call would be used (e.g. the buckets value in an AuthenticationBlock). This is similar to what we do a few line above for a non-list type.

Sincerely,
--
Mathias


phobologic:
Great, thanks @mbrossard

PR #180, opened 2015-01-16
Remove properties from elasticbeanstalk.Application

ikben:
Both 'ApplicationVersions' and 'ConfigurationTemplates' are missing from the
current [UserGuide](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-beanstalk.html).

I've removed them from the code, to match this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## pending
+## 1.7.0 (2016-07-07)
 - Convert fake AWSHelperFns into AWSProperties (#478)
 - cfn script: allow update (#484)
 - Validate the template against AWS before stack creation (#485)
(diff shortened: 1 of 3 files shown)


ikben:
I just realised that troposphere follows a slightly different approach then what the user guide suggest.

I'll wait on some input before reworking elasticbeanstalk.py (if needed)

markpeek:
@cdandekar perhaps you could help us with this one. This is an example where Troposphere was written against one version of the users guide and now it appears the AWS::ElasticBeanstalk::Application type doesn't have ApplicationVersions or ConfigurationTemplates properties. Are those still valid properties or are they deprecated? Thanks.

cdandekar:
Those properties are still supported for backward compatibility. However, while authoring new templates, one should use the new ApplicationVersion and ConfigurationTemplate resource types:

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-beanstalk-version.html
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-beanstalk-configurationtemplate.html

todds02:
From experimenting, it looks like CloudFormation will reject a stack update if the ApplicationVersions property (old format) has an element removed.  You get an error along the lines of "you cannot modify ApplicationVersions".  I don't recall if it complained when adding a new ApplicationVersion or not.

Using the format in the AWS docs, Cloudformation happily adds the new ApplicationVersions, upgrades the beanstalk apps, and deletes the unused ApplicationVersions automatically.

I wouldn't consider this bug benign - at a minimum you need external scripts to remove old ApplicationVersion entries from each Beanstalk application.

PR #183, opened 2015-01-23
Update README.md

bbl232:
Readme needed a bit of work in my humble opinion


markpeek:
Thanks! I was recently thinking some additional docs would be good as well. Let me know if you have any thoughts on what you'd like to see.

PR #184, opened 2015-01-26
allow AWS helper functions for elasticache VPCSecurityGroupId

jstrunk:
I wanted to use Fn:Select for this property like I did with RDS.


markpeek:
I believe this was solved more generally with #179. Please reopen if this doesn't work for you on master. Thanks.

PR #185, opened 2015-01-27
DynamoDB Table Names Are Too Restrictive

RonRothman:
According to the docs, valid characters for DynamoDB table names include underscore, hyphen, and period.

http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html

But Troposphere v0.7.1 restricts table names to alphanumeric only:

```
  File "dynamodb.py", line 124, in print_tables
    table = Table(table_name)
  File "/usr/local/lib/python2.7/site-packages/troposphere/__init__.py", line 43, in __init__
    raise ValueError('Name "%s" not alphanumeric' % title)
ValueError: Name "my_table" not alphanumeric
```

Troposphere is preventing access (not just creation) to any table with a hyphen, underscore or period, all of which are valid.

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -40,7 +40,7 @@ class BaseAWSObject(object):
 
         # unset/None is also legal
         if title and not self.valid_names.match(title):
-            raise ValueError('Name "%s" not alphanumeric' % title)
+            raise ValueError('Name "%s" is not valid ()' % (title, self.valid_names.pattern))
 
         # Create the list of properties set on this object by the user
         self.properties = {}
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -25,10 +25,10 @@ AWS_REGION = 'AWS::Region'
 AWS_STACK_ID = 'AWS::StackId'
 AWS_STACK_NAME = 'AWS::StackName'
 
-valid_names = re.compile(r'^[a-zA-Z0-9]+$')
-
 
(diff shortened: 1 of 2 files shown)


markpeek:
I don't believe there is a bug here. The error that is being flagged is the name specified here as "Logical ID" and has the stated requirement "The logical ID must be alphanumeric (A-Za-z0-9) and unique within the template":
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resources-section-structure.html

I believe what you are wanting is to specify it this way where the Logical ID is "myDynamoDBTable" and the DynamoDB Table name is "my_table". There is an enhancement which could be considered an issue where the Table class could add a validation to ensure TableName is restricted to "a-zA-Z0-9_-." and meets the name length requirements.

```
myDynamoDB = t.add_resource(Table(
    "myDynamoDBTable",
    TableName="my_table",
    AttributeDefinitions=[
        AttributeDefinition(Ref(hashkeyname), Ref(hashkeytype)),
    ],
    KeySchema=[
        Key(Ref(hashkeyname), "HASH")
    ],
    ProvisionedThroughput=ProvisionedThroughput(
        Ref(readunits),
        Ref(writeunits)
    )
))
```

RonRothman:
Ah, I see--thanks for the explanation, much appreciated.

markpeek:
No problem. Happy coding!

PR #188, opened 2015-02-11
Added to the constructor of the Projection class the optional argument '...

adrianmajo:
This change allows troposphere to create Secondary Global Indexes adding the NonKeyAttributes field in the Projection attribute. At this moment, the library only provides a way to define the ProjectionType.


markpeek:
Thanks!

PR #189, opened 2015-02-13
Change AutoScaling ScheduledAction StartTime and EndTime to False

evanstachowiak:
StartTime and EndTime are not required Parameters according to AWS
Documentation:

http://amzn.to/1uPqyqd


iblazevic:
True, btw you can also make recurrence as optional while at it

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-as-scheduledaction.html#cfn-as-scheduledaction-recurrence

markpeek:
Thanks!

PR #192, opened 2015-02-17
attribute definitions required

iblazevic:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dynamodb-table.html#cfn-dynamodb-table-attributedef


markpeek:
Thanks!

Issue #199, opened 2015-03-04
elb sample


BardiaAfshin:
In the elb sample, if you attempt to create a stack from the generated cloud formation script you will get an error because 'us-east' is not defined. This needs to be changed to

```python
def AddAMI(template):
    template.add_mapping("RegionMap", {
        "us-west-1": {"AMI": "ami-3bcc9e7e"},
        "us-east-1": {"AMI": "ami-3bcc9e7e"}
    })
```

phobologic:
Fortunately @k-ong helped us out with this one!  The latest source should handle this for you - as well as a ton of other regions.

PR #203, opened 2015-03-06
Updating AMI IDs for other Regions. Adding CFN template metadata.

k-ong:
Updating issue: https://github.com/cloudtools/troposphere/issues/199


phobologic:
Awesome - thanks @k-ong!

PR #205, opened 2015-03-10
Cloudformation expects a string as a FilterPatter

urtens:
Cloudformation expects a string as a FilterPatter

### DIFF
--- a/troposphere/logs.py
+++ b/troposphere/logs.py
@@ -22,7 +22,7 @@ class MetricFilter(AWSObject):
     resource_type = "AWS::Logs::MetricFilter"
 
     props = {
-        'FilterPattern': ([basestring], True),
+        'FilterPattern': (basestring, True),
         'LogGroupName': (basestring, True),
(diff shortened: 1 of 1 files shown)


markpeek:
Will have to investigate as the docs say "A list of strings".

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-metricfilter.html

urtens:
You're right. It might be an outdated documentation though. CloudFormation spits out an error if you pass a list of strings:
"AWS::Logs::MetricFilter" "Value of property FilterPattern must be of type String"

jorgebastida:
Hitting this issue too. It is true that the documentation say that ``FilterPattern`` but their examples http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/quickref-cloudwatchlogs.html use strings and not lists of strings.

Tested this PR and CloudFormation is happy with the output. :+1:

markpeek:
Merged. Thank you @urtens for the fix and @jorgebastida for the confirmation.

PR #209, opened 2015-03-16
Elasticache VpcSecurityGroupIds should accept GetAtt

benjdlambert:
According to the latest docs for `Elasticache` the `VpcSecurityGroupIds` is `GetAtt` from the `SecurityGroup` not `Ref`.

I've left `Ref` as not to break people that are using it.

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticache-cache-cluster.html#cfn-elasticache-cachecluster-vpcsecuritygroupids

### DIFF
--- a/troposphere/elasticache.py
+++ b/troposphere/elasticache.py
@@ -3,7 +3,7 @@
 #
 # See LICENSE file for full license.
 
-from . import AWSObject, Ref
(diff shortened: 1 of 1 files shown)


phobologic:
Awesome, thanks!

Issue #210, opened 2015-03-18
ElasticBeanstalk ConfigurationTemplate properties are incorrect


mache:
ref: https://github.com/cloudtools/troposphere/blob/master/troposphere/elasticbeanstalk.py#L38-L46

According to the CF docs, there are some properties missing from troposphere for this resource.
 
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-beanstalk-configurationtemplate.html

The `TemplateName` is optional if you would like to inherit from another config template, however it's incorrectly required as a property in troposphere. Additionally, the `ApplicationName` must not be empty.

phobologic:
Thanks for pointing this out - either we were *way* off when we created ElasticBeanstalk, or else the template objects have changed drastically.

I'm working on an update.

phobologic:
BTW, I have a fix on deck in the branch: https://github.com/cloudtools/troposphere/tree/GH-210

That said, it looks like they got rid of the example template that we used for our examples & tests, so I'll need to convert one of the new ones (the nodejs one?) to this new format.  I'll try to get to that by this weekend.  If someone else wants to take a crack at it, or check out my branch to test it manually, that'd be awesome.

Thanks again @mache

mache:
Great, thanks for the quick response.

phobologic:
Ok, I've gone ahead and made a pull request that I believe fixes this and updates the example app/test.

https://github.com/cloudtools/troposphere/pull/213

@mache Any chance you could test the new classes, etc?  I don't use ElasticBeanstalk, so haven't played with it before.

phobologic:
Had a bug due to a misunderstanding of the different Policy types in the IAM module.  Updated the pull with the new version - also tested that that version launches in CloudFormation.  Looks good so far.

phobologic:
@mache Ok, we've merged GH-213.  Should be working as documented!

PR #213, opened 2015-03-21
Gh 210

phobologic:
Fixes GH-210


phobologic:
The CloudFormation template launched correctly, and when I went to the ELB URL in the Output I got:

https://www.dropbox.com/s/l0ee5wfx3rvqplf/Screenshot%202015-03-23%2010.15.25.png?dl=0

So this looks good.

PR #215, opened 2015-03-23
Readme in rst

phobologic:
Mostly this is to clean up the ugliness that results in using markdown on pypi.


markpeek:
+1 Thanks!

PR #220, opened 2015-04-12
Missing OriginPath in cloudfront.py

ltagliamonte:
Origin object is missing OriginPath attribute, as described here: http://docs.aws.amazon.com/AmazonCloudFront/latest/APIReference/DistributionConfigDatatype.html

### DIFF
--- a/troposphere/cloudfront.py
+++ b/troposphere/cloudfront.py
@@ -64,7 +64,7 @@ class CustomOrigin(AWSProperty):
 class Origin(AWSProperty):
     props = {
         'DomainName': (basestring, True),
-        'OriginPath' : (basestring, False),
+        'OriginPath': (basestring, False),
         'Id': (basestring, True),
         'S3OriginConfig': (S3Origin, False),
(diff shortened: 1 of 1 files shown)


markpeek:
Thank you for the fix. I'm wondering if this change should be made as well to ensure that list are only class Origin:

```
diff --git a/troposphere/cloudfront.py b/troposphere/cloudfront.py
index 2b6a93b..35e0932 100644
--- a/troposphere/cloudfront.py
+++ b/troposphere/cloudfront.py
@@ -120,7 +120,7 @@ class DistributionConfig(AWSProperty):
         'DefaultRootObject': (basestring, False),
         'Enabled': (boolean, True),
         'Logging': (Logging, False),
-        'Origins': (list, True),
+        'Origins': ([Origin], True),
         'PriceClass': (basestring, False),
         'Restrictions': (Restrictions, False),
         'ViewerCertificate': (ViewerCertificate, False),
```

ltagliamonte:
@markpeek i think that you are right.

PR #225, opened 2015-04-17
Some new properties (Apr 16, 2015 release)

philtay:


markpeek:
Thank you for getting these implemented. I did implement the HostedZoneTags using the base Tags class in an previous commit.

PR #228, opened 2015-04-20
Python 3.4 support

philtay:


philtay:
http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

markpeek:
Thanks for the patch and for including the faster builds in containers.

PR #230, opened 2015-04-22
SNS Topic fields are not required

pas256:
Believe it or not, none of the SNS Topic fields are required. If nothing ...is specified, CF will generate a new based on the name of the stack.

This PR gets the definition matching the Doc:

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sns-topic.html


phobologic:
Crazy!  Thanks @pas256

PR #234, opened 2015-04-27
autoscaling & cloudfront properties

philtay:


troposphere/autoscaling.py
```diff
@@ -56,7 +56,7 @@ def JSONrepr(self):
         return self.tags
 
 
-class NotificationConfiguration(AWSProperty):
```
markpeek: Thank you for implementing these changes. I was wondering if NotificationConfiguration still worked and whether to leave it as a backwards compatible interface. Thoughts?

philtay:
Dunno if it still works or not. For sure is not officially supported anymore. If the next troposphere release will be 0.8, I suggest to simply remove it as I did. In any case it isn't a widely used property or a particularly important one.

phobologic:
Thanks @philtay - planning on cutting a new release today, so this should be in there.

PR #236, opened 2015-05-01
Make AvailabilityZones parameter optional in AutoScalingGroup

cdodd:
Currently, when creating an `AutoScalingGroup` object, the `AvailabilityZones` parameter is required. However, the [AWS documentation](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-group.html#cfn-as-group-availabilityzones) states: `Required: Conditional. If you don't specify the VPCZoneIdentifier property, you must specify this property.`

This change makes the `AvailabilityZones` parameter optional.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Add properties CharacterSetName, KmsKeyId, and StorageEncrypted to AWS::RDS::DBInstance [GH-224]
 - Add Route53 HostedZoneVPCs, HostedZoneTags, HealthCheckTags
 - Add new properties from 2015-04-16 CloudFormation release [GH-225]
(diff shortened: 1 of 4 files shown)


phobologic:
Hey @cdood, thanks for this.  What do you think about adding the check that either AvailabilityZones or VPCZoneIdentifier are specified to the validate() method, that way we can catch if someone forgets both of them early on?

cdodd:
@phobologic, I've updated the `validate` method to check at least one of `AvailabilityZones` or `VPCZoneIdentifier` is defined.

I also had to update some of the tests to get them to pass. I'm slightly confused how these passed previously as `AvailabilityZones` was a required property but was not defined in the test cases.

phobologic:
Awesome - thanks for the pull request @cdodd !

PR #237, opened 2015-05-04
examples fix: users is list

alexlo03:
this brings the example in line with
https://s3.amazonaws.com/cloudformation-templates-us-east-1/IAM_Users_Groups_and_Policies.template

The current example CFN template will fail if you try to create-stack (I modified it for my use and found this bug).  

FYI I wrote a test harness to validate the example templates against the AWS CLI, but the validate-template does not catch this bug.  Let me know if you're interested in having this test harness, unfortunately it introduces a need for the AWS CLI and real IAM credentials that has `cloudformation:ValidateTemplate` rights, so it might be too heavy to be useful.

### DIFF
--- a/examples/IAM_Users_Groups_and_Policies.py
+++ b/examples/IAM_Users_Groups_and_Policies.py
@@ -27,13 +27,13 @@ cfnkeys = t.add_resource(AccessKey(
 users = t.add_resource(UserToGroupAddition(
     "Users",
     GroupName=Ref(cfnusergroup),
-    Users=Ref(cfnuser),
+    Users=[Ref(cfnuser)],
 ))
 
 admins = t.add_resource(UserToGroupAddition(
     "Admins",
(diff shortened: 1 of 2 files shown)


phobologic:
Thanks Alex!  Not sure we have the setup in place for troposphere to have it connect to AWS to do the validation.  If we made it a part of regular testing I'd be worried that changes that people would submit would be blocked whenever AWS changed their CF templates, and that would probably be bad.  Might be something we should check out for regular testing though, just to see if we can catch the changes a bit better.

Anyway, merged this change - thanks!

alexlo03:
@phobologic Do we think AWS is going to change their CF templates very often?  Or you mean the example templates?  This test is based on the tests/example_output that is already versioned in the troposphere repo (the regression set).  I have it up here if you want to check it out: https://github.com/cloudtools/troposphere/compare/master...alexlo03:alexlo03/test_templates

Issue #241, opened 2015-05-12
Broken support for Conditions


amos-stan:
Hi,

There are multiple issues revolving around conditions. I don't know whether you'll end up addressing them in the same code change or not so I'm reporting them in one go here:

1. cfn2py completely ignores conditions - the "Conditions" section of the source json as well as the "Condition" resources attributes are not set in the output .py file.
1. Once I added conditions manually to the generated script, the "Conditions" section output the Condition in the wrong format, i.e.:
```json
    "Conditions": {
        "EnableAutoScale": {
            "Condition": {
                "Fn::Equals": [
                    {
                        "Ref": "EnableAutoScaleFlag"
                    },
                    "true"
                ]
            }
        }
    }
```
instead of:
```json
    "Conditions" : {
        "EnableAutoScale" : {
            "Fn::Equals" : [
                {
                    "Ref": "EnableAutoScaleFlag"
                },
                "true"
            ]
        },
    }
```
(i.e. only the "Condition.condition" field should be rendered instead of the entire object)
On the other hand, as far as I can tell, the "Condition" attribute is printed correctly (i.e. only the condition name) in the Resources once I added their definitions.

This is with version 1.0.0

Thanks.

phobologic:
RE: The first point, unfortunately the cfn2py script is a bit of a second class citizen, and doesn't get much love, mostly because it's kind of a nightmare to deal with.  It's basically always been a 'best effort' tool.  If you would like to submit pull requests though, we'd be happy to take them :)

I think the second point is a misunderstanding of what the Condition resource is used for.  You don't actually use the Condition resource in the Conditions section for the most part (this isn't always true, more on that below).  The reason is that Condition is an attribute of the Condition itself.  It's confusing to explain, and easier to see in action - which you can do here:

https://github.com/remind101/stacker/blob/master/stacker/blueprints/asg.py#L62

Let me know if that doesn't make any sense :)

phobologic:
BTW, this helps explain it a little bit more:

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-conditions.html

amos-stan:
Thanks for your response.

I tried to look into cfn2py to add condition support but didn't manage to get my head around how it works (yet, perhaps I'll get back to it)
I just noticed that my change breaks the test_ec2Conditions test, which has a validating format.
Still - the output without my change generates invalid format for conditions - what's the right fix?

phobologic:
[...]

PR #242, opened 2015-05-13
GH-241: remove one level of container from Condition storage

amos-stan:
This change makes the "Conditions" section have the right content.
With this fix, the generated output passes cloudformation validation.


phobologic:
See GH-241 for discussion on this.

PR #243, opened 2015-05-13
added AWS::CloudFormation::Stack NotificationARNs property, see http:…

sharabash:
…//docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-stack.html#cfn-cloudformation-stack-notificationarns for reference


markpeek:
Thanks!

PR #249, opened 2015-05-18
add true s3 bucket name validator

ryno75:
Added a validator for s3 bucket names to get past the basestring (alphanumeric) limitation.  Used a regex to handle that.
Also... added a validator test for the new s3_bucket_name validator.

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -59,7 +59,11 @@ class TestValidators(unittest.TestCase):
         network_port(Ref(p))
 
     def test_s3_bucket_name(self):
-        s3_bucket_name('wick3d-sweet.bucket')
+        for b in ['a'*3, 'a'*63, 'wick3d-sweet.bucket']:
+            s3_bucket_name(b)
+        for b in ['a'*2, 'a'*64, 'invalid_bucket']:
+            with self.assertRaises(ValueError):
+                s3_bucket_name(b)
 
 if __name__ == '__main__':
     unittest.main()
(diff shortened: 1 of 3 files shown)


markpeek:
Thank you for implementing this patch. I noticed your regex allows for 255 character "non-dns-compliant" bucket names. Should this validator really enforce the DNS compliant naming?
http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html

Also, I saw you removed a negative test. I think what you want is:
```
with self.assertRaises(ValueError):
    s3_bucket_name('!invalid_bucket')
```

ryno75:
Oh, yeah... nice catch on the DNS compliancy... I forgot to change that from the standard bucket name length to the DNS compliant length.  While the cloudformation docs aren't clear, I wouldn't be surprised if the stack create choked on bucket names longer than 63 chars too.

Yeah, I think you are right on the negative test.  I didn't even consider using the double-neg like that.

Shall I implement these changes and re-create the pull request?

markpeek:
That would be great if you could commit and push those changes (it should automatically update this pull request). For extra credit on testing you could add things like this to validate the bounds checking:

```
s3_bucket_name('a' * 63)
s3_bucket_name('a' * 3)
with self.assertRaises(ValueError):
    s3_bucket_name('aa')
with self.assertRaises(ValueError):
    s3_bucket_name('a' * 64)
```

ryno75:
Ok. Challenge accepted. ;-)  Have a look.

markpeek:
Looks great. Thanks!

PR #251, opened 2015-05-20
Replace strict `int` comparison by flexible `troposphere.validators.integer`

hltbra:
I had issues in the past with troposhere and its strict type comparisons, and it hit me one more time. I passed a `long` to `rds.DBInstance.props['Iops']` and it errored out. I would assume anything that looks like an integer should have worked. That applies to all types.

This pull request adds a test to make sure troposphere *never* uses `int` when declaring properties and this pull request also replaces all `int` definitions by `tropospher.validators.integer`.

An improvement after this PR would be to review all types to make them more strict using `positive_integer` where necessary.


phobologic:
So one potential issue here is that things that would normally only accept an int can now, using integer, accept a float as well.  That said, I'm not sure if/where that could end up being a problem.  Maybe the integer() validator (or some new validator?) should be updated to be a little more strict in this regard?

Not sure what the best thing to do here is though, and this seems like heading in the right direction for the most part.

markpeek:
Thank you for this fix and for providing tests for it.

PR #252, opened 2015-05-26
Add validation for AZMode property on CacheCluster objects

cdodd:
Regarding the `AZMode` property on CacheCluster objects, the [AWS documentation](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticache-cache-cluster.html#cfn-elasticache-cachecluster-azmode) states :

```
If you specify multiple Availability Zones in the PreferredAvailabilityZones
property, you must specify cross Availability Zones for this property.
```

This patch introduces validation on the CacheCluster object and ensures that `AZMode` is set to `cross-az` if more than one Availability Zone is specified in the `PreferredAvailabilityZones` property.

### DIFF
--- a/troposphere/elasticache.py
+++ b/troposphere/elasticache.py
@@ -32,6 +32,21 @@ class CacheCluster(AWSObject):
         'VpcSecurityGroupIds': ([basestring, Ref, GetAtt], False),
     }
 
+    def validate(self):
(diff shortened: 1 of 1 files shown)


phobologic:
Thanks!

PR #255, opened 2015-05-26
Adding AutoScalingType to OpsWorks Instance

obiwanakin:


phobologic:
Hey @obiwanakin - thanks for fixing this.  One small thing: Can you update your change so that properties are in alphabetic order?  We try to keep them that way for the most part.  Thanks!

markpeek:
Thanks!

PR #257, opened 2015-06-02
allow extending classes + tests

### LINKED ISSUE #233: Extend classes to add class params
It would be useful to allow one to create custom classes which inherit the validation logic from toposphere's but allow arbitrary properties to be set on the class. That way one can use a class hierarchy to reduce repetition of definitions in similar resources.

anfedorov:
fixes #233

this should allow one to use class hierarchies to remove redundant
definition of properties in similar resources

### DIFF
--- a/examples/ClassExtensions.py
+++ b/examples/ClassExtensions.py
@@ -1,0 +1,21 @@
+import troposphere
+import troposphere.ec2
+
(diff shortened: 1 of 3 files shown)


markpeek:
This looks useful. Thanks for the pull request and tests.

PR #261, opened 2015-06-12
Add property for Subnet: MapPublicIpOnLaunch

drboyer:
Noticed this feature existed in the [CloudFormation Docs](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-subnet.html) but not here, so I added it! Simplest change ever :)


markpeek:
Thanks!

PR #264, opened 2015-06-12
Update LoginProfile

philtay:


philtay:
An example fails. I can update it if you want. Here the related docs

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-iam-user-loginprofile.html

markpeek:
The example appears to be a backwards compatibility issue where the new version needs to pass in a Password kwargs. Sounds like this would have to be added to the changelog as a release note.

philtay:
Do you want me to do something? It's pretty easy to fix. The culprit is [this line](https://github.com/cloudtools/troposphere/blob/master/examples/IAM_Users_snippet.py#L17).

markpeek:
@philtay thanks for checking, no need to do anything. I'm just getting agreement from @phobologic and @bobveznat before committing this "breaking" change. There is just some latency given the weekend.

PR #267, opened 2015-06-16
Added cfn2pyclass - this is backwards compatible. Creates python classes of resources for AWS

sbraverman:


sbraverman:
pyflakes is complaining about imported modules that aren't being used.
Need to update the way imports are generated for each class.

PR #269, opened 2015-06-17
adding t2 large

elliotmoore:
https://aws.amazon.com/about-aws/whats-new/2015/06/introducing-a-new-amazon-ec2-t2-instance-t2-large/


markpeek:
Thanks!

PR #270, opened 2015-06-17
Add ManagedPolicyArns property to AWS::IAM::Role

ProgramZeta:
Add the ManagedPolicyArns property to the Role object, listed in AWS's docs here:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html#cfn-iam-role-managepolicyarns

### DIFF
--- a/troposphere/iam.py
+++ b/troposphere/iam.py
@@ -3,7 +3,7 @@
 #
 # See LICENSE file for full license.
 
-from . import AWSObject, AWSProperty, Ref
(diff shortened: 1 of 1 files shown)


markpeek:
I pulled this change in via #272 (which added it to some of the other IAM resources). Note: I don't believe the Join is necessary to be added into the allowed types. Thanks for the pull request.

PR #272, opened 2015-06-18
status required and managedpolicyarns

iblazevic:
*status is required http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-iam-accesskey.html
* tests for validators
*adding managedPolicyArns

### DIFF
--- a/examples/IAM_Users_Groups_and_Policies.py
+++ b/examples/IAM_Users_Groups_and_Policies.py
@@ -21,6 +21,7 @@ cfnadmingroup = t.add_resource(Group("CFNAdminGroup"))
 
 cfnkeys = t.add_resource(AccessKey(
     "CFNKeys",
+    Status="Active",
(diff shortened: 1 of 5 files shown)


iblazevic:
Not sure about 'ManagedPolicyArns' I wasn't using them, should they maybe be some other type or just a plain string

markpeek:
I was wondering the same thing. I'm guessing both a string (basestring) and Ref() of a ManagedPolicy object. The Ref() should be handled implicitly but I'll have to test it later.

iblazevic:
Do you want me to pull it out of this PR and update status and tests until this is figured out?

markpeek:
Thanks for the offer. Just leave it in for now.

markpeek:
I wrote a quick test and ManagedPolicy looks good to me. Thanks for the pull request.

PR #275, opened 2015-06-19
CustomResource added - with example

hngkr:
CustomResource added - I've used it in my own work and found that it's great for implementing resource types currently not supported by CloudFormation (for instance PlacementGroup's)

It's my first PR, so I very well might have missed something.


hngkr:
Does not pass Travis CI - pyflakes error - will fix and resubmit

PR #278, opened 2015-06-19
CustomResource including example, updated todo in readme

hngkr:
CustomResource added - I've used it in my own work and found that it's great for implementing resource types currently not supported by CloudFormation (for instance PlacementGroup's).
Known limitations: The CustomResource properties are not checked.

It's my first PR, so I very well might have missed something.


markpeek:
Looks like a good enhancement. Thanks for adding it.

PR #282, opened 2015-07-05
Update opsworks.py

s1l0uk:
Adding the AutoScaling prop to Opsworks Instance - http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-opsworks-instance.html


markpeek:
This was already added in #255. Thanks though for submitting it.

PR #284, opened 2015-07-08
Update __init__.py

gimballock:
BaseAwsObject initialize flag set with one value and checked with another

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -53,7 +53,7 @@ class BaseAWSObject(object):
             self.resource = self.properties
         if hasattr(self, 'resource_type') and self.resource_type is not None:
             self.resource['Type'] = self.resource_type
-        self.__initialized = True
+        self._BaseAWSObject__initialized = True
 
         # Check for properties defined in the class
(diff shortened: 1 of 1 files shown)


markpeek:
I don't believe this is an issue. This is being used for the base class to initialize it (self.__initialized) and for the subclass to access it (self._BaseAWSObject__initialized). Here's a quick example:
```
>>> class BaseAWSObject(object):
...     def __init__(self):
...         self.__initialized = True
...
>>> class sub(BaseAWSObject):
...     pass
...
>>> s = sub()
>>> dir(s)
['_BaseAWSObject__initialized', '__class__', '__delattr__', '__dict__', '__doc__', '__format__', '__getattribute__', '__hash__', '__init__', '__module__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__']
>>> s._BaseAWSObject__initialized
True
```
Please let me know if there is a test case that is failing due to this implementation.

markpeek:
Response timeout. Please reopen with more information. Thanks.

PR #285, opened 2015-07-20
add FindInMap type check for AutoScalingGroup validation of group sizes

jdc0589:
The AutoScalingGroup validation already checks to see if the min/max sizes are instances of a `Ref` prior to casting them to integers and doing size validation. However, they could also be instances of `FindInMap`, which currently blows up on the integer cast if MaxSize is a `FindInMap`.

Thoughts?

### DIFF
--- a/troposphere/autoscaling.py
+++ b/troposphere/autoscaling.py
@@ -3,7 +3,7 @@
 #
 # See LICENSE file for full license.
 
-from . import AWSHelperFn, AWSObject, AWSProperty, Ref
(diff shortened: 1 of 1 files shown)


markpeek:
Merged and then tweaked it a little. Please check master for the new version and validate. Thanks!

jdc0589:
looks good to me. Thanks!
any chance of getting it published to  PyPI soon?

markpeek:
@jdc0589 sorry for the delay. This just got pushed out in v1.1.2 (and is on PyPI).

PR #295, opened 2015-08-17
Path is optional for Role and InstanceProfile

samcrang:
The `Path` property on IAM Roles and Instance Profiles is optional. I added an example that works when deploying to AWS and tweaked the appropriate bools to make this behaviour possible:

```bash
$ aws cloudformation create-stack --stack-name=test-no-path --template-body file://tests/examples_output/IAM_Roles_and_InstanceProfiles.template --capabilities "CAPABILITY_IAM"
{
    "StackId": "arn:aws:cloudformation:eu-west-1:MYACCOUNT:stack/test-no-path/GUID"
}
$ aws cloudformation describe-stacks
{
    "Stacks": [
        {
            "StackId": "arn:aws:cloudformation:eu-west-1:MYACCOUNT:stack/test-no-path/GUID", 
            "Description": "AWS CloudFormation Sample Template: This template demonstrates the creation of IAM Roles and InstanceProfiles.", 
            "Tags": [], 
            "CreationTime": "2015-08-17T14:09:59.837Z", 
            "Capabilities": [
                "CAPABILITY_IAM"
            ], 
            "StackName": "test-no-path", 
            "NotificationARNs": [], 
            "StackStatus": "CREATE_COMPLETE", 
            "DisableRollback": false
        }
    ]
}
```

As an aside this now lines up with the expected behaviour defined in the documentation for [Roles](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html#cfn-iam-role-path). However, the documentation for [InstanceProfiles](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-instanceprofile.html#cfn-iam-instanceprofile-path) continues to say the `Path` is required even though the above implies that it's not (at least not in all cases). I'm assuming the documentation is out of date in this instance.

### DIFF
--- a/examples/IAM_Roles_and_InstanceProfiles.py
+++ b/examples/IAM_Roles_and_InstanceProfiles.py
@@ -1,0 +1,32 @@
+from troposphere import Template, Ref
+
+from troposphere.iam import Role, InstanceProfile
+
+from awacs.aws import Allow, Statement, Principal, Policy
+from awacs.sts import AssumeRole
+
+t = Template()
+
+t.add_description("AWS CloudFormation Sample Template: This template "
+                  "demonstrates the creation of IAM Roles and "
+                  "InstanceProfiles.")
+
+cfnrole = t.add_resource(Role(
+    "CFNRole",
+    AssumeRolePolicyDocument=Policy(
+        Statement=[
+            Statement(
+                Effect=Allow,
+                Action=[AssumeRole],
+                Principal=Principal("Service", ["ec2.amazonaws.com"])
+            )
+        ]
+    )
+))
+
+cfninstanceprofile = t.add_resource(InstanceProfile(
+    "CFNInstanceProfile",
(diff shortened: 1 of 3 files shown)


phobologic:
Hey @samcrang - we've definitely seen the docs fall out of date with reality in the past.  Have you contacted AWS support about it?

Also, can you share the template that used to create the stack that demonstrates this?

samcrang:
It didn't even cross my mind to contact Amazon support about this--I'll do that too.

The template I used to show that `Path` isn't required is actually part of the [pull request and forms one of the tests](https://github.com/samcrang/troposphere/blob/3e55a455963cf51c1e316b94fc888b4dc285f9cb/tests/examples_output/IAM_Roles_and_InstanceProfiles.template).

markpeek:
Thanks!

PR #300, opened 2015-08-25
Add constants for the parameter types

ikben:
This includes all the parameter types from http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html

### DIFF
--- a/troposphere/constants.py
+++ b/troposphere/constants.py
@@ -202,3 +202,32 @@ CACHE_M2_4XLARGE = 'cache.m2.4xlarge'
 CACHE_C1_XLARGE = 'cache.c1.xlarge'
 
 CACHE_T1_MICRO = 'cache.t1.micro'
+
+#
+# Parameter types
+#
+STRING = 'String'
+NUMBER = 'Number'
+LIST_OF_NUMBERS = 'List<Number>'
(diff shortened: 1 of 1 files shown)


troposphere/constants.py
```diff
@@ -202,3 +202,32 @@
 CACHE_C1_XLARGE = 'cache.c1.xlarge'
 
 CACHE_T1_MICRO = 'cache.t1.micro'
+
+#
+# Parameter types
+#
+STRING = 'String'
```
phobologic: So I worry about generic constants like this - making this just STRING assumes that there'd be no use of STRING elsewhere, and could lead to confusion.  I almost wonder if we shouldn't turn constants into a module, and have things like parameter constants, etc.

ikben: I agree. I prefer using a module, because adding something like _TYPE to all of these will make most of them very long.

I currently use a constants module in my projects, having this as `constants.parameters.types.STRING`. I also have a `constants.parameters.patterns` for common validation patterns (alphanumeric, ...), so depending on which kind of constants you want in troposphere, this could also be `constants.parameter_types`.

markpeek:
Thanks!

PR #301, opened 2015-08-25
Add lambda ZipFile property

ikben:


troposphere/awslambda.py
```diff
@@ -4,11 +4,35 @@
...
+        'S3Bucket': (basestring, False),
+        'S3Key': (basestring, False),
         'S3ObjectVersion': (basestring, False),
+        'ZipFile': (basestring, False)
     }
 
+    def __init__(self, title=None, **kwargs):
+        super(Code, self).__init__(title, **kwargs)
+
+        if 'ZipFile' in kwargs:
```
phobologic: I would move this logic into validate() - the reason being that many times I've seen people create a troposphere resource/property, then modify it's attributes after the fact.  Validate() lets them do that still, without running into this.

ikben: I didn't know about validate(), and will look into this. I was copying from some other part of troposphere (don't remember which one).

phobologic: Cool - it may be that we do it like this somewhere else, but I think validate() is probably the right place in general.  @markpeek - any opinion?

ikben:
@phobologic I've added the logic to the validate method (and rewrote it a little bit).

If you like this as a single commit I can squash it too.

markpeek:
Thanks!

PR #303, opened 2015-08-27
Allow an optional resource_prefix in Template

phobologic:
I find myself doing "MyStack" + <resource_name> really often in my
templates. It seemed like there would be a better way, and so I thought
I'd just try something.  I'm not 100% sold on this - it's a little
magical, but I thought I'd put it out there since it was such a simple
change and it could clean up a lot of boilerplate code.

The place where I can see this being problematic is when you later are
Ref() or GetAtt()'ing a resource that has had a prefix added to it.

Anyway, just an experiment. Be curious what people think about it.

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -36,6 +36,13 @@ class TestBasic(unittest.TestCase):
         instance = ExtendedInstance('ec2instance', attribute='value')
         self.assertEqual(instance.attribute, 'value')
 
+    def test_resource_prefix(self):
+        name = "FakeInstance"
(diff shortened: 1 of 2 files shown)


ikben:
I think I'm missing the use case here. I can't think of a different behavior between adding MyStack in front of everything and not adding MyStack at all.

phobologic:
@ikben Actually, the more I think about it, the less I think this is necessary/useful.  It's based around old assumptions about how to handle templates that I think I've moved past.  Thanks for the head-check @ikben!  Closing this.

PR #305, opened 2015-09-03
Add AWS::EC2::VPCEndpoint support

cdodd:
Add support for VPC endpoints, as per http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-vpcendpoint.html

### DIFF
--- a/troposphere/ec2.py
+++ b/troposphere/ec2.py
@@ -402,6 +402,16 @@ class VPNConnectionRoute(AWSObject):
     }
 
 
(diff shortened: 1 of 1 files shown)


pas256:
Oh, I just noticed this. I submitted #307, sorry for the duplicate work.

However, looking closer, #307 is the full implementation of the doc including supporting `Ref`s and a policy document.

cdodd:
True, I'll close this PR.

PR #315, opened 2015-09-09
Refactor tail

mhahn:


mhahn:
re: lots of junk, thats why i included the `include_initial`, i am always passing that in as false in the stacker lib to avoid getting the junk: https://github.com/remind101/stacker/pull/90/files#diff-65a106c54767f86dd2517c76ee516287R127.

checking the timestamp works too, but i think then we'd have to parameterize that value and i think `include_initial` gets you most of the way there.

for stacks that don't exist yet, i just start tailing and handle the boto server error and retry so that i get the first logs, during an update i'll only get the update logs.

mhahn:
re: stack_making_progress, i was trying to change as little as possible so i didn't break any existing workflow, i'm handling this manually within stacker: https://github.com/remind101/stacker/pull/90/files#diff-206158bd04d7e3c19da1781d581c956fR208.

if the stack was in a state of "ROLLBACK_COMPLETE", would the check for `endswith('IN_PROGRESS')` still print the reason behind the rollback? looking at some of my cloud formation event streams i think the answer is yes since those specific resources would print why they failed.

but if that doesn't break an existing workflow, i'm happy to make that change!

markpeek:
Thanks!

Issue #320, opened 2015-09-14
RDS validation check for "AvailabilityZone and MultiAZ" breaks conditional parameters


sterwill:
This validation check in rds.py:195 causes my valid templates to fail:

        if 'AvailabilityZone' in self.properties and \
                self.properties.get('MultiAZ', None):
            raise ValueError("AvailabiltyZone cannot be set on DBInstance if "
                             "MultiAZ is set to true.")

My code, which worked before this validation check was added, sets MultiAz to a Ref of a required boolean parameter and AvailabilityZone to a conditional that may evaluate to AWS::NoValue (meaning CloudFormation disregards the parameter):

    database = template.add_resource(
        troposphere.rds.DBInstance(
            ...
            AvailabilityZone=If(
                'DbAvailabilityZoneSpecified',
                Ref(db_availability_zone),
                Ref(AWS_NO_VALUE)),
            ...
            MultiAZ=Ref(db_multi_az),
        )
    )

This is the only pattern I know of for handling optional parameters in nested stacks when the resource requires that a parameter be unset to trigger the default.  

I think the fix is to simply remove the exclusivity check in validate(), or possibly fail only if the value is not one of the conditional functions.

markpeek:
@sterwill looks like we both provided fixes for this issue. :-) I'll let @phobologic take a look and review.

sterwill:
Your is the better fix (and I made a mess of my branch history in my fork) so feel free to ignore mine. 

Thanks for the lightning fast fix.

sterwill:
types.NoneType was removed from Python 3, so the check may have to add some additional "is None" checks.

markpeek:
Merged. Thanks for the bug report.

PR #321, opened 2015-09-14
Fix issue #320 (don't validate AWSHelperFns)

sterwill:
Only check for the presence of both MultiAZ and AvailabilityZone when they are both not AWSHelperFn instances.


markpeek:
Closing per @sterwill by pulling in a different patch.

PR #327, opened 2015-09-23
tail: only add unseen events

amosshapira:
Since you already know whether the even was seen or not, take advantage of that knowledge to avoid redundant work.


markpeek:
@mhahn As a consumer of the tail functionality, what do you think of this change? Will this help or break your usage in stacker?

amosshapira:
This is a tiny optimisation, not a functional change.

mhahn:
+1, this is already a set so its functionaly the same thing

markpeek:
Thanks!

PR #329, opened 2015-10-02
Add ElastiCache (Redis) Example

rcuza:
This PR creates a python script that produces a CloudFormation template that produces a Redis ElastiCache cluster and a web server instance for interacting with it. This script reproduces an example from [AWS CloudFormation templates](http://aws.amazon.com/cloudformation/aws-cloudformation-templates/).

The underscore in the python script name was removed to make it a valid class name, in case someone wants to evolve the script in that direction.

The PR includes the output from the script as an output_example in the tests directory. I am not sure how to use the test tools, so I did not verify using them. I did successfully bring up the infrastructure using the outputted template.


rcuza:
Suggestion: It would make for more readable embedded scripts if the check allowed lines to be 100 long as opposed to 80. Not a big deal, but offering it as a suggestion.

Thank you for such a useful tool!

markpeek:
Thank you for adding this example (and fixing the pep8 issues). Good suggestion on relaxing the line length limitation. Let me discuss with the other committers.

PR #334, opened 2015-10-20
Add __add__ method to Tags object to allow concatenation

rednut:
## Requirement

To be able to define a common list of stack tags and be able to apply them to a set of resource tags, eg:

```
CommonTags=Tags(Bar='foo')
Tags=CommonTags + Tags(Foo='Bar')
```

## Summary

Added a `__add__` method to the `Tags` class to facilitate using the append / concatenate operator `+` or `+=` to join(append) multiple Tags objects together thus allow things like `CommonTags` to be applied to a custom resource specific list of `Tags`.

Internally this uses the `tags` list/array within the `Tags` object.

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -360,6 +360,15 @@ class Tags(AWSHelperFn):
                 'Value': v,
             })
 
+    # allow concatenation of the Tags object via '+' operator
+    def __add__(self, newtags):
+        self.tags.append(newtags.tags)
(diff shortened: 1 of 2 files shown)


rednut:
updated to fix unpassing tests

markpeek:
Thanks for adding this feature. I squashed your commits before merging, reworked it a little to maintain ordering, and added a test.

PR #338, opened 2015-10-28
Correct EC2 SpotFleet LaunchSpecifications

xabufr:
LaunchSpecifications should be an array, as in the reference documentation
http://docs.aws.amazon.com/fr_fr/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-spotfleet-spotfleetrequestconfigdata.html


markpeek:
Thanks!

PR #339, opened 2015-10-29
Fixing LaunchSpecifications should be an array

renanvicente:
Fixing TypeError: <class 'troposphere.ec2.SpotFleetRequestConfigData'>: None.LaunchSpecifications is <type 'list'>, expected <class 'troposphere.ec2.LaunchSpecifications'>

Since LaunchSpecifications should be an array.

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-spotfleet-spotfleetrequestconfigdata.html#cfn-ec2-spotfleet-spotfleetrequestconfigdata-launchspecifications

### DIFF
--- a/troposphere/ec2.py
+++ b/troposphere/ec2.py
@@ -498,7 +498,7 @@ class LaunchSpecifications(AWSProperty):
 class SpotFleetRequestConfigData(AWSProperty):
     props = {
         'IamFleetRole': (basestring, True),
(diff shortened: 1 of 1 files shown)


markpeek:
This is a dup of #338 which just got merged (along with another fix). Thank you for the PR.

PR #345, opened 2015-11-10
Update codedeploy.pp

thkrmr:
As per the docs, S3Location Version Property is not a default requirement. 
http://docs.aws.amazon.com/codedeploy/latest/APIReference/API_S3Location.html


markpeek:
Thanks!

PR #353, opened 2015-11-27
Privileged attribute support for ECS container definition

cornelf:
* [x] Ready for review

Adds `Privileged` attribute support for ECS container definition


cornelf:
AWS CF does not support this.

Issue #354, opened 2015-11-30
DeletionPolicy ignored for LogGroup unless RetentionInDays is also specified


simonschmidt:
When creating a `LogGroup` the deletion policy is lost unless `RetentionInDays` is also specified:

```python
from troposphere import Template, Retain
from troposphere.logs import LogGroup

t = Template()
t.add_version("2010-09-09")
t.add_description("Test DeletionPolicy for LogGroup")

t.add_resource(LogGroup(
    "LogGroupWithIgnoredDeletionPolicy",
    DeletionPolicy=Retain
))

t.add_resource(LogGroup(
   "LogGroupWithDeletionPolicyAndRetentionInDays",
   DeletionPolicy=Retain,
   RetentionInDays=7
)) 

print(t.to_json())
```

#### Output

```json
{
    "AWSTemplateFormatVersion": "2010-09-09",
    "Description": "Test DeletionPolicy for LogGroup",
    "Resources": {
        "LogGroupWithDeletionPolicyAndRetentionInDays": {
            "DeletionPolicy": "Retain",
            "Properties": {
                "RetentionInDays": 7
            },
            "Type": "AWS::Logs::LogGroup"
        },
        "LogGroupWithIgnoredDeletionPolicy": {
            "Type": "AWS::Logs::LogGroup"
        }
    }
}
```

craigbruce:
I don't think this is restricted to just LogGroup, I'm struggling to get S3 to work

```
from troposphere import Template, Retain, s3

t = Template()
t.add_version("2010-09-09")
t.add_description("Test DeletionPolicy for S3")

t.add_resource(s3.Bucket(
    "BucketWithIgnoredDeletionPolicy",
    DeletionPolicy=Retain
))

print(t.to_json())
```
Output
```
{
    "AWSTemplateFormatVersion": "2010-09-09",
    "Description": "Test DeletionPolicy for S3",
    "Resources": {
        "BucketWithIgnoredDeletionPolicy": {
            "Type": "AWS::S3::Bucket"
        }
    }
}
```

PR #355, opened 2015-11-30
Possible fix for #354

simonschmidt:
This is more explicit about only filtering out 'Properties' to avoid `DeletionPolicy` getting removed for LogGroup in #354 
Not sure if this has any unintended side effects, but I ran the tests and they passed.

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -165,7 +165,10 @@ class BaseAWSObject(object):
         if self.properties:
             return self.resource
         elif hasattr(self, 'resource_type'):
-            return {k: v for k, v in self.resource.items() if k != 'Properties'}
+            return {
+                k: v for k, v in self.resource.items()
+                if k != 'Properties'
+            }
         else:
             return {}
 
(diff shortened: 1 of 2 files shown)


markpeek:
@simonschmidt I squashed the commits and pushed it here https://github.com/cloudtools/troposphere/commit/77320c91cf86589df0f1e861d1a273d251f4f127. Thank you for the fix.

troposphere/__init__.py
```diff
@@ -160,12 +160,14 @@ def JSONrepr(self):
...
                     "Resource %s required in type %s" % (k, rtype))
         self.validate()
-        # If no other properties are set, only return the Type.
         # Mainly used to not have an empty "Properties".
         if self.properties:
             return self.resource
         elif hasattr(self, 'resource_type'):
-            return {'Type': self.resource_type}
+            return {
+                k: v for k, v in self.resource.items()
```
oliviervg1: Hi @simonschmidt,

This change has caused Troposphere to throw errors when using python 2.6 as it doesn't support dictionary comprehensions.

Would it possible to fix this per chance?

Thanks!

oliviervg1:
Hi,

I'd like to re-open this issue as it Troposphere now throws an error when using Python 2.6. I've commented on the offending line.

Thanks!

markpeek:
@oliviervg1: I just pushed a change to support 2.6.9.

PR #357, opened 2015-12-15
Added applicationName parameter to the application

theseanything:
Previously missing the applicationName parameter to the Application Resource


markpeek:
@theseanything thank you for the PR.
@Gerrit-Janssen thanks for noticing the required field...I pushed that change as well.

PR #358, opened 2015-12-15
DeploymentGroupName property missing

theseanything:
DeploymentGroupName property missing from the DeploymentGroup Resource


markpeek:
Thanks!

Issue #359, opened 2015-12-21
CloudFront TTL issue - undocumented AWS change


slikk66:
Suddenly working with CloudFront via CloudFormation if you pass in a MinTTL value for your cache behavior, you will get an error saying:

**MinTTL must be less equal to DefaultTTL must be less equal to MaxTTL**

however, according to the docs (as of 12/21) there are no DefaultTTL or MaxTTL values: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-defaultcachebehavior.html#cfn-cloudfront-defaultcachebehavior-minttl

Troposphere reflects this fact and won't let you add MaxTTL or DefaultTTL.  However, now all CloudFront adjustments using TTLs will fail.  If you manually add in the MaxTTL and DefaultTTL values to the code, now it works!  Just verified and tested working with adjusted code.  PR to fix issue is #360

markpeek:
Fixed with merge of #360

PR #360, opened 2015-12-21
add in properties for max, default

slikk66:


markpeek:
@slikk66 thank you for the fix. Per the error in #359, should there be logic added to validate these TTL values by adding a validate() method?

slikk66:
Personally I'm not a fan of the local validate() methods, we've actually had an issue before where the logic in the validate() we had to remove due to the complexity of the situation (multiple cache nodes + zones restrictions).  I'm sure it would be a simple **X<=Y<=Z** but the decision whether or not to include would be up to the maintainers!

markpeek:
Thanks!

PR #362, opened 2015-12-22
Add CodePipeline support

craigbruce:
As of Dec 3 2015 CloudFormation supports CodePipeline http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/ReleaseHistory.html


markpeek:
Looks great! Thanks for also including the example.

PR #363, opened 2015-12-28
ElastiCache: NotificationTopicArn is a string not a list of strings. 

craigbruce:
See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticache-replicationgroup.html#cfn-elasticache-replicationgroup-notificationtopicarn


markpeek:
Thanks!

PR #376, opened 2016-01-11
added a helper script to import userdata from a file

theseanything:
Created a helpers folder inside troposphere that contains helper scripts. Created a userdata helper script that contains a function to import userdata from a file. Created tests for this helper script along with dummy userdata files to be imported, all located within the testing folder.


markpeek:
@theseanything thank you for the pep8/pyflakes changes. I believe the final issue with getting it to pass tests is adding 'troposphere.helpers' into the setup.py file.

Looking over the change there are two questions I have:
1. Error handling: the use of "print" to indicate errors is a little non-standard. I'd probably prefer to fail with an exception if the file doesn't exist or has a parse issue.
2. Is there a use case to not strip the blank lines within the file? Should this be an option (with a default) for the function?

Thoughts?

theseanything:
@markpeek thanks, it was the setup.py file!

So not exactly sure how to elegantly handle the file errors. (All the example on the internet used Print(), not a fan of that either).  It now raises an exception, instead of printing a message. Hope this is more like the rest of troposphere!

Agreed that the stripping of files should be a option! But decided to get rid of it all together, and possibly add it properly later.

Cheers mark!

markpeek:
Sorry for the delay in merging this PR. There were some extra files in env/ so I did some rebase work to remove them. The commit got merged via a25db2aaff290aa34099c62bbad93b498e5aa8d1. Thanks!

philipbjorge:
@theseanything Why are you joining these things with a comma? When I read in a bash script, cloudformation outputs the following which obviously doesn't work:

```
#!/bin/bash
,echo hello
,echo world
```

PR #380, opened 2016-01-29
ReplicationConfigurationRules Destination is now an object

mbrossard:
In [Amazon S3 ReplicationConfiguration Rules](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-replicationconfiguration-rules.html) the Destination field use to be a string but it's now an object with [2 documented fields](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-replicationconfiguration-rules-destination.html): Bucket and StorageClass.

### DIFF
--- a/troposphere/s3.py
+++ b/troposphere/s3.py
@@ -176,9 +176,16 @@ class NotificationConfiguration(AWSProperty):
     }
 
 
+class ReplicationConfigurationRulesDestination(AWSProperty):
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks and sorry for the delay.

PR #385, opened 2016-02-15
Give more info on func validator failures

phobologic:
Not sure how best to test this - open to any suggestions.  Also not a
huge fan of the way I'm giving the info, but can't think of a better way
without breaking any previous use that already handled errors in a
particular way.

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -6,6 +6,7 @@
 
 import json
 import re
+import sys
 import types
 
 from . import validators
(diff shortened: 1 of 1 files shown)


phobologic:
BTW, this results in this sort of error: ```
<class 'troposphere.directoryservice.SimpleAD'>: simplead.CreateAlias function validator 'boolean' threw exception:
  File "<stdin>", line 8, in <module>
  File "troposphere/__init__.py", line 67, in __init__
    self.__setattr__(k, v)
  File "troposphere/__init__.py", line 105, in __setattr__
    value = expected_type(value)
  File "troposphere/validators.py", line 12, in boolean
    raise ValueError
ValueError
```

Still kind of hard to see, but at least something is there (right at the top, before the exception).

PR #389, opened 2016-02-23
extended title validation

webratz:
This PR adds the option to have custom title validators, as the default one is often to restrictive (AWS allows more characters)
Also there are some new checks for IAM Objects which help you catching errors before you run the CF template.

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -38,8 +38,11 @@ class BaseAWSObject(object):
                            'Metadata', 'UpdatePolicy',
                            'Condition', 'CreationPolicy']
 
-        # unset/None is also legal
-        if title and not valid_names.match(title):
+        # some objects allow more characters than others
+        if title and 'validate_title' in dir(self):
+            self.validate_title()
+        elif title and not valid_names.match(title):
+            # unset/None is also legal
             raise ValueError('Name "%s" not alphanumeric' % title)
 
         # Create the list of properties set on this object by the user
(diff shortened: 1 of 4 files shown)


troposphere/__init__.py
```diff
@@ -38,8 +38,11 @@ def __init__(self, title, template=None, **kwargs):
                            'Metadata', 'UpdatePolicy',
                            'Condition', 'CreationPolicy']
 
-        # unset/None is also legal
-        if title and not valid_names.match(title):
+        # some objects allow more characters than others
+        if title and 'validate_title' in dir(self):
```
phobologic: Does it maybe make more sense to do away with this logic and instead move the `valid_names.match` into a default `validate_title` method?  It'd make this a bit cleaner.  This would become:

```
if title:
    self.validate_title()
```

webratz:
You are right, this is a nicer way to implement this. Updated the PR

troposphere/__init__.py
```diff
@@ -38,9 +38,9 @@ def __init__(self, title, template=None, **kwargs):
                            'Metadata', 'UpdatePolicy',
                            'Condition', 'CreationPolicy']
 
-        # unset/None is also legal
-        if title and not valid_names.match(title):
-            raise ValueError('Name "%s" not alphanumeric' % title)
+        # try to validate the title if its there
+        if self.title is not None:
```
phobologic: `if self.title` is more pythonic, and should do what you want here.  Otherwise I think this is good to go.

webratz:
updated

phobologic:
Awesome, thanks @webratz !

PR #394, opened 2016-02-26
Added EC2 Nat Gateway resource

jarosser06:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-natgateway.html

cc @martinb3

### DIFF
--- a/examples/NatGateway.py
+++ b/examples/NatGateway.py
@@ -1,0 +1,123 @@
+# Example Network with a NAT Gateway
+
+from troposphere import Output, Ref, Template, Parameter, GetAtt
+from troposphere import ec2
+
+
+t = Template()
+
+t.add_description(
+    "AWS CloudFormation Sample Template NatGateway: Sample template showing "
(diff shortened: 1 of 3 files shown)


mwarkentin:
So excited.. :godmode:

tfeak:
Thank you for this. 

Though, the Route resource will also need updating for this to be useful. They've added a NatGatewayId property to allow you to add the Nat to your route tables.

jarosser06:
I'll take a look and add that change as well.

martinb3:
I went to the docs for AWS::EC2::Route and added the missing field, and verified there weren't any others missing. This should be good to go.

DaveHewy:
LGTM. Maybe some example usage in examples/ ?

jarosser06:
Added an example in examples.

phobologic:
Awesome - so ready to nuke my NAT boxes.  Thanks @jarosser06!

phobologic:
Thanks!

PR #395, opened 2016-02-26
AWS::ECR::Repository

philtay:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecr-repository.html

### DIFF
--- a/troposphere/ecr.py
+++ b/troposphere/ecr.py
@@ -5,6 +5,7 @@ try:
 except ImportError:
     policytypes = dict,
 
+
 class Repository(AWSObject):
     resource_type = "AWS::ECR::Repository"
 
--- a/troposphere/ecr.py
+++ b/troposphere/ecr.py
@@ -1,0 +1,14 @@
+from . import AWSObject
+try:
+    from awacs.aws import Policy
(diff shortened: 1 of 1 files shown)


jeanphix:
@philtay maybe you could ``squash`` these two commits.

phobologic:
Looks good @philtay - could you add an example so that we get tests?  The one here should be good: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecr-repository.html#d0e45175

philtay:
@phobologic Currently Amazon ECR is not present in awacs. A testable example like the one you linked can be added only after the inclusion of ECR in awacs. Moreover, given the fact that troposphere installs awacs from PyPI, a new version of awacs should be released before adding the example. I'm going to open a PR on awacs for this. In the meantime, considering that this is a very simple resource (15 lines of Python the entire file), I think that it can be safely merged. As soon as a new release of awacs will be out I will add the test.

phobologic:
Ok - I honestly was fine with just a dict implementation for an example/tests, but I can see waiting for an awacs integrated example as well.  Thanks @philtay

PR #396, opened 2016-02-26
Add KmsKeyId and StorageEncrypted to DBCluster

philtay:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbcluster.html


phobologic:
Great, thanks!

PR #397, opened 2016-02-26
recommended soft dependency

philtay:
`pip install troposphere[policy]` will install troposphere + awacs

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -29,6 +29,11 @@ issuing::
 
     $ pip install troposphere
 
+To install troposphere with `awacs <https://github.com/cloudtools/awacs>`_
+(recommended soft dependency)::
+
+    $ pip install troposphere[policy]
+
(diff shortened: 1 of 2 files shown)


phobologic:
Awesome, thanks!

hammadzz:
How does this work? Do I insert something in policy? As it is, it does not work for me on pip 9.0.1

philtay:
@hammadzz run the following command in your terminal:

```$ pip install troposphere[policy]```

`pip` will install both `troposphere` and `awacs`.

hammadzz:
@philtay which pip version does that command work on? Does not work on pip 9.0.1

philtay:
@hammadzz it works on pip 9.0.1, I just tried. This is standard pip / setuptools feature. Google for `pip extras_require` to find out more.

PR #404, opened 2016-02-29
Capitalize examples

martinb3:
Use capital initial letters for examples. Makes it easier to locate the Lambda examples.


phobologic:
Thanks!

PR #410, opened 2016-03-01
Support for ECR Repositories

matalo33:
New class and object to support creation of ECR repositories
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecr-repository.html


matalo33:
*sigh*, duplicates #395. Didn't check open PRs before doing this.

PR #414, opened 2016-03-03
Add support for elasticsearch domains

irvined1982:
For example:
`es = t.add_resource(ElasticSearchDomain(
        "MySearchDomain",
        DomainName="mysearchdomain",
        AccessPolicies={
                        "Version": "2012-10-17",
                        "Statement": [
                            {
                              "Effect": "Allow",
                              "Principal": {
                                "AWS": "arn:aws:iam::12345:root"
                              },
                              "Action": "es:*",
                              "Resource": "*"
                            },
                        ]
                    }
    ))`

### DIFF
--- a/troposphere/elasticsearch.py
+++ b/troposphere/elasticsearch.py
@@ -1,0 +1,48 @@
+# Copyright (c) 2013, David Irvine <irvined@gmail.com>
+# All rights reserved.
+#
+# See LICENSE file for full license.
+
+from . import AWSObject, Ref, GetAtt, Tags, AWSProperty
+from .validators import boolean, integer
+
+
+class ElasticSearchEBSOption(AWSProperty):
+    props = {
+        "EBSEnabled": (boolean, False),
+        "Iops": (integer, False),
+        "VolumeSize": (integer, False),
+        "VolumeType": (basestring, False),
+    }
+
+
+class ElasticSearchClusterConfig(AWSProperty):
+    props = {
+        "DedicatedMasterCount": (integer, False),
+        "DedicatedMasterEnabled": (boolean, False),
+        "DedicatedMasterType": (basestring, False),
+        "InstanceCount": (integer, False),
+        "InstanceType": (basestring, False),
+        "ZoneAwarenessEnabled": (boolean, False)
+    }
+
+
+class ElasticSearchSnapshotOption(AWSProperty):
+    props = {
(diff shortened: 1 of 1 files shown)


phobologic:
@irvined1982 it looks like there is a blank line at the end of the commit that is breaking pep8, can you fix that?

Also can you add this example to the `examples` directory (be sure to put its output in `tests/examples_output` as well): http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticsearch-domain.html#d0e51584

Thanks for submitting this!

troposphere/elasticsearch.py
```diff
@@ -0,0 +1,48 @@
...
+# See LICENSE file for full license.
+
+from . import AWSObject, Ref, GetAtt, Tags, AWSProperty
+from .validators import boolean, integer
+
+
+class ElasticSearchEBSOption(AWSProperty):
+    props = {
+        "EBSEnabled": (boolean, False),
+        "Iops": (integer, False),
```
phobologic: The validator for this (and a few others in here) should probably be `positive_integer` I think.

troposphere/elasticsearch.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class ElasticSearchDomain(AWSObject):
+    resource_type = "AWS::Elasticsearch::Domain"
+
+    props = {
+        'AccessPolicies': (dict, False),
+        'AdvancedOptions': (dict, False),
+        'DomainName': (basestring, True),
+        'EBSOptions': (ElasticSearchClusterConfig, False),
```
phobologic: The expected_type is wrong here.

troposphere/elasticsearch.py
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) 2013, David Irvine <irvined@gmail.com>
+# All rights reserved.
+#
+# See LICENSE file for full license.
+
+from . import AWSObject, Ref, GetAtt, Tags, AWSProperty
+from .validators import boolean, integer
+
+
+class ElasticSearchEBSOption(AWSProperty):
```
phobologic: I would drop the `ElasticSearch` in all of these class names.  The module path should handle that for us.  So this would become just `EBSOptions`

phobologic:
Going to go ahead and close this since #415 seems to be further along.  Thanks for the submission!

PR #416, opened 2016-03-04
Apply 2to3 on cfn2py script

satyrius:
`script/cfn2py` doesn't work on python 3, generally because of `print` function


satyrius:
py2 and py3 results differs, working on it

satyrius:
not it works fine

PR #418, opened 2016-03-04
Update config.py

scottjpack:
Fixing DeliveryChannel ConfigSnapshotDeliveryProperties type


scottjpack:
ConfigSnapshotDeliveryProperties property is set as basestring, should be ConfigSnapshotDeliveryProperties

PR #421, opened 2016-03-05
Adding support for EMR resources

alonsodomin:
Created new module to hold all the new EMR resources and types plus a basic example that creates an EMR cluster will all the applications (Hadoop, Spark, etc.) installed in it.


troposphere/emr.py
```diff
@@ -0,0 +1,126 @@
...
+    def __init__(self, key, value):
+        self.data = {'Key': key, 'Value': value}
+
+    def JSONrepr(self):
+        return self.data
+
+
+class Application(AWSProperty):
+    props = {
+        'AdditionalInfo': (dict, False),
```
phobologic: This doesn't appear to be a dict in the Cloudformation docs - it might call for it's own validator.

troposphere/emr.py
```diff
@@ -0,0 +1,126 @@
...
+        'Args': (list, False),
+        'Name': (basestring, False),
+        'Version': (basestring, False)
+    }
+
+
+class BootstrapActionConfig(AWSProperty):
+    props = {
+        'Name': (basestring, True),
+        'ScriptBootstrapAction': (basestring, True)
```
phobologic: Despite the example saying this is a string, it looks like it's another property type if you look at the full explanation for that field: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-emr-cluster-bootstrapactionconfig.html#cfn-emr-cluster-bootstrapactionconfig-scriptbootstrapaction

troposphere/emr.py
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+
+class ClusterConfiguration(AWSProperty):
+    props = {
+        'Classification': (basestring, False),
+        'ConfigurationProperties': (dict, False),
```
phobologic: This looks like it would use the Validator from AdditionalInfo above.

troposphere/emr.py
```diff
@@ -0,0 +1,126 @@
...
+        'Configurations': (list, False)
```
phobologic: This looks like it should be a list of it's own property types: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-emr-cluster-configuration.html#cfn-emr-cluster-configuration-configurations

troposphere/emr.py
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+
+class InstanceGroupConfigProperty(AWSProperty):
+    props = {
+        'BidPrice': (basestring, False),
+        'Configurations': (list, False),
```
phobologic: Looks like this is a list of it's own property type from the docs.

troposphere/emr.py
alonsodomin: `AdditionalInfo` seems to be documented as a _JSON Object_, that's why I chose the `dict` to represent it. Is there a better way?

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-emr-cluster.html#cfn-emr-cluster-additionalinfo

phobologic: That's in the `Cluster` resource, not the `Application` Property: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-emr-cluster-application.html#cfn-emr-cluster-application-additionalinfo

troposphere/emr.py
alonsodomin: Mind pointing to an example of how to express a list in which each item must be custom validated? Added the `emr_configuration_property` for the `ConfigurationProperties` as it seems to be a list of _colon-separated key-value pairs_ but if I express it as:

```python
'ConfigurationProperties': ([emr_configuration_property], False)
```

I get the following error when running the tests:

```
TypeError: isinstance() arg 2 must be a class, type, or tuple of classes and types
```

It works fine if it weren't a list - it validates the value using the regex.

troposphere/emr.py
alonsodomin: oh yes! sorry, it's using another _colon-separated key-value pair_ list... ¬¬

troposphere/emr.py
alonsodomin: Ok, think I got it, I'll need to write a custom `AWSHelperFn`

alonsodomin:
Right, so I added more strict validations and enriched the test template with more details plus the sample step from AWS docs.

There seems to be some incongruences in the AWS docs anyway. One of them is regarding that self-recursive `Configuration` "dictionary" as in [here](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-emr-cluster-configuration.html) the docs say that it's compoound of three attributes: `Classification`, `ConfigurationProperties` and additional `Configuration`s. However when getting into more detail [here](http://docs.aws.amazon.com//ElasticMapReduce/latest/ReleaseGuide/emr-configure-apps.html), now the attribute `ConfigurationProperties` is named simply as `Properties`.

Another incongruence can be seen in the configuration of the step. The `HadoopJarStepConfig` structure documente in [here](http://docs.aws.amazon.com//ElasticMapReduce/latest/ReleaseGuide/emr-configure-apps.html) has an `Args` attribute which, apparently, is of type `String`. In the other hand, in the sample JSON given for an step configuration in [here](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-emr-step.html#cfn-emr-step-hadoopjarstep), we can see that the `Args` attribute is not a list of strings instead.

Will try to test the last version of the example template once I get a chance in our cluster to see what is the correct encoding.

phobologic:
Cool - thanks for diving into this @alonsodomin - I'll try to get a chance to look at it today!

alonsodomin:
Sorry for the delay on this, getting a slot to experiment cluster configuration in the cluster usually doesn't sit at the top of the priority list ;)

Anyway, it seems that the correct encoding for the cluster & instance group configuration, using `ConfigurationProperties` instead of `Properties` as some other docs says, is the right one.

Same for the `HadoopJarStepConfig`, the example JSON is right.

alonsodomin:
btw, also took the chance to rebase my fork on top of master.

troposphere/emr.py
```diff
@@ -0,0 +1,186 @@
...
+        if not isinstance(v, basestring):
+            raise ValueError('AdditionalInfo values must be strings')
+
+    return xs
+
+
+class Application(AWSProperty):
+    props = {
+        'AdditionalInfo': (additional_info_validator, False),
+        'Args': (list, False),
```
phobologic: I think this should be `[basestring]`

troposphere/emr.py
```diff
@@ -0,0 +1,186 @@
...
+            raise ValueError("Configuration '%s' must be of "
+                             "Configuration type" % x)
+    return xs
+
+
+class Configuration(AWSProperty):
+    props = {
+        'Classification': (basestring, False),
+        'ConfigurationProperties': (properties_validator, False),
+        'Configurations': (configurations_validator, False)
```
phobologic: I think the validator here could be replaced with just `[Configuration]`

troposphere/emr.py
```diff
@@ -0,0 +1,186 @@
...
+    if x not in valid_values:
+        raise ValueError("Market must be one of: %s" %
+                         ', '.join(valid_values))
+    return x
+
+
+class InstanceGroupConfigProperty(AWSProperty):
+    props = {
+        'BidPrice': (basestring, False),
+        'Configurations': (configurations_validator, False),
```
phobologic: Same as comment above ^^

troposphere/emr.py
```diff
@@ -0,0 +1,186 @@
...
+        'InstanceCount': (integer, True),
```
phobologic: This should probably be `positive_integer` - there can't be a negative number here right?

troposphere/emr.py
```diff
@@ -0,0 +1,186 @@
...
+
+class PlacementType(AWSProperty):
+    props = {
+        'AvailabilityZone': (basestring, True)
+    }
+
+
+class JobFlowInstancesConfig(AWSProperty):
+    props = {
+        'AdditionalMasterSecurityGroups': (list, False),
```
phobologic: This and the next should be `[basestring]`

troposphere/emr.py
```diff
@@ -0,0 +1,186 @@
...
+        'InstanceRole': (basestring, True),
+        'InstanceType': (basestring, True),
+        'Market': (market_validator, False),
+        'Name': (basestring, False)
+    }
+
+
+class HadoopJarStepConfig(AWSProperty):
+    props = {
+        'Args': ([basestring], False),
```
phobologic: Weird - in the docs this is listed as 'string', but then the doc paragraph itself says this is a list.  Any idea which is actually correct?

troposphere/emr.py
```diff
@@ -0,0 +1,186 @@
...
+        'Jar': (basestring, True),
+        'MainClass': (basestring, False)
```
phobologic: This is missing the `StepProperties` property.

troposphere/emr.py
alonsodomin: Yeah, there are some inconsistencies in the documentation, seems they wrote it in a rush! It seems that the actual value is a list of strings, as the doc example points out.

Will double check though.

troposphere/emr.py
alonsodomin: Yeah, totally right.

troposphere/emr.py
alonsodomin: Main reason I chose for this custom validator function in this specific case is because Python won't allow me to _self reference_ the `Configuration` type in the nested `Configurations` property.

troposphere/emr.py
alonsodomin: Stupid me, I put the `StepProperties` under the wrong type

alonsodomin:
Long time since last update. I managed to test the incongruence in the AWS docs regarding the `HadoopJarStepConfig` showing that the `Args` attribute is a `String` in one place but an array in another. As it is right now seems to be the correct implementation.

Let me know your thoughts about the other comments (i.e.: the validator for the self-recursive `Configuration` type) and whether you are happy or not to start the standard rebase-squash process to get the branch ready for being merged...

troposphere/emr.py
phobologic: Ahh, gotcha.  Ok, that makes sense.

phobologic:
Great - thanks @alonsodomin - I went ahead and used the new 'squash & merge' feature of github to merge this.

PR #422, opened 2016-03-05
Fix `ecs.TaskDefinition.Volumes` that was incorrectly flagged as required

jeanphix:
See: http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html


phobologic:
Thanks!

PR #423, opened 2016-03-05
AWS::ECR test example

philtay:
@phobologic as promised

ref: https://github.com/cloudtools/troposphere/pull/395


phobologic:
Awesome, thanks @philtay

PR #427, opened 2016-03-11
Add cloudfront hostedzoneid for route53

khornberg:
When creating an A record for a cloudfront distribution, the `AliasTarget` must refer to the cloudfront `HostedZoneId`.

Reference: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-aliastarget.html


phobologic:
Awesome, thanks!

PR #432, opened 2016-03-15
ScalingAdjustment is an integer

victortrac:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-policy.html#cfn-as-scalingpolicy-scalingadjustment


markpeek:
Thanks!

PR #435, opened 2016-03-17
Provide Valid Lambda Function Memory Values for use in Parameters

ink-adavison:
Lambda functions can only have certain amounts of memory in multiples of 64
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-memorysize

This commit allows MEMORY_VALUES to be used against Parameters' AllowedValues attribute, so that the Cloudformation web console form has a drop down selection of memory values in the same way it currently is in the Lambda web console.

I've updated the Lambda example to demonstrate.

I'm aware this could become unwieldy if AWS increase the amounts of memory available to Lambda functions, but if that occurs, this could be reverted or compared with any UI changes in the Lambda web console.


phobologic:
Thanks @ink-adavison - do you think it'd be worth updating the validator for https://github.com/cloudtools/troposphere/blob/master/troposphere/awslambda.py#L53 ?

ink-adavison:
Like this? I've tested it against AWS helper functions (Ref, If), valid number strings, invalid number strings, actual text strings, booleans, valid and invalid numbers. Looks like it works as intended.

phobologic:
Sorry for the delay @ink-adavison - this looks great.  There are conflicts now - can you fix those and we'll merge ASAP?

ink-adavison:
ack I really messed that up, let me revert a load of this and try again

ink-adavison:
Sorry, this is so messed up I'm just going to nuke it and start again with a new pull request.

PR #437, opened 2016-03-21
Added missing S3OriginConfig parameter

jvgutierrez:
According to [AWS CloudFormation docs](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-origin.html#cfn-cloudfront-origin-s3origin) the parameter S3OriginConfig is mandatory for a S3 backed origin.


markpeek:
Thanks!

PR #447, opened 2016-04-04
Add support for new VpcConfig property in Lambda::Function

Bazze:
This adds support for the new `VpcConfig` property to `awslambda.Function`. I'm a first time contributor to troposphere so let me know if I've missed something!


Bazze:
I see now that this was already added over the weekend, great! Closing...

PR #449, opened 2016-04-04
Provide Valid Lambda Function Memory Values for use in Parameters

ink-adavison:
This is a retry of PR435 after I thoroughly messed up the commit history there. This PR should be much cleaner.

Lambda functions can only have certain amounts of memory in multiples of 64
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-memorysize

This commit allows MEMORY_VALUES to be used against Parameters' AllowedValues attribute, so that the Cloudformation web console form has a drop down selection of memory values in the same way it currently is in the Lambda web console.

I've updated the Lambda example to demonstrate.

I'm aware this could become unwieldy if AWS increase the amounts of memory available to Lambda functions, but if that occurs, this could be reverted or compared with any UI changes in the Lambda web console.

This also includes the updated validation as requested by phobologic

### DIFF
--- a/troposphere/awslambda.py
+++ b/troposphere/awslambda.py
@@ -1,10 +1,7 @@
 from . import AWSObject, AWSProperty
 from .validators import positive_integer
 
-MEMORY_VALUES = [128, 192, 256, 320, 384, 448, 512,
-                 576, 640, 704, 768, 832, 896, 960, 1024,
-                 1088, 1152, 1216, 1280, 1344, 1408, 1472,
-                 1536]
+MEMORY_VALUES = [x for x in range(128, 1600, 64)]
 
 
 def validate_memory_size(memory_value):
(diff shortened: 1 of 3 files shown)


troposphere/awslambda.py
```diff
@@ -1,6 +1,23 @@
 from . import AWSObject, AWSProperty
 from .validators import positive_integer
 
+MEMORY_VALUES = [128, 192, 256, 320, 384, 448, 512,
```
markpeek: Could you shorten this to:
```MEMORY_VALUES = [x for x in range(128, 1600, 64)]```

ink-adavison:
Thanks for suggesting that, it is much tidier, yes.

markpeek:
Thanks!

Issue #450, opened 2016-04-04
AWS Lambda Function class doesn't support FunctionName property


3dsorcery:
When creating an AWS Lambda function via troposphere, I need to be able to provide a deterministic name for the function. To do this, I need to specify the FunctionName property, but the current Function class's properties list doesn't support that. I will be adding a pull request with a fix shortly.

terodox:
w00t!  I was wondering what piece of magic I was missing to name the function!

3dsorcery:
The fix for this issue has been merged - Thanks @markpeek!

Bazze:
Anyone knows when this `FunctionName` property in CluodFormation was released by AWS? Can't find any announcement or anything about it, I've been waiting for it for ages!

PR #452, opened 2016-04-04
Add FunctionName to Lambda::Function #450

3dsorcery:
Sorry for the previous pull request fiasco. This should be a bit smoother :)


markpeek:
Awesome. Thanks for the commit.

PR #457, opened 2016-04-08
DomainName isn't a required parameter

### LINKED ISSUE #456: Elasticsearch Domain DomainName shouldn't be required
According to the CF documentation, `DomainName` isn't required: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticsearch-domain.html

mhahn:
Resolves #456


markpeek:
Thanks!

Issue #424, opened 2016-03-08
Configsets support


lony:
When I use the following method
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-init.html#aws-resource-init-configsets

I get  `ValueError('config property is required')` as a workaround I change one of my changesets to config, but is there better way to "validate" Metadata?

phobologic:
Can you give a snippet of the code that you're using that isn't working the way you expect?

lony:
```
import boto3

import troposphere
import troposphere.ec2 as ec2
from troposphere.autoscaling import Metadata
from troposphere.cloudformation import Init, InitFile, InitFiles, InitConfig
from troposphere.iam import InstanceProfile

template = troposphere.Template()
template.add_version("2010-09-09")

param_type = template.add_parameter(troposphere.Parameter(
        'InstanceType',
        Description='Type of EC2 instance',
        Type='String',
))
param_key = template.add_parameter(troposphere.Parameter(
        'KeyName',
        Description='Name of an existing EC2 KeyPair to enable SSH access to the instance',
        Type='String',
))
param_ami = template.add_parameter(troposphere.Parameter(
        'AMI',
        Description='Machine image to use (AMI)',
        Type='String',
))

instance_ec2 = template.add_resource(ec2.Instance(
        "Foo",
        InstanceType=troposphere.Ref(param_type),
        ImageId=troposphere.Ref(param_ami),
        Metadata=Metadata(
                # Usage
                # http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-init.html
                Init({
                    "configSets": {
                        "ascending": ["config1", "config2"]
                    },
                    '**config1**': InitConfig(
                            files=InitFiles({
                                '/var/lib/jenkins/.ssh/config': InitFile(
                                        content=troposphere.Join('\n', [
                                            "# TBD"
                                        ]),
                                        mode='000600',
                                        owner='jenkins',
                                        group='jenkins'),
                                '/var/lib/jenkins/.ssh/jenkins_deploy': InitFile(
                                        content=troposphere.Join('\n', [
                                            "# TBD"
                                        ]),
                                        mode='000600',
                                        owner='jenkins',
                                        group='jenkins'),
                            }),
                    ),
                    'config2': InitConfig(
                            files=InitFiles({
                                '/etc/httpd/conf.d/jenkins-vhost-apache.conf': InitFile(
[...]

Issue #393, opened 2016-02-26
LaunchConfig's Metadata missing Comment field


martinb3:
Hello! I currently have a CF template that has comments inside the metadata of a launch configuration. I'm unable to produce the same template via troposphere:

```
LaunchConfig = t.add_resource(LaunchConfiguration(
    "Foo",
    Metadata=Metadata(Comment="Testing")
    ))
```

Yields `__init__() got an unexpected keyword argument 'Comment'`.

I'm looking for this to be produced (which seems valid on AWS itself):
```
        "LaunchConfig": {
            "Metadata": {
              "Comment": "Foo"
...```
```

Issue #367, opened 2016-01-04
No Example for ParameterGroups


zachzeid:
Would be nice to see if Troposphere supports this, in particular the AWS::CloudFormation::Interface functionality in CloudFormation.

ikben:
Leaving this here, in case someone has time to test and add this to the examples:

````
from troposphere import Template, Parameter

t.add_metadata({
    'AWS::CloudFormation::Interface': {
        'ParameterGroups': [
            {
                'Label': {'default': 'AWS Configuration'},
                'Parameters': ['AccountId']
            },
        ],
        'ParameterLabels': {
            'AccountId': {'default': 'AWS Account Id'},
        }
    }
})

t.add_parameter(Parameter(
    'AccountId',
    Type='Number',
    Default="000000000000",
    Description="The AWS account id"
))

print(t.to_json())
```

IIRC, you can be a bit more clever by defining the parameter first and using `paramter.title` instead of ÀccountId`.

zachzeid:
I've confirmed this works, thanks!

tobiasmcnulty:
I ended up creating a wrapper class for `Template` that lets me define the group and label at the time the parameter is added to the template. This way I don't have to worry about keeping the two in sync. Adding below in case this is useful to anyone else and/or could be incorporated into Troposphere proper some day.

    from collections import OrderedDict

    from troposphere import Template


    class InterfaceTemplate(Template):
        """
        Custom Template class that allows us to optionally define groups and labels for
        CloudFormation Parameters at the time they're added to the template. Groups and
        labels specified, if any, will be added to a custom AWS::CloudFormation::Interface
        """

        def __init__(self, *args, **kwargs):
            super(InterfaceTemplate, self).__init__(*args, **kwargs)
            # use OrderedDict() so we can keep track of the order in which groups are added
            self.parameter_groups = OrderedDict()
            self.parameter_labels = {}
            self.group_order = []

        def add_parameter(self, parameter, group=None, label=None):
            """
            Save group and/or label, if specified, for later generation of
            'AWS::CloudFormation::Interface' in to_dict().
            """
            parameter = super(InterfaceTemplate, self).add_parameter(parameter)
            if group:
                if group not in self.parameter_groups:
                    self.parameter_groups[group] = []
                self.parameter_groups[group].append(parameter.title)
            if label:
                self.parameter_labels[parameter.title] = label
            return parameter

        def set_group_order(self, group_order):
            """
[...]

Issue #298, opened 2015-08-24
Adding Metadata to an existing ec2.Instance instance generates an AttributeError


dlongman:
Using the [example](https://github.com/cloudtools/troposphere/blob/master/examples/CloudFormation_Init_ConfigSet.py) it is possible to set the Metadata for an ec2 instance when it is initialised. However you get an exception if you try to define the Metadata property after the instance has been initialised, for e.g.

```Python
myInstance = ec2.Instance('MyEC2Instance')
myInstance.Metadata = cloudformation.Metadata(...)
``` 
generates the following error:
`AttributeError: AWS::EC2::Instance object does not support attribute Metadata`

All the other parameters in the example (i.e. UserData, KeyName, ImageId, InstanceType, etc) can be set on an existing instance object. 

Is there a reason why setting the Metadata property is not supported in this way? I can set the same Metadata against the Template without error.

markpeek:
Sounds like a bug. At first glance I think it could be fixed by pushing the various check/set of self.attributes into self.__setattr__(). I'll have to look when I have more time. Thanks for reporting it.

markpeek:
@dlongman please try this (quick) [patch](https://github.com/cloudtools/troposphere/commit/7445ab6c638c1345feef2f63cbbcfbc7b804eb8d) to see if it works with your use case.

dlongman:
@markpeek that patch fixes the issue I was seeing, thanks!

dlongman:
sorry closed by mistake - probably should wait until the patch is merged :-)

markpeek:
Merged.

PR #459, opened 2016-04-11
Add AutoScaling Example With Scaling Rules and Notifications

rcuza:
This pull requests creates an example troposphere script that creates the [AutoScalingMultiAZWithNotifications.template](https://s3-us-west-2.amazonaws.com/cloudformation-templates-us-west-2/AutoScalingMultiAZWithNotifications.template).

In order to create a comment that was in the example CloudFormation template, I added a `Comment` class to `troposphere/cloudformation.py`. I am not sure if this is the right place. I was able to create the template when I added the same class to `troposphere/__init__.py`. I did not ultimately choose that location because I do not know if the `Comment` property is universal to all resources in CloudFormation templates.

Perhaps I should of done with the `Comment` property what I had to do with the `NotificationConfigurations` property in the `AutoScalingGroup` resource, namely edit the template to match the troposphere code. I could not create the `NotificationConfigurations` without passing it a list, which is supported by the [AWS AutoScalingGroup documentation](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-group.html).


rcuza:
To make it easier to see how I changed the original Amazon template, here is the diff:
```
(troposphere) $ cat tests/examples_output/AutoScalingMultiAZWithNotifications-orig.template | python -m json.tool  > tests/examples_output/AutoScalingMultiAZWithNotifications-orig-sorted.template
(troposphere) $ diff tests/examples_output/AutoScalingMultiAZWithNotifications-orig-sorted.template tests/examples_output/AutoScalingMultiAZWithNotifications.template
833,841c833,843
<                 "NotificationConfiguration": {
<                     "NotificationTypes": [
<                         "autoscaling:EC2_INSTANCE_LAUNCH",
<                         "autoscaling:EC2_INSTANCE_LAUNCH_ERROR",
<                         "autoscaling:EC2_INSTANCE_TERMINATE",
<                         "autoscaling:EC2_INSTANCE_TERMINATE_ERROR"
<                     ],
<                     "TopicARN": {
<                         "Ref": "NotificationTopic"
---
>                 "NotificationConfigurations": [
>                     {
>                         "NotificationTypes": [
>                             "autoscaling:EC2_INSTANCE_LAUNCH",
>                             "autoscaling:EC2_INSTANCE_LAUNCH_ERROR",
>                             "autoscaling:EC2_INSTANCE_TERMINATE",
>                             "autoscaling:EC2_INSTANCE_TERMINATE_ERROR"
>                         ],
>                         "TopicARN": {
>                             "Ref": "NotificationTopic"
>                         }
843c845
<                 }
---
>                 ]
```

markpeek:
I like the example and think it is good to add it. The issue I'm wrestling with is whether adding Comment would lead to confusion since it will only work in a Metadata attribute and people might try using it in other objects. Let me ponder this one for a bit more.

rcuza:
It looks like `Metadata` is an attribute for every resource (ref: [Adding Comments inside AWS CloudFormation Templates](http://blogs.aws.amazon.com/application-management/post/Tx2CE9T00GRYVHD/Adding-Comments-inside-AWS-CloudFormation-Templates) and [AWS Metadata Attribute doc](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-metadata.html)). *Metadata* is a valid key at the top of an CF template, too.

As you pointed out, every *Metadata* attribute can have a *Comment* key, but *Comment* is not a attribute for every resource. What is the best way to put *Comment* inside of *Metadata*?

When I search the troposphere code base I find *Metadata* in `__init__.py`, `autoscaling.py`, and `cloudformation.py`. For the latter two the `Metadata` class is based on the `AWSHelperFn` class. For `__init__.py`, *Metadata* is just listed as an attribute. Why is there this difference? My first stab at the answer is that while all resources have a *Metadata* attribute, each resource has different keys within that it expects. A map resource with a *Metadata* attribute will not be able to do anything with a "AWS::CloudFormation::Init" key, but a LaunchConfig resource will be able to use it.

Instead of `__init__.py` declaring Metadata as a simple attribute of every *BaseAWSObject*, maybe it should be a class that could be extended for each resource if it needs it. The *Comment* key can be a method of this class or just an attribute. I'm not sure which would be easier to validate or more intuitive to use. Does this accomplish our goal? (yes, I am asking before I actually tried coding it. :scream_cat:)

markpeek:
As you pointed out Metadata is in some of the existing objects. I'll have to spend some time looking at git history to try to figure out why/when those were added. This is sometimes difficult given the lack of historical docs. Given that Metadata is really just a json blob we could remove Comment and just attach and arbitrary dict instead. Or just leave Comment as-is in your PR and see what happens. I'm leaning towards the latter but still thinking about it. I understand your comment about adding Metadata to every BaseAWSObject but wondering if the complexity is worth it for this issue.

rcuza:
I tried adding the *Comment* using a variety of formats around [commit 7a8026d](https://github.com/rcuza/troposphere/commit/7a8026d925b2516c07a0662f36f8f6a1d35196d9#diff-535f2b079bef1e16a0e98aa8f686fd8d), but it wouldn't build or pass tests (I can't remember which).

Whichever way you decide, let me know and I can adjust the PR. I might try a POC for making *Metadata* more robust.

rcuza:
Doing some digging I found the following issues related to *Metadata*:
* **Workaround for missing AWS resource types?** #3 
* **updating InitConfig per AWS docs** #120 
* **Add of Metadata property to the template anatomy?** #286 
* **generates an AttributeError** #298
*  **No Example for ParameterGroups** #367 
*  **LaunchConfig's Metadata missing Comment field** #393  (This matches the problem I had)
*  **Configsets support** #424 
*  **Metadata auth** #155 (This PR adds the `Metadata` class to `autoscaling`)
*  **more metadata options** #153 
* Commit 5e76bf73cfb3c48d321cbdf4f19684ed60965b71 adds the Metadata to `cloudformation.py`.
* Commit 33aedb8ab400d54da326eb5e5e83e9790fd1214c handles CF templates to create Troposphere objects

Unfortunately, this exercise did not give me an opinion on which way to move forward. I'm sharing the links as it may be useful to you.

rcuza:
The failure is in 
```
FAIL: test_template_generator (tests.test_examples_template_generator.test_AutoScalingMultiAZWithNotifications)
```
If there is any interest in this let me know and I'll figure out how to update the tests to match the current code base. Otherwise I'll just close it.

PR #460, opened 2016-04-11
move Groups to property, add policy template version

dpalomin:
Updated IAM_Users_Groups_and_Policies.py example to make it work out of the box.


markpeek:
Thanks!

PR #461, opened 2016-04-12
Fix Elasticsarch Domain object naming

### LINKED ISSUE #455: ElasticSearch Naming
The class for ElasticsearchDomain in the elasticsearch module should maybe be named Domain to match the name in CloudFormation better and prevent stuttering in the API.

That is troposphare.elasticsearch.ElasticsearchDomain becomes troposphare.elasticsearch.Domain 
https://github.com/cloudtools/troposphere/blob/master/troposphere/elasticsearch.py#L58

Another place might be ElasticsearchClusterConfig, however I'm hesitant to rename that. How is the normal naming for this in troposphare?
https://github.com/cloudtools/troposphere/blob/master/troposphere/elasticsearch.py#L41

edubxb:
Fixes #455

### DIFF
--- a/examples/ElasticsearchDomain.py
+++ b/examples/ElasticsearchDomain.py
@@ -2,7 +2,7 @@
 # http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticsearch-domain.html#d0e51519
 
 from troposphere import Template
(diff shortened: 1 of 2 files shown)


markpeek:
Merged. Thanks! BTW, I added a backward compatibility change as well.

edubxb:
:+1:

PR #468, opened 2016-04-19
Import JSON Templates

muikrad:
This new Template subclass makes it possible to import most Cloudformation templates as JSON and parse them as Troposphere objects. This is different from cfn2py in that it's not trying to write actual python code, but rather instantiate the actual objects without the need to edit it afterwards.

There are probably a few tropo-specific AWSHelperFns that won't be parsed correctly, but the ones included in the tests/examples_output are all supported. I believe an effort should be done to normalize the helper functions into something discoverable with a consistent use - but I have yet to really ponder on that issue.


markpeek:
Thanks!

PR #469, opened 2016-04-21
Fix Source object to take a list of SourceDetails

JuanCanham:
The documentation is unclear but the example (and my testing) shows that this is the correct syntax

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-config-configrule.html#cfn-config-configrule-source

### DIFF
--- a/troposphere/config.py
+++ b/troposphere/config.py
@@ -26,7 +26,7 @@ class SourceDetails(AWSProperty):
 class Source(AWSProperty):
     props = {
         'Owner': (basestring, True),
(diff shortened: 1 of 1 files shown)


rcuza:
The documentation does in deed say that the SourceDetails property should be a list. This change appears to make the change. 

I am not an expert on python tests, but I don't see a test for this file.

:+1: with the understanding that I did not perform a functional test.

markpeek:
Thanks!

PR #470, opened 2016-04-22
Update Contribute Document to Use Requirements.txt

rcuza:
Made a minor simplification change to the `CONTRIBUTE.md` document. Instead of listing all the commands for installing modules, I put the modules into a `requirements.txt` file (via `pip freeze`) and then updated the docs on how to use the new file.

This change should help with ~~code~~ documentation rot.

One possible improvement is to use `>=` on the versioning, but would only want to do that if the CI tests always used the bleeding edge version.


philtay:
I think the dev dependencies listed in the docs/requirements.txt file should not be pinned to specific versions.

rcuza:
@philtay - Great point. I removed the version pinning.

markpeek:
Thanks!

PR #472, opened 2016-04-25
Implement LifecycleRule Transitions property

muikrad:
This adds support for the new 'Transitions' property.

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-lifecycleconfig-rule.html

AWS deprecated 'Transition' so I added some forward-compatibility support just in case.


markpeek:
Thanks!

PR #473, opened 2016-04-26
Better AWSHelperFn support in template generator

muikrad:
The template generator's fallback try/catch was a bit too focused; it should always be used as a last resort.

This fixes the "Tags" case where args is a mapping and gets called with a single *unpacking star.


markpeek:
Thanks!

PR #475, opened 2016-04-26
Fix Bucket AccessControl to allow Ref

muikrad:
The current implementation will break when attempting to "Ref" the AccessControl to a stack parameter.


markpeek:
Thanks!

PR #478, opened 2016-04-27
Convert fake AWSHelperFns into AWSProperties

jonapich:
This updates 3 objects that were incorrectly subclassing AWSHelperFn. They are now subclassing AWSProperty as intended. This should be backward compatible.

### DIFF
--- a/troposphere/ec2.py
+++ b/troposphere/ec2.py
@@ -3,7 +3,7 @@
 #
 # See LICENSE file for full license.
 
-from . import AWSHelperFn, AWSObject, AWSProperty
+from . import AWSObject, AWSProperty
 from .validators import (
     boolean, integer, integer_range, network_port, positive_integer
 )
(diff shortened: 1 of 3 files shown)


muikrad:
The remaining AwsHelperFn objects that don't represent a real AWS Intrinsinc Function are very problematic and I haven't found a way to convert them in a backward-compatible manner.

I got really close with the "Tags" object provided in troposphere's __init__. It cannot be translated into an AWSProperty because it is a list of tags, making the props = {} impossible to write. I wrote a new AWSProperty called "Tag" with a proper 'props' object (Key/Value as required basestrings), and then changing Tags for a subclass of list that internally creates Tag objects and acts as a list of such. Then I changed a lot of the 'Tags: (list, False)' into 'Tags: ([Tag], False)' and got clean test runs, but this introduces a breaking change if anyone actually passed a list of {Key: k, Value:v} dictionaries because the inner types are now checked. Before, it only made sure it was a list.

The other major problem will be the Metadata elements from the cloudformation module. I never worked with them but I believe they don't have a set structure like typical properties. For those, maybe it should be better to just wrap them into a new no-validation object that would be separate from AWSHelperFn, instead of trying to wrap them into actual AWSProperty objects...?

muikrad:
The Tags object's breaking change could potentially be non-breaking by implementing multiple-type checks?

```
props = {'Tags': (([(Tag, dict)]), False)
```

markpeek:
@jonapich regardless the issue with Tags, is the rest of this PR good to go?

jonapich:
@markpeek yep!

markpeek:
Thanks!

Issue #344, opened 2015-11-04
ValueError('Name "%s" not alphanumeric' % title) on resource which allows non-Alphanumeric characters in name


ocsi01:
I just faced the issue, that I'm not allowed to give my custom name to a Cloudwatch Log-Group:
"tomcat/error.log".
AWS allows the name above for a Log-Group, however Troposphere not.

markpeek:
I suspect this is one of the FAQ we get about CloudFormation usage. The CloudFormation resource name is very strict and not the same as names used within the resource. An example illustrating this is #311. But feel free to post a snippet of your log group creation and we might be able to give you more concrete advice.

ikben:
This is more a limitation of CloudFormation, not of Troposphere.

There is no 'LogGroupName' property in the AWS::Logs::LogGroup resource (http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-loggroup.html), and like @markpeek said, resource names must be alphanumeric. In this case it's not possible to use a non-alphanumeric name.

If you really need the alphanumeric name, you can use a Custom Resource, or generatie the log group with the API the first time you create a log stream.

markpeek:
Closing based on previous comments.

Issue #311, opened 2015-09-07
S3 Bucket - valid_names regex


jrasell:
The regex used for validating S3 bucket names omits a number of valid S3 bucket names:

```
valid_names = re.compile(r'^[a-zA-Z0-9]+$')
```
```
    raise ValueError('Name "%s" not alphanumeric' % title)
ValueError: Name "jrasell-internal" not alphanumeric
```

It would be handy if it could at least support dashes.

ikben:
There is a difference between bucketName and the name of the resource.

This works:

    bucket = s3.Bucket('NoHyphens', BucketName='with-hyphens')

This doesn't:

    bucket = s3.Bucket('Invalid-Hyphen', BucketName='valid-hyphen').

jrasell:
My bad.
Thanks very much for the help @ikben.

bjinwright:
This should be added to the S3 example

PR #483, opened 2016-05-04
Leverage S3 bucket name validation

bryanvh:
The `Bucket` class extends from a base class `BaseAWSObject` that validates (during object initialization) the resource title using a regular expression that is too strict for S3 buckets. For example, the name "my-bucket" is perfectly valid, however Troposphere reports an error.

```
Name "my-bucket" not alphanumeric: ValueError
...
  File "/var/task/troposphere/s3.py", line 222, in __init__
    super(Bucket, self).__init__(name, **kwargs)
  File "/var/task/troposphere/__init__.py", line 44, in __init__
    self.validate_title()
  File "/var/task/troposphere/__init__.py", line 159, in validate_title
    raise ValueError('Name "%s" not alphanumeric' % self.title)
ValueError: Name "my-bucket" not alphanumeric
```

Adding a `validate_title()` method to the `Bucket` class overrides this with correct behavior that leverages the existing `s3_bucket_name()` validator.

This correction caused a number of test cases to break, as they were improperly using a bucket name with capital letters. These test cases have been fixed. I also added more test cases to catch situations where the bucket name starts or ends with a period, or contains consecutive periods.
http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html

Finally, it seems pyflakes 1.2.0 does not work, so I pinned the version to 1.1.0 to help Travis CI pass... not sure if there was a more appropriate solution to that problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_install:
   - pip install --upgrade pip setuptools wheel
 
 install:
-  - pip install --upgrade pep8 pyflakes
+  - pip install --upgrade pep8 pyflakes==1.1.0
 
 before_script:
   - if [[ $TRAVIS_PYTHON_VERSION == 3* ]]; then 2to3 -n -w --no-diffs troposphere examples/CustomResource.py; fi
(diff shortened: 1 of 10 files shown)


markpeek:
Thank you for the change but the title (used by CloudFormation) is different from the S3 bucket name. Closing dup of #311 and #344.

bryanvh:
Ahh, I see your point. I believe the validation logic could be better, so perhaps I will submit a new PR just for that. Thanks!

PR #485, opened 2016-05-05
Validate the template against aws before stack creation

chrisgilmerproj:
Working with @markfalk discovered its helpful to have AWS validate the template before stack creation. This change includes these updates while maintaining backwards compatibility:

- Can now use -u or --update to trigger a cfn update action instead of a create action
- Fixes pep8 trailing whitespace
- Fixes pep8 line length to be < 80 chars

### DIFF
--- a/scripts/cfn
+++ b/scripts/cfn
@@ -31,6 +31,23 @@ def upload_template_to_s3(conn, region, bucket_name, key_name, template):
     return cfn_template_url
 
 
+def validate_template(conn, template=None, url=None):
+    """Validate the template"""
+    print("Validating Template")
+    try:
+        if url:
+            conn.validate_template(template_url=url)
+        else:
+            conn.validate_template(template_body=template)
+    except boto.exception.BotoServerError as e:
+        # XXX - need to figure out why this isn't getting parsed from boto.
+        print("Error: %s" %
+              (literal_eval(e.error_message)['Error']['Message'],))
+        print("Exiting...")
+        sys.exit(1)
+    print("Template Validated")
+
+
 def create_stack(
     conn,
     stackname,
@@ -45,10 +62,11 @@ def create_stack(
         parameters=params,
         capabilities=capabilities,
     )
-    
+    validate_template(conn, template=template, url=url)
+
     try:
         stack_id = conn.create_stack(stackname, **kwargs)
-    
+
     except boto.exception.BotoServerError as e:
         # XXX - need to figure out why this isn't getting parsed from boto.
         print("Error: %s" %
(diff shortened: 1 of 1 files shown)


chrisgilmerproj:
Friendly ping - Hoping to get a review on this PR.

scripts/cfn
```diff
@@ -23,14 +23,32 @@ def upload_template_to_s3(conn, region, bucket_name, key_name, template):
...
+            conn.validate_template(template_url=url)
+        else:
+            conn.validate_template(template_body=template)
+    except boto.exception.BotoServerError as e:
+        # XXX - need to figure out why this isn't getting parsed from boto.
+        print("Error: %s" %
+              (literal_eval(e.error_message)['Error']['Message'],))
+        print("Exiting...")
+        sys.exit(1)
+    print("Template Validated")
```
rcuza: What happens if both `template` and `url` are `None`? Wont this return "Template Validated"? This does not seem like a desired result.

What about returning `True` or `False` would make the function more flexible? The code that calls it could then decide what to do in either case.

chrisgilmerproj: I like these ideas, I'll update this and get it back to you.

chrisgilmerproj: Anything that validates returns a `boto.cloudformation.template.Template` object.  If it does not validate it will throw an error:

```py
BotoServerError: BotoServerError: 400 Bad Request
<ErrorResponse xmlns="http://cloudformation.amazonaws.com/doc/2010-05-15/">
  <Error>
    <Type>Sender</Type>
    <Code>ValidationError</Code>
    <Message>Template format error: unsupported type or structure. (line 2, column 5)</Message>
  </Error>
  <RequestId>aacaaf5b-2442-11e6-96bc-4793db5d2621</RequestId>
</ErrorResponse>
```

The style of this script is that any error causes a fault.  We could cause this method to return `True` or `False`, but it steps away from the style of the other methods.  Your call, I'm happy to implement either.

chrisgilmerproj: Also, I just updated to cause an error if neither `url` or `template` are provided.

rcuza: Good point about not wanting to loose the error message. 

Out of scope for this PR, but there is something wrong with the way the `logging` feature and `print` is used in this script over all, but I'm not sure what it is.

rcuza:
These changes LGTM. 👍

chrisgilmerproj:
Fantastic.  I can't merge this so I'll leave it up to your team to do so.  Thanks for the review!

markpeek:
Thanks for the PR!

@rcuza thanks for the code review feedback

PR #486, opened 2016-05-05
Tune validation logic and test cases for S3 bucket names

bryanvh:
Outline of changes:
- added validation to disallow consecutive periods in bucket name
- added validation to disallow bucket names that look like an IP address
- updated `Bucket` initialization to clarify resource title vs. bucket name, to help avoid more confusion about that in the future
- updated test cases to catch bucket names with capital letters, starting/ending with a period, containing consecutive periods, use of IP addresses

Also, it seems pyflakes 1.2.0 does not work, so I pinned the version to 1.1.0 to help Travis CI pass... not sure if there was a more appropriate solution to that problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_install:
   - pip install --upgrade pip setuptools wheel
 
 install:
-  - pip install --upgrade pep8 pyflakes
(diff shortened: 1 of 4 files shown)


markpeek:
Thanks!

PR #491, opened 2016-05-11
Allow setting a different region for S3 upload

chrisgilmerproj:
It would be nice if uploading to S3 could be set to a different region than the stack that is being created. This adds a single option to so so with either `-s` or `--s3-region` and maintains backwards compatibility.

### DIFF
--- a/scripts/cfn
+++ b/scripts/cfn
@@ -112,7 +112,7 @@ if __name__ == "__main__":
                         help="stack parameters in key=value form")
     parser.add_argument("-r", "--region", default="us-east-1",
                         help="Region (default %(default)s)")
-    parser.add_argument("-s", "--s3-region", default="us-east-1",
+    parser.add_argument("-s", "--s3-region", default=None,
                         dest="s3region",
                         help="S3 Region (default %(default)s)")
     parser.add_argument("-R", "--resources", action='store_true',
@@ -138,6 +138,9 @@ if __name__ == "__main__":
     if values.debug:
         import logging
         logging.basicConfig(filename="boto.log", level=logging.DEBUG)
+
+    if not values.s3region:
+        values.s3region = values.region
 
     conn = boto.cloudformation.connect_to_region(values.region)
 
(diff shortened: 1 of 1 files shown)


scripts/cfn
```diff
@@ -131,7 +134,11 @@ if __name__ == "__main__":
 
         if values.s3bucket:
             # Upload to S3 and create the stack
-            s3conn = boto.s3.connect_to_region(values.region)
+            s3conn = None
```
markpeek: I don't believe this is needed.

scripts/cfn
```diff
@@ -131,7 +134,11 @@ if __name__ == "__main__":
...
+            if values.s3region != values.region:
```
markpeek: Would this be better stated as ```if values.s3region is not None:```?

scripts/cfn
chrisgilmerproj: My linter complained :( It's not necessary but its apparently more correct.  Let me know if you'd like me to remove it.

scripts/cfn
chrisgilmerproj: I set the default to be `us-east-1`.  Thinking about it, I could just forgo this entire statement and just change the original to:

```py
s3conn = boto.s3.connect_to_region(values.s3region)
```

What do you think?  If you'd prefer it to be `None` as a default I can update that as well.

chrisgilmerproj:
Take a look now, I think this clears it up.

markpeek:
The concern I had was existing usage of the command where the s3region was assumed to be the same as the region. With your change you have to set both. It might be better to default s3region to None in the  add_argument() call and then set s3region to region if s3region is None. In other words, I liked your previous version better but had just commented on slight tweaks to it. Thanks!

chrisgilmerproj:
@markpeek - I see what you we're saying.  I've made the change you suggested and also updated the help text to read correctly.

I merged this with the latest master so it ought to pass tests.  If you need me to rebase it into one commit please let me know but I think its ready for your sign off.

markpeek:
Thanks!

chrisgilmerproj:
Thank you!

PR #492, opened 2016-05-12
fix attribute for ApiKey (Enable -> Enabled)

justone:
Extended the test too.


markpeek:
Thanks for the change and also for extending the test.

PR #493, opened 2016-05-12
Invoke join correctly

mhahn:
This is python, not javascript 😉


phobologic:
Don't program while drinking I guess? :)  Thanks.

Issue #495, opened 2016-05-18
Role titles with non-alphanumeric values


viswaug:
Troposphere is awesome! Thank you everyone for working on it.

I was able to create a Role with title value containing "-" i.e. alphanumeric. This passes troposphere validation, but fails when running the template. Has anyone else seen this behavior?

markpeek:
Thanks for the issue. Can you post a small sample code showing this behavior?

viswaug:
Sure thing. I just replaced the title of the role in the Lambda sample with "Lambda-Execution-Role" like here

```
from troposphere.constants import NUMBER
from troposphere import FindInMap, GetAtt, Join, Output
from troposphere import Parameter, Ref, Template
from troposphere.awslambda import Function, Code, MEMORY_VALUES
from troposphere.cloudformation import CustomResource
from troposphere.ec2 import Instance
from troposphere.ec2 import SecurityGroup
from troposphere.iam import Role, Policy


t = Template()

t.add_version("2010-09-09")

ExistingVPC = t.add_parameter(Parameter(
    "ExistingVPC",
    Type="AWS::EC2::VPC::Id",
    Description=(
        "The VPC ID that includes the security groups in the "
        "ExistingSecurityGroups parameter."
    ),
))

InstanceType = t.add_parameter(Parameter(
    "InstanceType",
    Default="t2.micro",
    Type="String",
    AllowedValues=["t2.micro", "m1.small"],
))

ExistingSecurityGroups = t.add_parameter(Parameter(
    "ExistingSecurityGroups",
    Type="List<AWS::EC2::SecurityGroup::Id>",
))

MemorySize = t.add_parameter(Parameter(
    'LambdaMemorySize',
    Type=NUMBER,
    Description='Amount of memory to allocate to the Lambda Function',
    Default='128',
    AllowedValues=MEMORY_VALUES
))

Timeout = t.add_parameter(Parameter(
    'LambdaTimeout',
    Type=NUMBER,
    Description='Timeout in seconds for the Lambda function',
    Default='60'
))

t.add_mapping("AWSInstanceType2Arch",
              {u'm1.small': {u'Arch': u'PV64'},
               u't2.micro': {u'Arch': u'HVM64'}}
              )

t.add_mapping("AWSRegionArch2AMI",
              {u'ap-northeast-1': {u'HVM64': u'ami-cbf90ecb',
                                   u'PV64': u'ami-27f90e27'},
[...]

PR #499, opened 2016-05-28
Validate that iam role name is alphanumeric

mbrannigan:
Fixes [#495](https://github.com/cloudtools/troposphere/issues/495)

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -107,6 +107,9 @@ class TestValidators(unittest.TestCase):
         for s in ['a'*65, 'a'*128]:
             with self.assertRaises(ValueError):
                 iam_role_name(s)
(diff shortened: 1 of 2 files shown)


ikben:
Can we use the regex from http://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateRole.html ?

`[\w+=,.@-]+` Is the regex AWS documents as valid

ikben:
I'm going to add a comment on #495 to propose a better fix. Your code does prevent adding a wrong resource name for an IAM::Role, but I think it may be confusing to do that in the `iam_role` validator, especially now that we can set the Role Name in CloudFormation.

mbrannigan:
Cool.  That's great that RoleName and GroupName exists.

Issue #465, opened 2016-04-18
Why does the UserData helper join the results with ","?


philipbjorge:
We tried to use the helper defined [here](https://github.com/cloudtools/troposphere/blob/master/troposphere/helpers/userdata.py) to load a shell script.
When we deployed the template, we were a bit surprised to see that the bash script failed to run because commas had been prepended to every line...

```
,
,service tyk-dashboard restart
,service tyk-gateway restart
```

Why does the UserData helper join with a `,`?

markpeek:
This was code added from a pull request. The use case seemed to want that behavior. You might want to comment back to #376 to see what the authors intention was with the change.

PR #502, opened 2016-06-06
Re-written userdata from_file helper

j-carl:
Rewritten the helper to be more flexible, what kind of delimiter you want to have. The default is now "", which reflects the most used use-case. It's also addresses the issue #465.


markpeek:
Thanks!

PR #504, opened 2016-06-09
Add undocumented ACM Certificate properties for ViewerCertificate

chrisgilmerproj:
I'm doing some testing, but this appears to be an undocumented feature of the ViewerCertificate.  I have these resources to back me up from the Ruby and Java SDKs:

- [Ruby SDK](http://docs.aws.amazon.com/sdkforruby/api/Aws/CloudFront/Types/ViewerCertificate.html)
- [Java SDK](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/cloudfront/model/ViewerCertificate.html)

For reference, here are the docs I was using originally where it is not mentioned:

[CloudFront DistributionConfiguration ViewerCertificate](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distributionconfig-viewercertificate.html#cfn-cloudfront-distributionconfig-viewercertificate-iamcertificateid)

### DIFF
--- a/troposphere/cloudfront.py
+++ b/troposphere/cloudfront.py
@@ -113,8 +113,11 @@ class ViewerCertificate(AWSProperty):
     props = {
         'CloudFrontDefaultCertificate': (boolean, False),
         'IamCertificateId': (basestring, False),
+        'ACMCertificateArn': (basestring, False),
         'MinimumProtocolVersion': (basestring, False),
         'SslSupportMethod': (basestring, False),
+        'Certificate': (basestring, False),
+        'CertificateSource': (basestring, False),
     }
 
 


chrisgilmerproj:
Some more info here, apparently it works in some regions and not in others:

https://forums.aws.amazon.com/thread.jspa?messageID=723396

chrisgilmerproj:
I've tested and at least CloudFront doesn't recognize any of these parameters, so this PR might be a little early given the announcement only came out in January:

https://aws.amazon.com/about-aws/whats-new/2016/01/amazon-cloudfront-integrates-with-aws-certificate-manager/

chrisgilmerproj:
Yep, looks like its not ready so people use workarounds - http://stackoverflow.com/questions/35914161/how-to-provision-a-cloudfront-distribution-with-an-acm-certificate-using-cloud-f

I'll come back to this once CloudFront supports it.

markpeek:
Yes, this appears to be due to lag between features in CloudFront getting implemented in CloudFormation. Closing for now until it gets implemented in CloudFormation.

chrisgilmerproj:
Sounds good to me, thanks.

PR #505, opened 2016-06-10
added support for Kinesis Firehose

drmaples:
- https://aws.amazon.com/about-aws/whats-new/2016/06/aws-cloudformation-adds-support-for-amazon-vpc-flow-logs-amazon-kinesis-firehose-streams-and-other-updates/
- http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-kinesisfirehose-deliverystream.html

### DIFF
--- a/troposphere/firehose.py
+++ b/troposphere/firehose.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2014, Guillem Anguera <ganguera@gmail.com>
+# Copyright (c) 2016, troposphere project
 # All rights reserved.
 #
 # See LICENSE file for full license.
--- a/troposphere/firehose.py
+++ b/troposphere/firehose.py
@@ -1,0 +1,111 @@
+# Copyright (c) 2014, Guillem Anguera <ganguera@gmail.com>
+# All rights reserved.
+#
+# See LICENSE file for full license.
+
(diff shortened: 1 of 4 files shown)


drmaples:
anything else i need to do here? i also tested it works by spinning up a kinesis firehose that loads data to redshift.

troposphere/firehose.py
```diff
@@ -0,0 +1,111 @@
+# Copyright (c) 2014, Guillem Anguera <ganguera@gmail.com>
```
markpeek: Is this copyright correct? I'm guessing you copied an existing file but then modified the entire contents? Shouldn't this have 2016 and your name in it?

drmaples: should i chop the whole thing? I personally dont care to have my info here, its in the commit logs if anyone cares.

markpeek: If you don't care could you switch it to (for lack of a better one):
# Copyright (c) 2016, troposphere project

I have no plans to ever change the licensing but I do care about keeping copyrights intact in case there is any issues. Thanks!

markpeek:
Thanks!

drmaples:
@markpeek any idea when this will make it to pypi?

PR #516, opened 2016-06-20
Emr configurations values: also allow AWS helper functions

cbamelis:
The current master branch contains a too strict EMR configurations validator method: this validator forces configuration values to be strings.  CloudFormation also allows helper functions (which allow you to use for instance a parameter value as configuration value instead of a direct string value).

This branch contains an updated validator method and updated EMR cluster example using an indirect configuration value.

I also updated the EMR example with a parameter dependent choice between spot instances or on-demand instances (using a 'condition').

The EMR example is successfully created in my AWS region eu-west-1.


markpeek:
Thanks!

PR #517, opened 2016-06-23
Handle all of the dynamodb types

mhahn:
http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_AttributeValue.html


mhahn:
? i guess you can only set these types in cloudformation: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-dynamodb-attributedef.html#cfn-dynamodb-attributedef-attributename

PR #518, opened 2016-06-23
Fix AssociationParameters Value type to list of strings

dmotles:
According to
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance-ssmassociations-associationparameters.html,
the SsmAssociations AssociationParameters Value is a list of strings,
not a string. This was causing instance creation failure on templates
attempting to use SsmAssociations.


markpeek:
Thanks!

PR #519, opened 2016-06-24
Add DependsOn to Deployment and remove Enabled from StageKey

therockstorm:
DependsOn is required if the RestApi and Method are created in the same template as explained here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-deployment.html#aws-resource-apigateway-deployment-examples


markpeek:
This change requires a corresponding update to tests/examples_output/ApiGateway.template in order to pass the tests.

markpeek:
Thanks!

PR #520, opened 2016-06-29
Removed the DBSubnetGroupName limitation on rds replica

JacobMMC:
DBSubnetGroupName is not a limitation of a RDS replica.
Previous Troposphere version had a limitation for when SourceDBInstanceIdentifier parameter was defined in a DBInstance.

AWS documentation has no reference of this limitation and it's the only way to create an rds replica on a custom VPC instead of in the default one.


markpeek:
Thanks. Dup of #515 which has been merged.

PR #524, opened 2016-07-06
Allows MultiAZ=false with AvailabilityZone in rds

### LINKED ISSUE #523: AvailabilityZone/MultiAZ Validation broken for RDS
```
        avail_zone = self.properties.get('AvailabilityZone', None)
        multi_az = self.properties.get('MultiAZ', None)
        if not (isinstance(avail_zone, (AWSHelperFn, nonetype)) and
                isinstance(multi_az, (AWSHelperFn, nonetype))):
            if 'AvailabilityZone' in self.properties and \
                    self.properties.get('MultiAZ', None):
                raise ValueError("AvailabiltyZone cannot be set on "
                                 "DBInstance if MultiAZ is set to true.")
```

The issue here is that you should be able to set `AvailabilityZone` to something, and at the same time set `MultiAZ` to `False`, but with that code you couldn't currently.

phobologic:
Fixes #523

### DIFF
--- a/tests/test_rds.py
+++ b/tests/test_rds.py
@@ -2,6 +2,8 @@ import unittest
 import troposphere.rds as rds
 from troposphere import If, Parameter, Ref
 
+AWS_NO_VALUE = "AWS::NoValue"
+
 
 class TestRDS(unittest.TestCase):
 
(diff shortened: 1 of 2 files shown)


markpeek:
Thanks!

PR #525, opened 2016-07-07
do not require Status as a property in iam.AccessKey

PartyImp:
Greetings,

Thank you for this excellent software. Per the following documentation:

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-iam-accesskey.html

'Status' is not a required property of IAM::AccessKey. I do not know what examples the OP was referring to, but IAM::AccessKey does actually default to 'Active'

Thanks!
Niko


markpeek:
Thanks!

PR #529, opened 2016-07-21
Support "UserName" property for AWS::IAM::User

dmm92:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-iam-user.html#cfn-iam-user-username

### DIFF
--- a/troposphere/iam.py
+++ b/troposphere/iam.py
@@ -5,7 +5,7 @@
 
 from . import AWSObject, AWSProperty
 from .validators import integer, boolean, status
-from .validators import iam_path, iam_role_name, iam_group_name
+from .validators import iam_path, iam_role_name, iam_group_name, iam_user_name
 
 try:
     from awacs.aws import Policy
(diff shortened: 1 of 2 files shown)


markpeek:
This looks good. It would be good to validate the length not being the empty string explicitly. Also, some positive/negative tests for the length and regex would be a welcome addition.

dmm92:
@markpeek is troposphere moving towards doing its own validation? In most places it is 'dumb' and has AWS do the validation. I was hesitant to even add validation as a result — what's the pattern? Is there a plan to gradually add validators for everything? If not I think it would be better to leave them out entirely

markpeek:
Besides being able to write templates in Python, the original intent for troposphere was to do as much validation at template generation time as possible to prevent having CF rollbacks from occurring. You're right, not everything is validated but there are both base and custom validators available and I'd welcome people contributing more validation. To the point of my comment, since you had some validation (length of string, regex) I thought it would be good to catch an empty string and also add tests to ensure the correct behavior was occurring.

markpeek:
Thanks. I added the empty string check and some unit tests in the above commit.

dmm92:
@markpeek Thanks! Sorry I didn't get to this before I went on vacation

PR #531, opened 2016-07-26
Remove double S from S3ObjectVersion (fixes #530)

### LINKED ISSUE #530: S3ObjectVersion is spelled "SS3ObjectVersion" in the lambda Code object validation
I just noticed [this](https://github.com/cloudtools/troposphere/blob/1f67fb140f5b94cf0f29213a7300bad3ea046a0f/troposphere/awslambda.py#L31) while I was reading through the code. I haven't run into problems as I haven't had to use this particular key, but it looks like something you might want to know about.

ikben:
Like @sonya noted in #530


markpeek:
Thanks!

PR #533, opened 2016-07-27
Fix TemplateGenerator import logic.

### LINKED ISSUE #532: Template Generator is broken.
When attempting to use the template generator on well defined inputs:

```
import json

from troposphere.template_generator import TemplateGenerator
dictionary = json.loads("""{
  "Resources": {
      "Type": "AWS::S3::Bucket",
      "Properties": {
        "BucketName": "Bob"
      }
  }
}""")
TemplateGenerator(dictionary)
```

yields the following error

```
AttributeError                            Traceback (most recent call last)
<ipython-input-7-a7d4c9f1b746> in <module>()
----> 1 template_generator.TemplateGenerator(json.loads(x))

/usr/lib/python2.7/site-packages/troposphere/template_generator.pyc in __init__(self, cf_template)
     57             self.add_condition(k, self._convert_definition(v, k))
     58         for k, v in cf_template.get('Resources', {}).iteritems():
---> 59             self.add_resource(self._convert_definition(v, k))
     60         for k, v in cf_template.get('Outputs', {}).iteritems():
     61             self.add_output(self._create_instance(Output, v, k))

/usr/lib/python2.7/site-packages/troposphere/__init__.pyc in add_resource(self, resource)
    450 
    451     def add_resource(self, resource):
--> 452         return self._update(self.resources, resource)
    453 
    454     def add_version(self, version=None):

/usr/lib/python2.7/site-packages/troposphere/__init__.pyc in _update(self, d, values)
    435                 d[v.title] = v
    436         else:
--> 437             if values.title in d:
    438                 self.handle_duplicate_key(values.title)
    439             d[values.title] = values

AttributeError: 'dict' object has no attribute 'title'
```

The reason for this is as follows:

`template_generator.py` incorrectly invokes `pkgutil.walk_packages` here:

https://github.com/cloudtools/troposphere/blob/master/troposphere/template_generator.py#L313

https://docs.python.org/3/library/pkgutil.html

`walk_packages` expects a `list` rather than a `string`, of file paths. This error is not caught by the tests because if a string is passed in, then the entire string is treated as a list of paths. The only valid path that does anything will be the `'.'` in `'template_generator.py'`, resulting in the current directory being searched for directories named `'troposphere'`. Obviously, in the test environment, it will find such a package and successfully include all the necessary modules needed to pass the tests.

cwgreene:
template_generator.py was incorrectly invoking pkgutil.walk_packages,
passing in the filepath directly rather than as a list of paths. By
coincidence, this resulted in the current directory being processed
instead, which allowed the tests to accidentally pass.

Fixes #532

### DIFF
--- a/troposphere/template_generator.py
+++ b/troposphere/template_generator.py
@@ -16,6 +16,7 @@ Usage:
 import inspect
 import pkgutil
 import importlib
+import os
 
 from collections import Sequence, Mapping
 
@@ -307,10 +308,11 @@ class TemplateGenerator(Template):
 
     def _import_all_troposphere_modules(self):
         """ Imports all troposphere modules and returns them """
+        dirname = os.path.join(os.path.dirname(__file__), "..")
         module_names = [
             pkg_name
             for importer, pkg_name, is_pkg in
-            pkgutil.walk_packages(__file__)
+            pkgutil.walk_packages([dirname])
             if not is_pkg and pkg_name.startswith("troposphere") and
             pkg_name not in self.DEPRECATED_MODULES]
         module_names.append('troposphere')
--- a/troposphere/template_generator.py
+++ b/troposphere/template_generator.py
@@ -313,8 +313,7 @@ class TemplateGenerator(Template):
             pkg_name
             for importer, pkg_name, is_pkg in
             pkgutil.walk_packages([dirname], prefix="troposphere.")
-            if not is_pkg and pkg_name.startswith("troposphere") and
-            pkg_name not in self.DEPRECATED_MODULES]
+            if not is_pkg and pkg_name not in self.DEPRECATED_MODULES]
         module_names.append('troposphere')
 
         modules = []
--- a/troposphere/template_generator.py
+++ b/troposphere/template_generator.py
@@ -308,11 +308,11 @@ class TemplateGenerator(Template):
 
     def _import_all_troposphere_modules(self):
         """ Imports all troposphere modules and returns them """
-        dirname = os.path.join(os.path.dirname(__file__), "..")
+        dirname = os.path.join(os.path.dirname(__file__))
         module_names = [
             pkg_name
             for importer, pkg_name, is_pkg in
-            pkgutil.walk_packages([dirname])
+            pkgutil.walk_packages([dirname], prefix="troposphere.")
             if not is_pkg and pkg_name.startswith("troposphere") and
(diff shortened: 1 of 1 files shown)


cwgreene:
So this change is a lot cleaner, it only walks over the troposphere directory. I found that when it was walking over other directories it seemed to start executing logic unexpectedly.

troposphere/template_generator.py
```diff
@@ -307,10 +308,11 @@ def _get_function_type(self, function_name):
 
     def _import_all_troposphere_modules(self):
         """ Imports all troposphere modules and returns them """
+        dirname = os.path.join(os.path.dirname(__file__))
         module_names = [
             pkg_name
             for importer, pkg_name, is_pkg in
-            pkgutil.walk_packages(__file__)
+            pkgutil.walk_packages([dirname], prefix="troposphere.")
```
markpeek: Is the ```pkg_name.startswith("troposhere")``` redundant with this ```prefix=``` change?

cwgreene: Ah, good point; Yes I believe that this can now be removed; I'll try to update this pr later tonight.

cwgreene: Removed!

cwgreene: @markpeek, is there any additional changes you'd like me to make?

markpeek: @cwgreene I looked at this earlier today and did have a question I was going to pose after I had time looking at the walk_packages interface. But I'll just ask it. I thought the ```prefix="troposphere"``` was added to limit the walk. So I would have thought keeping it and removing the ```pkg_name.startswith``` was the change you were going to make and the main change in this PR. But you removed the prefix and kept the startswith. Just wanted to verify that is what you intended to do.

cwgreene: Sorry, something seems to have gone wrong, I thought this had updated, but that does not seem to have happened.

Prefix needs to be kept, since we start by iterating over the troposphere package. Otherwise, the DEPRECATED_PACKAGES won't work as expected.

You are correct that pkg_name.startswith can be removed, because we have forced everything to start with troposphere.

I have uploaded this now.

cwgreene: Oh, the other reason for prefix is so that the import works correctly. If it wasn't prefixed, import would recieve "iam" rather than "troposphere.iam" and fail to import.

markpeek:
Thanks!

PR #535, opened 2016-07-30
Add Name attributes for IAM groups and roles

pinkieee:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-iam-group.html#cfn-iam-group-groupname
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html#cfn-iam-role-rolename


markpeek:
Thanks!

PR #540, opened 2016-08-10
CloudFront updates (Aug 9, 2016)

philtay:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/ReleaseHistory.html

### DIFF
--- a/troposphere/cloudfront.py
+++ b/troposphere/cloudfront.py
@@ -120,6 +120,7 @@ class Restrictions(AWSProperty):
 
 class ViewerCertificate(AWSProperty):
     props = {
+        'AcmCertificateArn': (basestring, False),
         'CloudFrontDefaultCertificate': (boolean, False),
         'IamCertificateId': (basestring, False),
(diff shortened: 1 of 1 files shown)


troposphere/cloudfront.py
```diff
@@ -111,6 +120,7 @@ class Restrictions(AWSProperty):
 
 class ViewerCertificate(AWSProperty):
     props = {
+        'AcmCertificateArn': (basestring, False),
```
ikben: It would be nice to have a test that checks that only one of `AcmCertificateArn`, `CloudFrontDefaultCertificate`, or `IamCertificateId`  is used.

markpeek:
@philtay let me know if you want to add the suggested check or whether I should merge as-is.

philtay:
@markpeek I think it should be merged as-is. In some modules (e.g. `ec2.py` and `waf.py`) the conditional attributes have the Python comment `# Conditional`. Do you want me to add the comment?

markpeek:
Thanks! Looks fine as-is.

PR #541, opened 2016-08-10
EFS updates (Aug 9, 2016)

philtay:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/ReleaseHistory.html


markpeek:
Thanks!

PR #543, opened 2016-08-11
Extend Template constructor.

rhattersley:
Similar to the resource classes, this allows Description and Metadata to be set on construction.

Currently you have to do:
```python
template = Template()
template.add_description('Blah blah blah ...')
```

This PR makes it more consistent with other resources and allows:
```python
template = Template(Description='Blah blah blah ...')
```


markpeek:
Thanks!

PR #544, opened 2016-08-11
adding new application loadbalancer objects and properties, following AWS announcement https://a…

axelpavageau:
https://aws.amazon.com/blogs/aws/new-aws-application-load-balancer/

### DIFF
--- a/examples/ApplicationELB.py
+++ b/examples/ApplicationELB.py
@@ -1,0 +1,213 @@
+# Converted from ELBSample.template located at:
+# http://aws.amazon.com/cloudformation/aws-cloudformation-templates/
+
+from troposphere import Base64, FindInMap, GetAtt, Join, Output
+from troposphere import Parameter, Ref, Template
+import troposphere.ec2 as ec2
+import troposphere.elasticloadbalancingv2 as elb
+
+
+def AddAMI(template):
+    template.add_mapping("RegionMap", {
+        "us-east-1": {"AMI": "ami-6411e20d"},
+        "us-west-1": {"AMI": "ami-c9c7978c"},
+        "us-west-2": {"AMI": "ami-fcff72cc"},
+        "eu-west-1": {"AMI": "ami-37c2f643"},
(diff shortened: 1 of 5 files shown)


philtay:
Please also add the `TargetGroupARNs` property to `AWS::AutoScaling::AutoScalingGroup` in order to complete this PR as per `August 11, 2016` [CloudFormation release](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/ReleaseHistory.html). Thanks a lot!

philtay:
Also note that I have created a separated module for `AWS::ApplicationAutoScaling`. I mean, now in troposphere we have both `autoscaling` and `applicationautoscaling` modules. This PR, instead, includes both `AWS::ElasticLoadBalancingV2` and `AWS::ElasticLoadBalancing` in the same module. The natural name of the class for `AWS::ElasticLoadBalancingV2::LoadBalancer` would be `LoadBalancer` but you had to name it `ApplicationLoadBalancer` in order to avoid clashing with the already existing `AWS::ElasticLoadBalancing::LoadBalancer`. Basically I suggest to use separate modules for these resources too, but I think it's up to @markpeek to decide. I'm ready to merge `autoscaling` and `applicationautoscaling` if Mark wants me to.

axelpavageau:
You're right, it makes more sense to create a separate module. I hesitated between both options and went for a single module, but after a bit of sleep your way seems better. 
The object's names will be more consistent with AWS'.
I'll correct my PR

philtay:
LGTM

ntwaddell:
Awesome, thanks for getting this added so quickly.  :D

markpeek:
Thanks! Appreciate the rework and adding the example.

mitesh9875:
Hi, 
I have not get any example to setup application ELB, please share example for application loadbalancer.

phobologic:
@mitesh9875 The example troposphere code and the resulting cloudformation template are both included in this pull request.  You can find them in the `examples` directory as well.

https://github.com/cloudtools/troposphere/blob/master/examples/ApplicationELB.py

PR #548, opened 2016-08-15
Improve check_zip_file to calculate a minimum length

ikben:
This reworks the check_zip_file function (https://github.com/cloudtools/troposphere/pull/537 and https://github.com/cloudtools/troposphere/commit/1f69f01155105117eb2e543d2a96569e2bae3a3c) to always try to get a minimum length, even if not all parts are strings.

My original plan was to only replace `return` on line 56 with `continue`, but when I started writing it, I saw that I could also  use list comprehension instead of a loop (and make the variable name use snake_case).


markpeek:
Thanks for reworking this function and cleaning it up.

PR #549, opened 2016-08-17
Update ECS LoadBalancer property

philtay:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-service-loadbalancers.html


bpholt:
I was going to make a very similar PR. This change is necessary to use Application Load Balancers with ECS Services.

markpeek:
Thanks!

PR #555, opened 2016-08-22
Get support for new DomainVersion attribute

pfreixes:
As was announced right now AWS supports ElasticSearch 2.3, the default version if it is not specified is 1.5. This PR enables to the user choose a different version rather than the default one.


pfreixes:
There is an issue in AWS cloudformation and it is not still supporting `ElasticSearchVersion` variable, I will reopen it once it is fixed.

PR #556, opened 2016-08-24
Add support for IAM Roles in ECS Task Definitions

Canuteson:
Even though the Cloudformation Documentation has not been updated to mention the TaskRoleArn property for TaskDefinitions, it is supported and successfully provisions task definitions with IAM Roles, as described in the API documentation:
http://docs.aws.amazon.com/cli/latest/reference/ecs/register-task-definition.html

### DIFF
--- a/tests/test_dict.py
+++ b/tests/test_dict.py
@@ -1,6 +1,7 @@
 import unittest
-from troposphere import Template
+from troposphere import Template, Ref
 from troposphere import ecs
+from troposphere import iam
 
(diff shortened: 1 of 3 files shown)


ejholmes:
For what it's worth, I can also confirm that this is supported in CloudFormation, just not in the docs.

👍 this is great.

markpeek:
Thanks @Canuteson for the PR and @ejholmes for the review.

PR #557, opened 2016-08-29
Allow Tags on AWS::CloudFormation::Stack (http://docs.aws.amazon.com/…

magnusboman:
Allow Tags on the AWS::CloudFormation::Stack resource.
Reference: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-stack.html


markpeek:
Thanks!

PR #558, opened 2016-08-29
Added support for protocol in container definition's PortMapping property.

EricFortin:
This is now supported in CloudFormation. From http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdefinitions-portmappings.html


markpeek:
Thanks!

Issue #554, opened 2016-08-22
ECS example


mijndert:
I would love to see an example for ECS to get an insight into how a CloudFormation template should be formed.

markpeek:
Closing as @axelpavageau added an example in #559

PR #559, opened 2016-08-31
adding a sample ECS Cluster template

axelpavageau:
In response to #554 I'm adding an example of an ECS Cluster with a few attached policies


markpeek:
Thanks!

PR #560, opened 2016-09-02
Add support for ElasticsearchVersion in Elasticsearch Domain

acmcelwee:
Similar to #556, there's now cloudformation support for specifying the desired ES version, but it's not added to their official docs. I stumbled upon [this StackOverflow thread](http://stackoverflow.com/questions/38796017/aws-elasticsearch-domain-cloudformation-template) when I was trying to find out whether it's supported. I've verified that the param is indeed supported.


markpeek:
Thanks!

Issue #562, opened 2016-09-07
Retire old dynamodb module


phobologic:
dynamodb2 has been around a while and throwing warnings, time to ditch the old module.

markpeek:
@phobologic please "make it so"

phobologic:
#564 should get us started down the path.

PR #564, opened 2016-09-08
Replace dynamodb module w/ dynamodb2

phobologic:
This is the first step of #562 - we still keep dynamodb2 around, but it just imports everything from the dynamodb module and triggers a warning, letting people know.


troposphere/dynamodb2.py
```diff
@@ -1,99 +1,35 @@
...
+assert GlobalSecondaryIndex
+assert LocalSecondaryIndex
+assert StreamSpecification
+assert Table
+assert attribute_type_validator
+assert key_type_validator
+assert projection_type_validator
+
+
+warnings.warn("This module has replaced by troposphere.dynamodb. Please use "
```
markpeek: Perhaps:
```This module has been replaced by troposphere.dynamodb. Please import that module instead as troposphere.dynamodb2 will be removed soon.```

phobologic: Done.

PR #565, opened 2016-09-08
add Tags prop

jeffreyctang:
Kinesis Stream is missing the Tags property.

This seems a little too easy, let me know if there's something I'm missing.


markpeek:
Thanks!

PR #569, opened 2016-09-19
Added cross-stack references

msirull:
Tested from the code I've been using for the last couple of weeks


markpeek:
Thanks!

BTW, I was thinking about the implementation after I merged it. Do you think there should be a check to ensure the length of the dict is 1? Alternatively is the key always "Name"? If so, should we just be passing in the exported name as a string instead of a dict?

msirull:
Yeah, the string over dict definitely makes more sense, good call. I'll try updating tonight.

PR #570, opened 2016-09-19
New ALB-ECS templates

cb17:


examples/ALB-ECS-WithElasticSearchLogs-AppStack
```diff
@@ -0,0 +1,344 @@
+from troposphere import Join, GetAtt, ImportValue
```
phobologic: This should be a `.py` file, like the rest in the `examples` directory

examples/ALB-ECS-WithElasticSearchLogs-NetworkStack
```diff
@@ -0,0 +1,540 @@
+import troposphere.elasticloadbalancingv2 as elb
```
phobologic: This file name should end with `.py` (`ALB-ECS-WithElasticSearchLogs-NetworkStack` -> `ALB-ECS-WithElasticSearchLogs-NetworkStack.py`)

kienpham2000:
this is awesome guys, I just used this example to launch an ECS cluster with ALB!

phobologic:
Ping @cb17 - any chance you can update this?

patrickpierson:
I was able to get the APP stack to work. The network stack is having issues on the later versions of troposphere. Given that @cb17 hasnt responded. I think this is closable and we could use the APP stack example to close this in another PR. Same for https://github.com/cloudtools/troposphere/pull/640

cb17:
@patrickpierson @phobologic Hi guys, sorry for the radio silence. I had been meaning to circle back around to this, but got a little disconnected from these templates. FWIW, the code for the ElasticSearch JS Lambda function I just stole from what gets generated in the console (and minified it, and added a "var esdomain = Ref(ESDomain)" to the front of it). So if someone wants to add that back in, I'm sure it could be useful to someone.

patrickpierson:
I think that since this PR is over a year old it makes sense to close it and if you want to create a new one based on the newer version of troposphere it would be good.

PR #574, opened 2016-09-22
add support for Tags to Certificate Manager Certificates

bpholt:
Certificate Manager certificates support tagging: see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-certificatemanager-certificate.html

### DIFF
--- a/troposphere/certificatemanager.py
+++ b/troposphere/certificatemanager.py
@@ -15,4 +15,5 @@ class Certificate(AWSObject):
         'DomainName': (basestring, True),
         'DomainValidationOptions': ([DomainValidationOption], False),
         'SubjectAlternativeNames': ([basestring], False),
+        'Tags': (list, False)
     }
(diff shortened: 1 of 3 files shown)


troposphere/certificatemanager.py
```diff
@@ -1,6 +1,13 @@
 from . import AWSObject, AWSProperty
 
 
+class Tag(AWSProperty):
```
markpeek: A new Tag class should not be created for this purpose. There are actually 3 different Tag implementations already: 1) using a simple list, 2) Tags in \_\_init\_\_.py, and 3) a specialized one for autoscaling. For now you can just define it as a list which is what most are defined as in the code. Do a `grep -r Tags troposphere` to see the various implementations.

bpholt: Good call. Updated in 754c1d13.

markpeek:
Thanks!

PR #579, opened 2016-09-27
Update Export param

cb17:
Update Export param to use Export(string) instead of dict

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -502,8 +502,8 @@ class Export(AWSHelperFn):
 
 class Output(AWSDeclaration):
     props = {
-        'Description': (str, False),
-        'Value': (str, True),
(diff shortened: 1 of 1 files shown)


cb17:
BTW, sorry for the delay on this, it's been a busy week.

I used the standard pattern of creating a helper function, but is there a better way so you can just pass in a string instead of Export(str)?

markpeek:
Thanks!

cristian04:
According to Cloudformation documentation, the export parameter looks like this:

"Export" : {
      "Name" : {"Fn::Sub": "${AWS::StackName}-VPCID" }
    }

How can I do the same with troposphere now that the value is a String?

markpeek:
This was discussed in #610. Use:
```
Export=Export("string")
```

PR #583, opened 2016-10-06
Fixes #581

### LINKED ISSUE #581: RDS template - "Engine" required even when "DBSnapshotIdentifier" is set
When creating an RDS instance using a Cloudformation template, the "Engine" property should not be required when "DBSnapshotIdentifier" is set, according to the Cloudformation documentation here:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-engine

However even when i have set "DBSnapshotIdentifier" i am getting the following error:
`ValueError: Resource Engine required in type AWS::RDS::DBInstance`
Which goes away when i set the "Engine" property.

Here is a minimal template demonstrating the issue:

```
#!/usr/bin/env python

from troposphere.rds import DBInstance

template = Template()

rds_instance = template.add_resource(DBInstance(
    'MyDB',
    DBName='test',
    AllocatedStorage=25,
    DBInstanceClass='db.m4.large',
    DBSubnetGroupName='default',
    DBSnapshotIdentifier='id'
))

print template.to_json()
```

Thanks.

magnusboman:


troposphere/rds.py
```diff
@@ -137,7 +137,7 @@ class DBInstance(AWSObject):
         'DBSecurityGroups': (list, False),
         'DBSnapshotIdentifier': (basestring, False),
         'DBSubnetGroupName': (basestring, False),
-        'Engine': (validate_engine, True),
```
markpeek: The use of validate_engine can still be used here instead of adding additional code in validate but change required to False.

troposphere/rds.py
```diff
@@ -221,6 +228,10 @@ def validate(self):
                 raise ValueError("AllocatedStorage must be no less than "
                                  "1/10th the provisioned Iops")
 
+        engine = self.properties.get('Engine', None)
```
markpeek: With validate_engine used above this isn't needed.

troposphere/rds.py
```diff
@@ -187,6 +187,13 @@ def validate(self):
                 'AWS::RDS::DBInstance.'
             )
 
+        if 'Engine' not in self.properties and \
+           'DBSnapshotIdentifier' not in self.properties:
+            raise ValueError(
+                'Either Engine or DBSnapshotIdentifier is required '
+                'AWS::RDS::DBInstance.'
+            )
+
```
markpeek: Since Engine can be optional when specifying DBSnapshotIdentifier, perhaps this could be coded:
```
        if 'DBSnapshotIdentifier' not in self.properties:
            if 'Engine' not in self.properties:
                raise ValueError(
                    'Resource Engine required in type %s' % self.resource_type)```

troposphere/rds.py
magnusboman: But if Engine isn't specified, wouldn't validate_engine fail?

markpeek: Good question. The property checks (type or validate functions) are only called if the property is specified in the class instantiation. And then the class validate function is called later on to do validation across all of the properties.

markpeek:
Thank you for the PR and the additional rework.

PR #587, opened 2016-10-08
Provide better type checking for values in from_dict

phobologic:
This validates properties & sub-properties when using from_dict.

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -182,14 +182,13 @@ class BaseAWSObject(object):
                 except TypeError:
                     pass
                 if is_aws_object:
-                    v = kwargs[prop_name]
-                    if not isinstance(v, collections.Mapping):
+                    if not isinstance(value, collections.Mapping):
                         raise ValueError("Property definition for %s must be "
                                          "a Mapping type" % prop_name)
-                    value = prop_type._from_dict(**v)
+                    value = prop_type._from_dict(**value)
 
                 if isinstance(prop_type, list):
-                    if not isinstance(kwargs[prop_name], list):
+                    if not isinstance(value, list):
                         raise TypeError("Attribute %s must be a "
                                         "list." % prop_name)
                     new_value = []
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -162,10 +162,43 @@ class BaseAWSObject(object):
         pass
 
     @classmethod
-    def from_dict(cls, title, dict):
-        obj = cls(title)
-        obj.properties.update(dict)
-        return obj
+    def _from_dict(cls, title=None, **kwargs):
+        props = {}
+        for prop_name, value in kwargs.items():
+            try:
+                prop_attrs = cls.props[prop_name]
+            except KeyError:
+                raise AttributeError("Object type %s does not have a "
+                                     "%s property." % (cls.__name__,
+                                                       prop_name))
+            prop_type, required = prop_attrs
+            if prop_name in kwargs:
+                value = kwargs[prop_name]
+                is_aws_object = False
+                try:
+                    is_aws_object = issubclass(prop_type, BaseAWSObject)
+                # prop_type isn't a class
+                except TypeError:
+                    pass
+                if is_aws_object:
+                    value = prop_type._from_dict(**kwargs[prop_name])
+
+                if isinstance(prop_type, list):
+                    if not isinstance(kwargs[prop_name], list):
+                        raise TypeError("Attribute %s must be a "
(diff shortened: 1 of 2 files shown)


troposphere/__init__.py
```diff
@@ -162,10 +162,43 @@ def validate(self):
...
     @classmethod
-    def from_dict(cls, title, dict):
-        obj = cls(title)
-        obj.properties.update(dict)
-        return obj
+    def _from_dict(cls, title=None, **kwargs):
+        props = {}
+        for prop_name, value in kwargs.items():
+            try:
+                prop_attrs = cls.props[prop_name]
```
mhahn: `cls.props` could be missing too

troposphere/__init__.py
```diff
@@ -162,10 +162,43 @@ def validate(self):
...
+            except KeyError:
+                raise AttributeError("Object type %s does not have a "
+                                     "%s property." % (cls.__name__,
+                                                       prop_name))
+            prop_type, required = prop_attrs
```
mhahn: maybe just `prop_type = prop_attrs[0]` since you don't use `required`?

troposphere/__init__.py
```diff
@@ -162,10 +162,43 @@ def validate(self):
...
+                if is_aws_object:
+                    value = prop_type._from_dict(**kwargs[prop_name])
+
+                if isinstance(prop_type, list):
+                    if not isinstance(kwargs[prop_name], list):
+                        raise TypeError("Attribute %s must be a "
+                                        "list." % prop_name)
+                    new_value = []
+                    for v in value:
+                        new_value.append(prop_type[0]._from_dict(**v))
```
mhahn: what if `v` isn't a dict? it seems like this needs the same check for aws object

troposphere/__init__.py
```diff
@@ -162,10 +162,43 @@ def validate(self):
...
+            if prop_name in kwargs:
+                value = kwargs[prop_name]
+                is_aws_object = False
+                try:
+                    is_aws_object = issubclass(prop_type, BaseAWSObject)
+                # prop_type isn't a class
+                except TypeError:
+                    pass
+                if is_aws_object:
+                    value = prop_type._from_dict(**kwargs[prop_name])
```
mhahn: what if `kwargs[prop_name]` isn't a dict?

troposphere/__init__.py
phobologic: I don't think it can - this is on AWSBaseObject, and AWSBaseObject sets up `self.props` in its init.

troposphere/__init__.py
phobologic: Good call, updating.

phobologic:
Updated to address @mhahn 's review, with another test.

troposphere/__init__.py
mhahn: oh you're right, this is the `cls` object. i was thinking of the function you had before that took any class

troposphere/__init__.py
```diff
@@ -162,10 +163,51 @@ def validate(self):
...
+            if prop_name in kwargs:
+                value = kwargs[prop_name]
+                is_aws_object = False
+                try:
+                    is_aws_object = issubclass(prop_type, BaseAWSObject)
+                # prop_type isn't a class
+                except TypeError:
+                    pass
+                if is_aws_object:
+                    v = kwargs[prop_name]
```
mhahn: not a huge deal, but you already define `value = kwargs[prop_name]` above, i would just reuse that here

troposphere/__init__.py
```diff
@@ -162,10 +163,51 @@ def validate(self):
...
+                    if not isinstance(v, collections.Mapping):
+                        raise ValueError("Property definition for %s must be "
+                                         "a Mapping type" % prop_name)
+                    value = prop_type._from_dict(**v)
+
+                if isinstance(prop_type, list):
+                    if not isinstance(kwargs[prop_name], list):
```
mhahn: this could just be `value` that you pulled out aboe

troposphere/__init__.py
```diff
@@ -162,10 +163,50 @@ def validate(self):
...
+                    value = prop_type._from_dict(**value)
+
+                if isinstance(prop_type, list):
+                    if not isinstance(value, list):
+                        raise TypeError("Attribute %s must be a "
+                                        "list." % prop_name)
+                    new_value = []
+                    for v in value:
+                        if not isinstance(v, collections.Mapping):
+                            raise ValueError(
```
mhahn: how would this work if this was a list of role names? does this always have to be a list of mapping types?

PR #590, opened 2016-10-10
Rename LoadBalancerAttributes to be singular

matalo33:
The LoadBalancerAttributes object contains only one LoadBalancerAttribute and shouldn't end in an 's'. i.e. singular not plural.

This is already the case for TargetGroupAttribute which is named correctly.

### DIFF
--- a/troposphere/elasticloadbalancingv2.py
+++ b/troposphere/elasticloadbalancingv2.py
@@ -8,7 +8,7 @@ from .validators import (
     network_port, integer)
 
 
-class LoadBalancerAttributes(AWSProperty):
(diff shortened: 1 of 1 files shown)


markpeek:
The choice of plural is based on the AWS CloudFormation documentation for [LoadBalancerAttributes](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-loadbalancer-loadbalancerattributes.html). If anything TargetGroupAttribute should be pluralized for consistency (although would have backward compatibility concerns). I appreciate the PR but closing as "by design".

PR #596, opened 2016-10-20
Support for S3 Lifecycle Rule property NoncurrentVersionTransitions

SonOfBytes:
Add support for S3 Lifecycle Rule move from singular NoncurrentVersionTransition to a set of NoncurrentVersionTransitions.  Using same methodology as the deprecation of Transition.

Passes formatting, flake, and tests


troposphere/s3.py
```diff
@@ -126,6 +128,27 @@ def validate(self):
                     'properties on S3 Bucket Lifecycle Rule. Please use '
                     '"Transitions" since the former has been deprecated.')
 
+        if 'NoncurrentVersionTransition' in self.properties:
+            if 'NoncurrentVersionTransitions' not in self.properties:
+                # aws moved from a single transition to a list of them
+                # and deprecated 'NoncurrentVersionTransition', so let's
+                # just move it to the new property and not annoy the user.
+                self.properties['NoncurrentVersionTransitions'] = [
```
phobologic: Do you think it's worth adding a warning here to let the user know so they can transition their templates?  At some point I'm assuming that Cloudformation will stop accepting the old format, so it might be nice to guide people along.

SonOfBytes: This will always generate the new format even if you use the old standard in py.  So the output will always be compliant.  Given that I don't see are hard wall that someone will hit at some point as the output will always be compliant.

So a warning may generate more noise than positive behaviour.  If we feel that a warning is the right way to go we can look at that.

ikben: I think a warning may be a good idea, to encourage keeping troposphere as a thin layer above cloudformation.

Having a warning here also allows troposphere to stop supporting the old format and simplify the code base when increasing major version numbers.

SonOfBytes: Is there a preferred way to generate warnings in troposphere?

phobologic: @SonOfBytes See https://github.com/cloudtools/troposphere/blob/master/troposphere/dynamodb.py#L6 for an example of using warnings.  Thanks!

phobologic:
Thanks @SonOfBytes !

PR #597, opened 2016-10-20
Include resource title in required attr exception

phobologic:
Based on helping with this issue:
http://stackoverflow.com/questions/40118794/dynamically-create-n-number-of-ec2-instances-with-troposphere/40122167?noredirect=1#comment67537659_40122167

Now, if a resource is missing a required attribute, and that resource
has a title, we'll include it in the exception.

ValueError: Resource ImageId required in type AWS::EC2::Instance

vs

ValueError: Resource ImageId required in type AWS::EC2::Instance (title: MyInstance)


markpeek:
Looks good. Perhaps add a test like this?
```
diff --git a/tests/test_basic.py b/tests/test_basic.py
index c904d58..cd6c9d4 100644
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -26,6 +26,12 @@ class TestBasic(unittest.TestCase):
     def test_goodrequired(self):
         Instance('ec2instance', ImageId="ami-xxxx", InstanceType="m1.small")

+    def test_required_title_error(self):
+        with self.assertRaisesRegexp(ValueError, "title:"):
+            t = Template()
+            t.add_resource(Instance('ec2instance'))
+            t.to_json()
+
     def test_extraattribute(self):

         class ExtendedInstance(Instance):
```

phobologic:
Good call, done.

Issue #595, opened 2016-10-20
SpotFleetRequest and PlacementGroup


drandyziegler:
I'm not certain that the use of a PlacementGroup is correctly coded up in troposphere for SpotFleetRequest.

The "LaunchSpecifications" for a "SpotFleetRequestConfigData" requires a "Placement" object. The content of the Placement object is described here
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-spotfleet-spotfleetrequestconfigdata-launchspecifications-placement.html

and in JSON turns out to look like
{
  "AvailabilityZone" : String,
  "GroupName" : String
}

In troposphere I can only find a "PlacementGroup" object which carries different information than above. It would be great if the "GroupName" above could be populated from the "PlacementGroup" object though.
Thanks in advance for looking into this.

markpeek:
Should be fixed via #598

PR #598, opened 2016-10-20
Added Placement class for the Placement property in LaunchSpecifications.

pshucksmith:
This pull request resolves issue #595 by adding a Placement class and using this in the LaunchSpecifications property.


phobologic:
Looks good - thanks!

Issue #580, opened 2016-09-29
EFS Example


renlord:
It'd be nice if there was an EFS example.

markpeek:
Closed via #601

PR #601, opened 2016-10-25
Add EFS example

rtkmhart:
After deploying EFS myself and seeing https://github.com/cloudtools/troposphere/issues/580 decided to give back and add an EFS example.

I purposely made it a a reasonably interesting example with security groups, IAM artifacts etc since that's probably a more realistic real-world example. The examples have helped me tremendously in the past but you often have to pull info from several different ones to get a full picture.


markpeek:
Awesome...thanks!

PR #602, opened 2016-10-25
add support to old mysql db engine

ltagliamonte:
This is a Backward compatibility change proposal.
I got old templates generated with troposphere that needs as db engine parameter 'mysql' in order to avoid the replacement of the RDS during a stack update.


markpeek:
Thanks!

PR #603, opened 2016-10-26
Fix typo in Example Allowed Values

whithajess:
was mode not node


phobologic:
@whithajess be sure to update https://github.com/cloudtools/troposphere/blob/master/tests/examples_output/Redshift.template as well to make the tests pass.

whithajess:
@phobologic Cheers sorry missed this will do now

phobologic:
@whithajess looks great, thanks!

PR #606, opened 2016-11-01
Add support for conditions in cfn2py script

gscalise:
Hi! I was porting some JSON template files to Troposphere using cfn2py, and realised that the conditions (both their declaration and their use in resources) were being ignored by the script.

I've added support for conditions, and it has worked well so far.


markpeek:
Thanks!

PR #609, opened 2016-11-03
Add HttpVersion prop to DistributionConfig (CloudFront HTTP/2 Support)

imjacobclark:
CloudFront now supports HTTP/2, this allow that property to be set in CloudFormation via Troposphere.

Support landed in CF to support this property on the 3rd November 2016, see: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/ReleaseHistory.html

https://aws.amazon.com/about-aws/whats-new/2016/09/amazon-cloudfront-now-supports-http2/

http://docs.aws.amazon.com/AmazonCloudFront/latest/APIReference/CreateDistribution.html

### DIFF
--- a/examples/CloudFront_S3.py
+++ b/examples/CloudFront_S3.py
@@ -37,7 +37,8 @@ myDistribution = t.add_resource(Distribution(
                 QueryString=False
             ),
             ViewerProtocolPolicy="allow-all"),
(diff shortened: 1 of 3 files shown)


markpeek:
Thanks!

PR #614, opened 2016-11-10
Include the Rule in the template in CloudWatchEventsSample.py

sidick:


markpeek:
Thanks for the fix but didn't realize there were two of them. Fixed via #629.

PR #615, opened 2016-11-10
Add a validator for ELB names

chrisgilmerproj:
The LoadBalancerName has some constraints that we can validate for ahead of time:

> This name must be unique within your set of load balancers for the region, must have a maximum of 32 characters, must contain only alphanumeric characters or hyphens, and cannot begin or end with a hyphen.

From: http://docs.aws.amazon.com/elasticloadbalancing/2012-06-01/APIReference/API_CreateLoadBalancer.html

My current implementation of the regex doesn't quite work.  A one character name ought to work but I have created a regex that will fail on names of one character.  I could use some help on it.

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -76,7 +76,8 @@ class TestValidators(unittest.TestCase):
     def test_elb_name(self):
         for b in ['a', 'a'*32, 'wick3d-elb-name', 'Wick3d-ELB-Name']:
             elb_name(b)
-        for b in ['a'*63, 'invalid_elb', '-invalid-elb', 'invalid-elb-', '-elb-']:
+        for b in ['a'*63, 'invalid_elb', '-invalid-elb',
+                  'invalid-elb-', '-elb-']:
             with self.assertRaises(ValueError):
                 elb_name(b)
 
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -4,7 +4,7 @@ from troposphere.validators import boolean, integer, integer_range
 from troposphere.validators import positive_integer, network_port
 from troposphere.validators import s3_bucket_name, encoding, status
 from troposphere.validators import iam_path, iam_names, iam_role_name
-from troposphere.validators import iam_group_name, iam_user_name
+from troposphere.validators import iam_group_name, iam_user_name, elb_name
 
 
 class TestValidators(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


tests/test_validators.py
```diff
@@ -73,6 +73,14 @@ def test_s3_bucket_name(self):
             with self.assertRaises(ValueError):
                 s3_bucket_name(b)
 
+    def test_elb_name(self):
+        for b in ['a', 'a'*32, 'wick3d-elb-name', 'Wick3d-ELB-Name']:
```
markpeek: Given the regex you might want to add:
```
a-a
aa
aaa
```

tests/test_validators.py
```diff
@@ -73,6 +73,14 @@ def test_s3_bucket_name(self):
...
+            elb_name(b)
+        for b in ['a'*63, 'invalid_elb', '-invalid-elb',
```
markpeek: Probably want 'a'*33 instead of 63. Add 'a-' as well.

troposphere/validators.py
```diff
@@ -73,6 +73,14 @@ def s3_bucket_name(b):
         raise ValueError("%s is not a valid s3 bucket name" % b)
 
 
+def elb_name(b):
+    elb_name_re = compile(r'^[a-zA-Z0-9][a-zA-Z0-9\-]{0,30}[a-zA-Z0-9]{1}$')
```
markpeek: Try something like this:
```
^[a-zA-Z0-9](?:[a-zA-Z0-9\-]{0,30}[a-zA-Z0-9]{1})?$
```

tests/test_validators.py
chrisgilmerproj: Added in 9b55342afe7a2f14cb126f49441fb19a8d514470

tests/test_validators.py
chrisgilmerproj: Good catch.  Copy pasta.  Fixed and expanded in 9b55342afe7a2f14cb126f49441fb19a8d514470

troposphere/validators.py
chrisgilmerproj: This worked! Nice call.  I've added it here: 9b55342afe7a2f14cb126f49441fb19a8d514470

I tested manually a bit on http://pythex.org/ just for context.

chrisgilmerproj:
@markpeek - Thanks for the help on this.  It looks to have passed the tests.  Do you want me to squash the commits before you accept it?

markpeek:
@chrisgilmerproj thanks for making the changes. I'll have time to review further and commit later. I'll squash as part of the github commit flow.

chrisgilmerproj:
I appreciate it.  Thanks so much.

markpeek:
Thanks!

PR #622, opened 2016-12-08
Added SNS::Subscription object

lil-cain:
Added an SNS subscription object, to allow us to create SNS
subscriptions as seperate objects, rather than just as properties
of an SNS topic.


markpeek:
Thanks! I did change the name to reflect it is a resource.

PR #624, opened 2016-12-09
Added CodeBuild Project resource and a CodeBuild example

jarosser06:


troposphere/codebuild.py
```diff
@@ -0,0 +1,106 @@
...
+class EnvironmentVariable(AWSProperty):
+    props = {
+        'Name': (basestring, True),
+        'Value': (basestring, True),
+    }
+
+
+class Environment(AWSProperty):
+    props = {
+        'ComputerType': (basestring, True),
```
adamfokken: I think this should be 'ComputeType'

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codebuild-project-environment.html#cfn-codebuild-project-environment-computetype

jarosser06:
Nice catch, it has been corrected.

markpeek:
Awesome...thanks!

Issue #407, opened 2016-03-01
positive integer - fix the validator


mathewaugustine:
https://github.com/cloudtools/troposphere/blob/master/troposphere/validators.py
zero is neither positive nor negative and we need to correct the validator.

Initiated a discussion and added more details at the forum.
https://groups.google.com/forum/#!topic/cloudtools-dev/1-5u2ZaVpbo

PR #625, opened 2016-12-13
Update validators.py

placydo:
positive integer - fix the validator #407
https://github.com/cloudtools/troposphere/issues/407

### DIFF
--- a/troposphere/validators.py
+++ b/troposphere/validators.py
@@ -24,7 +24,7 @@ def integer(x):
 
 def positive_integer(x):
     p = integer(x)
-    if int(p) < 0:
+    if int(p) < 1:
         raise ValueError("%r is not a positive integer" % x)
     return x
 


phobologic:
Hmm, while technically this is the right thing to do, this breaks previous functionality.  For example, https://github.com/cloudtools/troposphere/blob/c5cf9b1fbed69e0eb6f28f01b79a80a5a8c2077c/troposphere/apigateway.py

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-authorizer.html#cfn-apigateway-authorizer-authorizerresultttlinseconds

So that technically can take a 0 value.  In order to accept this, you'll need to go through the existing uses of the validator and ensure that we don't break something else with this fix.

phobologic:
@placydo any interest in updating this per comments, or should I go ahead and close?

PR #626, opened 2016-12-15
Add back support for Python 2.6

BPScott:
The use of dictionary comprehension in #468 caused support for Python
2.6 to be dropped. This change converts these comprehensions to use
simple iteration.

See b8a6dae for a prior example of dictionary comprehensions breaking support. This fix was modeled after that commit. As per that commit the test suite does not run in Python 2.6, however I have tested that `pycodestyle .` and `pyflakes .` both pass when running using python2.6.6 and the unit tests pass when using 2.7.11.

Potentially Travis could be configured to run `pycodestyle` and `pyflakes` under python 2.6 to help with the "best effort" support so even if the unit tests can't run it'll help spot compilation errors but that's a discussion for another day.

(Oh I'm using python2.6 rather than upgrading because I'm still stuck on CentOS 6 for some things)

### DIFF
--- a/troposphere/template_generator.py
+++ b/troposphere/template_generator.py
@@ -75,21 +75,27 @@ class TemplateGenerator(Template):
     def inspect_resources(self):
         """ Returns a map of `ResourceType: ResourceClass` """
         if not self._inspect_resources:
-            TemplateGenerator._inspect_resources = {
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks!

PR #629, opened 2016-12-19
Fixed missing add_resource in example Cloudwatch rule

sambanks:
In the example it was missing t.add_resource and therefore wasn't outputting the rule in the json.


markpeek:
Thanks! I did fix it up to add the output which allows the tests to pass.

sambanks:
Great! Thanks Mark

PR #631, opened 2016-12-20
Create new property Environment for aws lambda Function

chrisgilmerproj:
@markpeek - The `Environment` variable for a lambda `Function` is not just a dict, it has a specific format as defined here:

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-environment.html

Discovered this after much head bashing.  I think this will improve usage of the code for others.

### DIFF
--- a/troposphere/awslambda.py
+++ b/troposphere/awslambda.py
@@ -110,13 +110,20 @@ class EventSourceMapping(AWSObject):
     }
 
 
+class Environment(AWSProperty):
+
+    props = {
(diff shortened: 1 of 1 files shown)


vsimakhin:
that would be nice to have this property in official tropo release :+1:

markpeek:
Thanks!

chrisgilmerproj:
Thanks for the quick turnaround on this.  I really appreciate it.

PR #635, opened 2017-01-05
Extra validation

### LINKED ISSUE #634: Template validation
Hi,

I'd like to add validaters on the template object for these limits:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html

Would this be better done by adding checks to the add_condition/add_resource,etc methods on the template objects? Or should I add a validate method to the template object, and call that as part of to_json?

lil-cain:
Added validation for parameters, and for reaching max resources on a template. Fixes #634

### DIFF
--- a/tests/test_parameters.py
+++ b/tests/test_parameters.py
@@ -1,0 +1,14 @@
+import unittest
+
+from troposphere import Parameter
+
+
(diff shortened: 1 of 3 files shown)


markpeek:
Nice! Thanks!

PR #640, opened 2017-01-12
Cb17 fixup

lil-cain:
Fixes for changes requested for #570 

I've had to remove the lambda used to input logs to elasticsearch, as it relied on a javascript file that's not in the PR. There's also some other changes required to make the troposphere templates actually compile - Exports were a dictionary, so I modified them to be a troposphere.export

I haven't reviewed the actually generated templates to check they make sense :-)


phobologic:
@lil-cain The tests for this are failing - any chance you can take a look at it?

lil-cain:
Sorry - been out on paternity leave. Will come back and sort out what's up
with the python 2 tests in the next fortnight or so.

phobologic:
Thanks @lil-cain !  Good luck with your paternity leave :)

lil-cain:
Hey - just been looking at this again - it looks like there's an issue in template_generator._generate_autoscaling_metadata, in that this assumes that AWS::CloudFormation::Init will always have a single key (config). From the cloud formation docs though, you can have multiple keys in this dict, with only configsets being required in this case.

Am I barking up entirely the wrong tree with this?

PR #646, opened 2017-01-18
Adding RDS engine support for oracle-se2

AlexLast:
Currently troposphere doesn't support oracle-se2 as an engine choice. Added support & tested, working as expected.


markpeek:
Thanks!

PR #647, opened 2017-01-18
Add Fn::Split function

RasmusWernerLarsen:
This implements the Fn::Split function announced here:

https://aws.amazon.com/about-aws/whats-new/2017/01/aws-cloudformation-introduces-support-for-listing-imports-split-function-and-resource-coverage-updates/

I guess there should be a test case, but there isn't one for its partner Fn::Join, so perhaps I should create one for both?

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -1,7 +1,7 @@
 import json
 import unittest
 from troposphere import awsencode, AWSObject, AWSProperty, Output, Parameter
-from troposphere import Ref, Sub, Template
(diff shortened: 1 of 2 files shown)


markpeek:
Thanks for implementing this new function. i believe the delimiter is required to be a string so adding validation in the init would be good. And adding tests for join and split would be great.

markpeek:
Thanks for adding this function and handling the tests as well. I did tweak the tests slightly with a subsequent commit.

PR #651, opened 2017-01-22
Added to_dict() method to troposphere.Template

jarosser06:
This removes the JSONrepr methods and replaces with to_dict() which
handle creating dicts for each of the individual objects. The to_json()
method then just calls to_dict() on self and marshals the resulting
dictionary. Adding to_dict() allows for output into other formats such
as YAML without needing to output JSON and then reload it back just to
output in another format.

This can be considered a breaking change if others are calling JSONrepr. If needed I can re-add the JSONrepr functionality back to make this backwards compatible.

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -45,6 +45,29 @@ def is_aws_object_subclass(cls):
     return is_aws_object
 
 
+def encode_to_dict(obj):
+    if hasattr(obj, 'to_dict'):
+        # Calling encode_to_dict to ensure object is
+        # nomalized to a base dictionary all the way down.
+        return encode_to_dict(obj.to_dict())
+    elif isinstance(obj, (list, tuple)):
+        new_lst = []
+        for o in list(obj):
+            new_lst.append(encode_to_dict(o))
+        return new_lst
+    elif isinstance(obj, dict):
+        props = {}
+        for name, prop in obj.items():
+            props[name] = encode_to_dict(prop)
+
+        return props
+    # This is useful when dealing with external libs using
+    # this format. Specifically awacs.
+    elif hasattr(obj, 'JSONrepr'):
+        return encode_to_dict(obj.JSONrepr())
+    return obj
+
+
 class BaseAWSObject(object):
     def __init__(self, title, template=None, **kwargs):
         self.title = title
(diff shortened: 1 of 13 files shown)


troposphere/__init__.py
```diff
@@ -318,153 +343,146 @@ class GenericHelperFn(AWSHelperFn):
     def __init__(self, data):
         self.data = self.getdata(data)
 
-    def JSONrepr(self):
-        return self.data
+    def to_dict(self):
```
markpeek: Given the code duplication I had been contemplating moving the generic JSONrepr() into the AWSHelperFn base class. I think the same could be done for to_dict().

jarosser06: I can make that change.

jarosser06:
I had not seen the YAML branch but I will take a look. We also had the desire to compare an existing JSON template with a potentially updated template and use a dict compare so adding a native to_dict() makes it a little easier as well.

phobologic:
Looks good to me!

jarosser06:
Updated with to_dict() in the AWSHelperFn class and overriding only when necessary.

markpeek:
This is a nice change. Thank you for submitting it.

bpholt:
> I don't believe we need to support the JSONrepr() interface as I can't imagine people are calling it outside of to_json().

FWIW, we were using `JSONrepr()` outside of `to_json()`. I'm currently in the process of rewriting a bunch of tests to use the new interface.

Maybe we shouldn't have been using `JSONrepr`, maybe we shouldn't have been pulling Troposphere from GitHub instead of PyPI… but since I don't think `JSONrepr` was marked as an internal function it might have been better to deprecate it more slowly instead of removing it suddenly.

markpeek:
@bpholt my apologies for causing you an issue with this change. I never considered JSONrepr() to be a troposphere interface and wasn't aware anyone was using it other than through to_json(). I appreciate you letting me know and glad you think it is better. Could you let us know your use case for using it so I make sure we don't make the same mistake again?

bpholt:
@markpeek Thanks. We have a helper function in our tests that used `JSONrepr` to convert Troposphere objects to a format that can be compared/expected with the robber library. It basically did exactly what `encode_to_dict` does (and has now been replaced entirely with a call to `encode_to_dict`).

markpeek:
@bpholt TIL 1) reminded again to send a message to the mailing list to solicit feedback on potential breaking changes and 2) there is a robber library. Sorry again but glad is was a straightforward change for you.

PR #655, opened 2017-01-27
Spot fleet's IamInstanceProfile has a name property

smmckay:
See http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_IamInstanceProfileSpecification.html. It would make life easier for me to have this property available in troposphere.

### DIFF
--- a/troposphere/ec2.py
+++ b/troposphere/ec2.py
@@ -594,6 +594,7 @@ class SecurityGroups(AWSProperty):
 class IamInstanceProfile(AWSProperty):
     props = {
         'Arn': (basestring, False),
+        'Name': (basestring, False),
     }
 
 


craigbruce:
While that is correct unfortunately the name property is not exposed via the CloudFormation API, see http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-spotfleet-spotfleetrequestconfigdata-launchspecifications-iaminstanceprofile.html.

I would recommend suggesting to AWS support that they add this property.

markpeek:
Closing until this is available in CloudFormation.

smmckay:
Ah, I didn't notice that. Thanks.

PR #656, opened 2017-01-31
alb example that showcases autoscaling groups and route53 a aliases

hmain:
The examples are extremely useful so I thought I'd give back and create one of my own that showcases:

- Auto scaling + Application load balancer target groups
- Application load balancer as an A Alias target for a Route53 Round Robin
- Metadata Parameter Grouping
- AWS Parameter Types which grabs info automatically from your account
- Cloudformation outputs that can be used within other templates with FN::ImportValue

If it's too much in one template, or if it's not up to standards, or something else is wrong;  please feel free to tell me so as this is my first pull-request in this project.

### DIFF
--- a/examples/ALB_Autoscaling_Route53_A_Alias_Sample.py
+++ b/examples/ALB_Autoscaling_Route53_A_Alias_Sample.py
@@ -1,21 +1,33 @@
-from troposphere import Base64, Join, Parameter, Ref, Template, Output, FindInMap, GetAtt, Select, Export, Sub
+from troposphere import Base64, \
+    Join, \
+    Parameter, \
+    Ref, \
+    Template, \
+    Output, \
+    FindInMap, \
+    GetAtt, \
+    Select, \
+    Export, \
+    Sub
 import troposphere.autoscaling as autoscaling
 import troposphere.elasticloadbalancingv2 as elb
 import troposphere.route53 as route53
 
 t = Template()
 
+
 # Windows server 2016 Base
 def AddAMI(template):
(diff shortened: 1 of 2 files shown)


markpeek:
@hmain there was an issue with a newer version of pycodestyle that I just fixed. Please re-push your changes to get another travis run with your changes. Note: you do need to add the json output into tests/examples_output in order for it to pass. Try running ```python setup.py test``` to have a clean test run.

hmain:
@markpeek Thank you for your guidance, one test is failing at the moment for python 2.7, which is much better than before.

hmain:
Ok, seems like the test is failing because of:
`AttributeError: AWS::AutoScaling::AutoScalingGroup object does not support attribute TargetGroupARNs`

Which shouldn't fail. I'm not sure how to add a new test which validates this correctly, so any help here would be awesome.

hmain:
I've had another look into troposphere/troposphere/autoscaling.py and it seems like it supports my attribute there so I am not sure why it complains. Feel free to point me where it could be and I'll have a look at changing the behavior.

markpeek:
@hmain thank you for bringing up this PR again. I looked at it a little today and need more time to look further. I think there are some issues with how the template generator is functioning (might not like certain types of imports and might not deal with nested functions) along with a fundamental issue with how lists of objects is handled (the Conditions should be a list but it isn't getting flagged as an error).

For example, the use of "import troposphere.autoscaling as autoscaling" seems to cause the above error but "from troposphere import autoscaling" looks to work.

bsushant-athena (commented):
please merge this

PR #659, opened 2017-02-09
Allow valid value for TargetGroup HealthCheckPort

chrisgilmerproj:
For the TargetGroup in ElasticLoadBalancerV2 you can set the HealthCheckPort to the string `traffic-port`, which is the default if not set.  This is documented here:

http://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_CreateTargetGroup.html

I've added a new validator and tests for this.


markpeek:
Excellent. Thanks for adding the tests.

chrisgilmerproj:
Thanks!

PR #665, opened 2017-02-16
A fully automated IPv6 example for VPC setup.

RasmusWernerLarsen:
Thought this little example might be useful for someone. I've had a bit of a fight with the IPv6 support in AWS and currently it doesn't seem Amazon provides any proper mechanism for chopping the provided /56 into /64 pieces for automatic provisioning, so did this little custom resource for it, until Amazon hopefully get around to creating an intrinsic function for it, or however they'll solve it.

Maybe someone else is interested. Also, it's a decent basic example of IPv6 provisioning.


RasmusWernerLarsen:
The failing test is because of the custom resource, it needs 2to3 to pass for the Python3 versions.

PR #666, opened 2017-02-17
Fix ApiGateway.py sample

ZipoKing:
I have corrected ApiGateway sample code so the stack is being created correctly now (original one raises errors during creation).


markpeek:
Thanks!

Issue #314, opened 2015-09-09
Missing Examples


Miserlou:
There are a few AWS services that don't have examples in the examples folder. Would be nice to be consistent and thoroughly documented!

The ones I've encountered so far:
 - CloudTrail / "Trail"
 - Lambdas
 - SES (Not implemented?)

vovanec:
+1 for AWS lambda example!

ikben:
I haven't tested it, but by using cfn2py on http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/quickref-lambda.html and minimal editing I was able to generate this. Is this the kind of example you're looking for?

```
from troposphere import FindInMap, GetAtt, Join, Output
from troposphere import Parameter, Ref, Template
from troposphere.awslambda import Function, Code
from troposphere.cloudformation import CustomResource
from troposphere.ec2 import Instance
from troposphere.ec2 import SecurityGroup
from troposphere.iam import Role, Policy

t = Template()

t.add_version("2010-09-09")

ExistingVPC = t.add_parameter(Parameter(
    "ExistingVPC",
    Type="AWS::EC2::VPC::Id",
    Description="The VPC ID that includes the security groups in the ExistingSecurityGroups parameter.",
))

InstanceType = t.add_parameter(Parameter(
    "InstanceType",
    Default="t2.micro",
    Type="String",
    AllowedValues=["t2.micro", "m1.small"],
))

ExistingSecurityGroups = t.add_parameter(Parameter(
    "ExistingSecurityGroups",
    Type="List<AWS::EC2::SecurityGroup::Id>",
))

t.add_mapping("AWSInstanceType2Arch",
              {u'm1.small': {u'Arch': u'PV64'},
               u't2.micro': {u'Arch': u'HVM64'}}
              )

t.add_mapping("AWSRegionArch2AMI",
              {u'ap-northeast-1': {u'HVM64': u'ami-cbf90ecb',
                                   u'PV64': u'ami-27f90e27'},
               u'ap-southeast-1': {u'HVM64': u'ami-68d8e93a',
                                   u'PV64': u'ami-acd9e8fe'},
               u'ap-southeast-2': {u'HVM64': u'ami-fd9cecc7',
                                   u'PV64': u'ami-ff9cecc5'},
               u'cn-north-1': {u'HVM64': u'ami-f239abcb',
                               u'PV64': u'ami-fa39abc3'},
[...]

PR #667, opened 2017-02-21
Adding example for CloudTrail

nfelsen:
Adding an example for CloudTrail.
I converted the official example from the [doc](http://docs.aws.amazon.com/AWSClo…udFormation/latest/UserGuide/aws-resource-cloudtrail-trail.html)
It's been requested in https://github.com/cloudtools/troposphere/issues/314

image


markpeek:
Thanks!

Issue #669, opened 2017-02-23
NetworkInterfaceProperty object does not support attribute Ipv6AddressCount


klingenm:
```python
troposphere.ec2.Instance(
            'SomeHost',
            ImageId='ami-0000',
            InstanceType='t2.micro',
            NetworkInterfaces=[NetworkInterfaceProperty(
                DeviceIndex=0,
                SubnetId='subnet-000000',
                AssociatePublicIpAddress=True,
                GroupSet=self.config.DefaultSecurityGroups,
                Ipv6AddressCount=1,
                Ipv6Addresses=[]
            )])
```

yields

`AttributeError: NetworkInterfaceProperty object does not support attribute Ipv6AddressCount`
and
`AttributeError: NetworkInterfaceProperty object does not support attribute Ipv6Addresses` (removing the Ipv6AddressCount argument naturally)

As a list item in troposphere.ec2.Instance argument NetworkInterfaces, NetworkInterfaceProperty should have member Ipv6AddressCount and Ipv6Addresses as documented on  
[http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-network-iface-embedded.html#aws-properties-ec2-network-iface-embedded-ipv6addresscount](url).

markpeek:
Closed via #670.

PR #670, opened 2017-02-23
Added missing props to ec2.NetworkInterfaceProperty (#669)

klingenm:


markpeek:
Nice! For some reason the CF release history didn't explicitly list the property so I missed it. Thanks.

PR #671, opened 2017-02-24
Initial take on SAM support in troposphere

dogonthehorizon:
Howdy,

We've taken a stab at implementing [SAM support](https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md) for Troposphere. Everything is wired up and works okay except for two related issues:

  - The events block in `Function` is a map of string key to object value. The current implementation technically generates this structure but the block titles are duplicated and it doesn't look very nice.
  - Because events are instantiated as full AWSObjects it's breaking tests in CodeBuild and CodePipeline

Any suggestions on how to move forward?

### DIFF
--- a/examples/Serverless_Api_Backend.py
+++ b/examples/Serverless_Api_Backend.py
@@ -1,0 +1,87 @@
+# Converted from api_backend located at:
+# https://github.com/awslabs/serverless-application-model/blob/dbc54b5d0cd31bf5cebd16d765b74aee9eb34641/examples/2016-10-31/api_backend/template.yaml
+
+from troposphere import Template, Ref
+from troposphere.awslambda import Environment
+from troposphere.serverless import Function, ApiEvent, SimpleTable
+
+t = Template()
+
+t.add_description(
+    "Simple CRUD webservice. State is stored in a SimpleTable (DynamoDB) "
+    "resource.")
+
+t.add_transform('AWS::Serverless-2016-10-31')
+
+simple_table = t.add_resource(
+    SimpleTable("Table")
+)
+
+t.add_resource(
+    Function(
+        "GetFunction",
+        Handler='index.get',
+        Runtime='nodejs4.3',
+        CodeUri='s3://<bucket>/api_backend.zip',
+        Policies='AmazonDynamoDBReadOnlyAccess',
+        Environment=Environment(
+            Variables={
+                'TABLE_NAME': Ref(simple_table)
+            }
+        ),
+        Events={
+            'GetResource': ApiEvent(
+                'GetResource',
+                Path='/resource/{resourceId}',
+                Method='get'
+            )
+        }
+    )
+)
+
+t.add_resource(
+    Function(
+        "PutFunction",
+        Handler='index.put',
+        Runtime='nodejs4.3',
+        CodeUri='s3://<bucket>/api_backend.zip',
+        Policies='AmazonDynamoDBReadOnlyAccess',
+        Environment=Environment(
+            Variables={
+                'TABLE_NAME': Ref(simple_table)
+            }
+        ),
+        Events={
+            'PutResource': ApiEvent(
+                'PutResource',
+                Path='/resource/{resourceId}',
+                Method='put'
+            )
+        }
+    )
+)
+
+t.add_resource(
+    Function(
+        "DeleteFunction",
+        Handler='index.delete',
+        Runtime='nodejs4.3',
+        CodeUri='s3://<bucket>/api_backend.zip',
+        Policies='AmazonDynamoDBReadOnlyAccess',
+        Environment=Environment(
+            Variables={
+                'TABLE_NAME': Ref(simple_table)
+            }
+        ),
+        Events={
+            'DeleteResource': ApiEvent(
+                'DeleteResource',
+                Path='/resource/{resourceId}',
+                Method='delete'
+            )
+        }
+    )
+)
+
+
+print(t.to_json())
--- a/examples/Serverless_S3_Processor.py
+++ b/examples/Serverless_S3_Processor.py
@@ -1,0 +1,38 @@
+# Converted from s3_processor located at:
+# https://github.com/awslabs/serverless-application-model/blob/dbc54b5d0cd31bf5cebd16d765b74aee9eb34641/examples/2016-10-31/s3_processor/template.yaml
+
+from troposphere import Template, Ref
+from troposphere.s3 import Bucket
+from troposphere.serverless import Function, EventSource
+
+t = Template()
+
+t.add_description(
+    "A function is triggered off an upload to a bucket. It logs the content "
+    "type of the uploaded object.")
+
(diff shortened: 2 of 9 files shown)


markpeek:
@dogonthehorizon Thank you for putting this together. There are a couple of easy pycodestyle changes needed. The easiest fix for the CodeBuild/CodePipeline issue is to add 'troposphere.serverless' into the DEPRECATED_MODULES in the template_generator.py. This prevents the duplication of those objects based on the type. I'll have to look further later into the Function issue you mentioned.

dogonthehorizon:
@markpeek thanks for the feedback! Sorry about the lint errors, I've gone ahead and fixed those up as well as adding `troposphere.serverless` to the deprecated modules for `TemplateGenerator`.

I'm seeing new test errors regarding a lack of title in the event source objects when they are instantiated by `TemplateGenerator`. I'll noodle on it a bit and see if there's a simple fix, but I suspect that this error will be resolved if/when we refactor the implementation for these event sources.

Cheers!

dyba:
@dogonthehorizon are you still working on this? I'm looking to pick up where you've left off.

dogonthehorizon:
Hey @dyba, I haven't been able to pick this work back up recently, feel free to carry it on!

Noah-Huppert:
@dogonthehorizon I have an interest in using the features in this PR. What needs to be done to get this in?

Noah-Huppert:
Forked @dogonthehorizon's fork [here](https://github.com/AdmitHub/troposphere). Looks like there are a couple pyflakes errors to fix and 2 failing test cases.

Noah-Huppert:
# The Problem
It took me a couple days to get familiar with the code base, but I have finally tracked down the issue.

It looks like the 2 test cases are failing due to an assertion in `TemplateGenerator._create_instance`. This assertion checks to make sure `args` is an instance of `Mapping`. 

I have determined that the root of this issue is due to some duplicate field names and values between some existing resources and the new resources being added by this PR. 

There is an AWS SAM Resource (new in PR) named `S3Event` with a `resource_type` of `S3`. There is also an AWS CodeBuild resource named `Source` (For `Project` source code in CodeBuild) with a field named `Type`, one of this `Type` field's valid values is `S3`. 

The issue occurs during the [CodeBuild example](https://github.com/AdmitHub/troposphere/blob/master/examples/CodeBuild.py) [test](https://github.com/AdmitHub/troposphere/blob/master/tests/test_examples_template_generator.py). The `TemplateGenerator` runs `_create_instance` on [this](https://github.com/AdmitHub/troposphere/blob/master/examples/CodeBuild.py#L17) CodeBuild `Source` resource in this form: 

```py
{
    'Location': "codebuild-demo-test/0123ab9a371ebf0187b0fe5614fbb72c",
    'Type': "S3"
}
```

When this resource has [`_convert_definition` called on it](https://github.com/AdmitHub/troposphere/blob/master/troposphere/template_generator.py#L234), `_convert_definition` sees that the Mapping has a `Type` field and attempts to set `expected_type` based on the value of the Mapping's `Type` field and the corresponding item in `inspect_resources` (`inspect_resources` is a mapping from AWS CloudFormation `resource_type`s to Troposhere resource class types):

```py
expected_type = self.inspect_resources[definition['Type']]
``` 

Previous to this PR when it attempted to do this `expected_type` would end up being `None` because there was not a module which declared a resource with type `S3`. Now however there is a `S3Event` object with `resource_type == S3`. 

As a result `_convert_definition` will find that `expected_type` is not `None` and therefore attempt to instantiate a `S3Event` where should actually be a regular mapping.

**tldr:** The `Type` field in the CodeBuild `Source` resource was never meant to cause the `TemplateGenerator` to instantiate a class. 

# The Solutions
I am looping in the original contributors to the `TemplateGenerator` as they might have better insight into the best way to solve this problem: @markpeek @BPScott @cwgreene @jonapich 

There are a couple ways I have thought of to solve this problem:

- Change the `resource_type` value of the `S3Event` from `S3` to something else so that it does not clash with CodeBuild `Source`.`Type` values
    - How would this affect CloudFormation template generation for the `S3Event` resource. I assume the CloudFormation API would still want the `Type` field of the `S3Event` to be `S3`
    - Perhaps something like `SAM::S3`
- Change the name of the `Type` field in the CodeBuild `Source` resource to be something other than `Type` b/c this sort of clashes with a lot

Which do you guys think would be the most practical / best?
Any other solutions are welcome.

markpeek:
The names used in resource_type and the property names in props are direct mappings to what CloudFormation requires for input. One thought, resource types are required at the top-level of the CF Resources. Perhaps only use the Type check at the top-level and then disallow resources (in other words, only allow properties) underneath them? Another thought would be to prefer matching to properties within a module based on the location of the initial resource?

Noah-Huppert:
I like the idea of making _convert_definition be a bit smarter about when
it creates a class.

We could also use the expected type from a resource's `props` to determine
which type of class we should be creating.

For example the CodeBuild `Project` class's `Source` prop says it
specifically takes a `Source` class. So we can use that information to
realize that we should not create an `S3Event`.

markpeek:
Yes, but just be aware that the expected type field can be complex with direct classes, functions (validators), list of classes (for properties that take lists), and tuple of classes (where multiple types could be used). In the case you're suggesting you could just verify the expected type is a class for a direct match and ignore otherwise.

Noah-Huppert:
Yea looking at all these functions there are a whole lot of cases to make
sure we cover.

markpeek:
And a comment I made on the mailing list, I'd be fine with adding code to exclude the serverless module and examples in the template generator for now (but create a new issue to track). That would allow the code itself to be merged and could still have the examples be used for tests against it.

Noah-Huppert:
*Note: When reading this is would probably be helpful to have `template_generator.py#_convert_defnition` open next to you, I refer to variable names and specific code branches*

I am going to try and add a `cls` argument to the `TemplateGenerator._convert_definition` method. 

This `cls` argument will represent the troposphere class that the `definition` argument represents. 

If it is `None` it means either:

1. `definition['Type']` does not exist
  1.1. Not an AWS resource (Probably a Parameter, Function, ect)
2. `definition['Type']` is not a type represented by any troposphere classes
  2.1 Processing a resource with a property named `Type` (For example a CodePipeline `Source` resource)
  2.2. Custom resource type in form `Custom::`

If `cls` is not `None` and the provided `definition` is an AWS resource (aka the `Type` field is present in the `definition` argument) then the `expected_type` variable will be set to `cls`. This value of `expected_type` will then be passed to `_create_instance` method. 

Then by passing a `cls` argument when calling `_convert_definition` in these places:

- `TemplateGenerator.__init__` when processing `cf_template` `Resources`
- Recursively in `TemplateGenerator._convert_definition`
- `TemplateGenerator._create_instance` & `_normalize_properties` `None` for all usages

This should prevent the Template Generator from seeing any field named `Type` and going off and trying to create an instance, as the decision to do so is now up to the caller of `_convert_definition`.

markpeek:
Closing per updated PR #754

PR #673, opened 2017-02-28
SSM Target fix Values instead of Value

Christoph-Schabert:
The SSM Target Property only support "Values" not "Value"

Source:
http://docs.aws.amazon.com/ssm/latest/APIReference/API_Target.html


radykal-com:
👍 I just faced this problem and was about to open a PR

markpeek:
Thanks!

PR #681, opened 2017-03-20
Do json validation on ApiGateway::Model Schema (Fix #679)

### LINKED ISSUE #679: ApiGateway Model Schema should be a Dict not a string
troposphere.apigateway.Model class contains a property called schema, it is defined as 'basestring', this property should be a dict.

Ideally, you should supply a class to represent 'Schema'

markpeek:
Allow both a string and dict for the ApiGateway::Model Schema.
The string form is checked for a valid json string and the dict
form is converted into a json string.

### DIFF
--- a/troposphere/apigateway.py
+++ b/troposphere/apigateway.py
@@ -1,5 +1,6 @@
 from . import AWSObject, AWSProperty
 from .validators import positive_integer
+import json
 
 
 def validate_authorizer_ttl(ttl_value):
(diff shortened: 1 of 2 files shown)


troposphere/apigateway.py
```diff
@@ -183,9 +184,21 @@ class Model(AWSObject):
         "Description": (basestring, False),
         "Name": (basestring, False),
         "RestApiId": (basestring, True),
-        "Schema": (basestring, False)
+        "Schema": ((basestring, dict), False)
     }
 
+    def validate(self):
+        if 'Schema' in self.properties:
```
phobologic: You need a way to skip this if `Schema` is an `AWSHelperFn`

PR #687, opened 2017-03-28
SQS: Add FifoQueue and ContentBasedDeduplication

horacio3:
Adding new SQS properties.

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sqs-queues.html


markpeek:
Thanks!

PR #691, opened 2017-03-29
Change EFS Tags to allow add tags as a python list

kaarolch:
Sometimes in the EFS Tags we use some advance tags like _"user:application"="XYZ"_. Unfortunately the FileSystemTags does not allow to use more advance tag' name when it uses the Tags type.


markpeek:
Thanks for the PR but I'm closing with the change in #692 which will allow a dict to be passed in and also doesn't break backward compatibility.

Issue #684, opened 2017-03-22
Issue with using `-` in Tags


brandonshough:
Having an issue with Tags when I try to use the `-` character for the key portion for a tag.

Example code that does work:
`Tags(testtag = '123456')`

Example code that doesn't work:
`Tags(test-tag = '123456')`

Is this because keywords is used for Tags?
https://docs.python.org/2.5/ref/identifiers.html

Can this be changed to allow more flexibility for Tag keys?

brandonshough:
Used this as a work around...
https://github.com/cloudtools/troposphere/issues/462

Issue #462, opened 2016-04-15
cfn2py and spaces in Tags


tomdeblende:
The output Python script fails to run when there are tags with spaces in the key name:

    Tags=Tags(
        Customer="XXX",
        environment="production",
        Patch Group="b",
        application="infra",
        Name="YYY",
    ),

Gives me:

 File "d:\Downloads\tropo.py", line 628
    Patch Group="b",
              ^
SyntaxError: invalid syntax

tamas7:
You can get around it by defining the key value pairs in a JSON style formatting and using `**kwargs` to unpack it.

Define the tags in a variable beforehand.

    tags = {
        'Customer': 'XXX',
        'environment': 'production',
        'Patch Group': 'b',
        'application': 'infra',
        'Name': 'YYY',
    }

    Tags=Tags(**tags)

Or do it inline.

    Tags=Tags(**{
        'Customer': 'XXX',
        'environment': 'production',
        'Patch Group': 'b',
        'application': 'infra',
        'Name': 'YYY',
    })

Or inline just for one tag (possible, but not that clean).

    Tags=Tags(
        Customer="XXX",
        environment="production",
        **{'Patch Group': 'b'},
        application="infra",
        Name="YYY",
    ),

PR #692, opened 2017-03-30
Allow a dict to be passed in to initalize Tags

markpeek:
The implementation of Tags has limitations due to the use of
kwargs for initialization. This change adds the ability to
initialize Tags using a dict.

@kaarolch @brandonshough @tomdeblende @tamas7 does this match the use cases for the issues filed on this behavior?

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -439,9 +439,18 @@ class ImportValue(AWSHelperFn):
 
 
 class Tags(AWSHelperFn):
-    def __init__(self, **kwargs):
+    def __init__(self, *args, **kwargs):
+        if not args:
+            # Assume kwargs variant
+            tag_dict = kwargs
+        else:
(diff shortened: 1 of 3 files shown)


troposphere/__init__.py
```diff
@@ -439,9 +439,21 @@ def __init__(self, data):
...
+    def __init__(self, *args, **kwargs):
+        if not args:
+            # Assume kwargs variant
+            tag_dict = kwargs
+        else:
+            if len(args) != 1:
+                raise(TypeError, "Multiple non-kwargs passed to Tags")
+
+            # Validate single argument passed in is a dict
+            if not isinstance(args[0], dict):
```
phobologic: Maybe use `collections.Mapping` here, in case someone wants to implement something other than `dict`.  Otherwise, this looks good.

markpeek:
This change should help with related issues #462, #684, and #691.

PR #694, opened 2017-04-12
feat(Cloudwatch): add support for latest Cloudwatch alarms properties

jleloup:
These new Cloudwatch alarms properties are documented here : http://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/AlarmThatSendsEmail.html#alarms-and-missing-data

CloudFormation API doc is not up to date right now though.

### DIFF
--- a/troposphere/cloudwatch.py
+++ b/troposphere/cloudwatch.py
@@ -26,7 +26,7 @@ class Alarm(AWSObject):
         'Dimensions': ([MetricDimension], False),
         'EvaluateLowSampleCountPercentile': (basestring, False),
         'EvaluationPeriods': (positive_integer, True),
-        'ExtendedStatistic': (basestring, False)
+        'ExtendedStatistic': (basestring, False),
         'InsufficientDataActions': ([basestring], False),
         'MetricName': (basestring, True),
         'Namespace': (basestring, True),
@@ -35,7 +35,7 @@ class Alarm(AWSObject):
         'Statistic': (basestring, False),
         'Threshold': (basestring, True),
         'TreatMissingData': (basestring, False),
-        'Unit': (basestring, False),
+        'Unit': (basestring, False)
     }
 
     def validate(self):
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks but holding off on this PR for now until the CloudFormation team implements and documents it being available.

Ashex:
AWS has added support for these properties:

https://aws.amazon.com/about-aws/whats-new/2017/06/aws-cloudformation-now-supports-amazon-emr-security-configurations-aws-lambda-and-aws-x-ray-integration-amazon-cloudwatch-percentile-redshift-resource-tagging-and-other-coverage-updates/

Can we also get `ExtendedStatistic` added to this PR?

jleloup:
Sure, let me have a look at it

jleloup:
Since `Statistic` is a conditional property related to `ExtendedStatistic` (optional too) : I have set it as non-mandatory.

Let me know if there is additionnal process to add for these dual-property conditions use case.

markpeek:
Thanks for adding the code. My reading is you must specify one of Statistic or ExtendedStatistic. If so, an example for adding validation can be found [here](https://github.com/cloudtools/troposphere/blob/master/troposphere/ec2.py#L306). Also, please alphabetize the properties.

Ashex:
Is there anything left to get this merged?

jleloup:
I don't think so, I'm leaving the last word to Mark :)

markpeek:
Thank you for the commit and adding the validation code. I only added one tweak just now which was to put the trailing comma back on the last props item. One of the things I like about Python allowing that and it reduces diffs later on when reordering/adding things.

jakul:
Hi,

Any idea when this will be released to PyPi? I would like to use `TreatMissingData`.

Thanks,
Craig

markpeek:
There is a regression due to including cfn_flip which I have hoped would be fixed upstream. I'll have to see whether to release with it or pull the yaml feature before I can release.

PR #695, opened 2017-04-13
Update autoscaling example to remove rebinding

levi-rs:
The example imports the `LaunchConfiguration` class object, but then rebinds the `LaunchConfiguration` name to an instance of the class, preventing any further use of the `LaunchConfiguration` class object. This patch binds the instance object to a new variable name, keeping `LaunchConfiguration` bound to the original class object.


markpeek:
Nice catch. Thanks!

PR #706, opened 2017-05-01
Added ECS PlacementConstraints, PlacementStrategy, and ServiceName

ktruckenmiller:
Added placement strategies and placement constraints for the ecs services. 

We could also add this to task definitions. 

Let us know if you'd like to do that and we can add it

### DIFF
--- a/troposphere/ecs.py
+++ b/troposphere/ecs.py
@@ -26,16 +26,32 @@ class DeploymentConfiguration(AWSProperty):
     }
 
 
+def placement_strategy_validator(x):
+    valid_values = ['random', 'spread', 'binpack']
+    if x not in valid_values:
+        raise ValueError("Placement Strategy type must be one of: %s" %
+                         ', '.join(valid_values))
+    return x
+
+
+def placement_constraint_validator(x):
+    valid_values = ['distinctInstance', 'memberOf']
+    if x not in valid_values:
+        raise ValueError("Placement Constraint type must be one of: %s" %
+                         ', '.join(valid_values))
+    return x
+
+
 class PlacementConstraint(AWSProperty):
     props = {
-        'Type': (basestring, True),
+        'Type': (placement_constraint_validator, True),
         'Expression': (basestring, False),
     }
 
 
 class PlacementStrategy(AWSProperty):
     props = {
-        'Type': (basestring, True),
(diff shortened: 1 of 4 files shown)


troposphere/ecs.py
```diff
@@ -26,6 +26,20 @@ class DeploymentConfiguration(AWSProperty):
...
+class PlacementConstraint(AWSProperty):
+    props = {
+        'Type': (basestring, True),
+        'Expression': (basestring, False),
+    }
+
+
+class PlacementStrategy(AWSProperty):
+    props = {
+        'Type': (basestring, True),
```
acmcelwee: Looks like we both added this, so I closed mine (#707). You could add validation on the `Type` field here that it's one of the allowed `random`, `spread`, `binpack`.

ktruckenmiller: I was adding that, but then I noticed that they had no other validators of this type. I could definitely do this, I wanted to see what cloudtools suggests

acmcelwee: Yeah, I think there's a mix of validation when it's simple, like in [emr](https://github.com/cloudtools/troposphere/blob/master/troposphere/emr.py#L87-L92), and just a `basestring` in others, so there's not really 100% consistency. IMHO, it's always nice to have early validation, but I can also see the argument that it's more future-proof to let cloudformation be the validator.

markpeek: Quite a bit of the validation was put in later in the project and there has never been an effort to go back to add it. If there is an easy way to constrain the value it is always good to add it now. This can be done like the emr examples of using a function instead of basestring, or easier by adding a validate() method to the object. The former will validate at object creation time and the latter at template creation time. I'm fine with either since that is better than having stacks rollback with illegal options.

troposphere/ecs.py
```diff
@@ -35,6 +65,9 @@ class Service(AWSObject):
         'DesiredCount': (positive_integer, False),
         'LoadBalancers': ([LoadBalancer], False),
         'Role': (basestring, False),
+        'PlacementConstraints': ([PlacementConstraint], False),
```
acmcelwee: I think the convention is for these props to be alphabetical. The `Placement...` ones should probably be moved above `Role`.

markpeek:
Thanks!

PR #708, opened 2017-05-01
Add ECS::Service ServiceName support

acmcelwee:
The [April 28, 2017 release](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/ReleaseHistory.html) included support for the `ServiceName` property on `AWS::ECS::Service`.


acmcelwee:
Looks like this is covered in #706

PR #710, opened 2017-05-02
Add missing CidrIpv6 property to ec2.SecurityGroupRule.

RasmusWernerLarsen:
Discovered a missing ipv6 parameter to the ec2.SecurityGroupRule object.

See: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group-rule.html


phobologic:
Thanks @RasmusWernerLarsen!

PR #712, opened 2017-05-03
logs.LogGroup#RetentionInDays is strictly defined list

ZipoKing:
When definining number of retention days for log group, it must be value from strictly defined list; if other value will be provided, CloudFormation will raise an error during stack creation


phobologic:
Thanks @ZipoKing!  Really like the new validator!

ZipoKing:
@phobologic Just trying to follow convention :) I am work a lot last days with troposphere with a bit more complex AWS stacks so finding such small issues

phobologic:
Awesome, well, PRs are definitely welcome :) Thanks again.

Issue #717, opened 2017-05-11
Cloudformation parameter of type "number" expects basestring for value attributes


wilvk:
With the following code:

```
ASGMinSize = t.add_parameter(Parameter(
    "ASGMinSize",
    Default=1,
    MinValue=1,
    Type="Number",
    Description="Minimum Instances for Autoscaling Group.",
    MaxValue=2,
))
```

I am getting the error:

> TypeError: <class 'troposphere.Parameter'>: ASGMinSize.Default is <type 'int'>, expected <type 'basestring'>

This is easily fixed by quoting the fields such as:

```
ASGMinSize = t.add_parameter(Parameter(
    "ASGMinSize",
    Default="1",
    MinValue="1",
    Type="Number",
    Description="Minimum Instances for Autoscaling Group.",
    MaxValue="2",
))
```

As the former is how it was originally converted from cloudformation templates using /scripts/cfn2py and given a number field should be a numeric, it would make sense to allow a special type for the field `Default` and for `MinValue` and `MaxValue` to be of type integer.

markpeek:
This looks like just a bug in Default that should be fixed. The validators for Min/Max Value/Length should allow for integer types to be specified. For reference the code is [here](https://github.com/cloudtools/troposphere/blob/master/troposphere/__init__.py#L578).

wilvk:
Have created PR #718 for this.

PR #718, opened 2017-05-12
Validation for parameters that can be numeric or string

wilvk:
This fixes issue #717 

as /script/cfn2py converts some values to integers that are considered strings in AWS, this validation allows numerics or strings to be defined for a parameter, thus preventing a validation exception being thrown when either would be a valid form.

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -580,7 +580,7 @@ class Parameter(AWSDeclaration):
     NUMBER_PROPERTIES = ['MaxValue', 'MinValue']
     props = {
         'Type': (basestring, True),
-        'Default': (basestring, False),
+        'Default': (validators.string_or_integer, False),
         'NoEcho': (bool, False),
         'AllowedValues': (list, False),
         'AllowedPattern': (basestring, False),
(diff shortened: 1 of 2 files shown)


troposphere/__init__.py
```diff
@@ -580,7 +580,7 @@ class Parameter(AWSDeclaration):
     NUMBER_PROPERTIES = ['MaxValue', 'MinValue']
     props = {
         'Type': (basestring, True),
-        'Default': (basestring, False),
+        'Default': (validators.string_or_integer, False),
```
markpeek: Thank you for the PR. Actually a new validator isn't needed. Using validators.integer will allow for both a string and integer value to be validated.

wilvk: Thanks for that - have amended accordingly

wilvk:
@markpeek looks like this change has broken the CI tests..

markpeek:
My apologies, I wasn't thinking about the other default types. This can still be done but will require adding the logic into the validate method as the type needs to be available to correctly validate and coerce the value. It might be a little tricky with allowing string, int, and float values. Plus needing to add tests. Feel free to take a stab or I'm happy to implement it over the weekend.

markpeek:
Closing as I have a different implementation to push which addresses this issue.

PR #722, opened 2017-05-18
Add required attribute, JobFlowId, to EMR::InstanceGroupConfig

alexwen:
The EMR InstanceGroupConfig is missing JobFlowId which is a required parameter and specifies the EMR cluster in which the instance group should be created. This PR adds that attribute and marks it as required.


markpeek:
Very odd that was missing. Thanks for the PR!

PR #724, opened 2017-05-22
Feature/support passing ref to user data

hybmg57:
Ref is now supported in the method from_file of helper file.
Also added unit test

### DIFF
--- a/tests/test_userdata.py
+++ b/tests/test_userdata.py
@@ -56,6 +56,10 @@ class TestUserdata(unittest.TestCase):
     def test_nonexistant_file(self):
         self.assertRaises(IOError, self.create_result, 'nonexistant.sh')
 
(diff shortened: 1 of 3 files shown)


markpeek:
Nice change. I have two questions:
1. It seems this will always look for Ref. Perhaps it should be an option or another function name search for Ref.
2. Right now it only looks for 1 Ref. Is there a use case to look for multiple Ref's?

hybmg57:
@markpeek thanks for the feedback. I've updated userdata script with multiple refs and asserting these results in the test. It passes with mutliple Refs being replaced.
I've also added an option to include Ref.

Thanks again!

hybmg57:
@markpeek do you have any feedback on the changes?

ollikaven:
Plans to merge this?

PR #725, opened 2017-05-23
fix for ElastiCacheRedis.py example to use awacs

wilvk:
Thought this would be useful as the comments suggest that it should be done with awacs.


markpeek:
Thanks!

Issue #732, opened 2017-06-06
Raise ValueError when template limits are hit


apmclean:
Currently support exists for maximum resources and maximum parameters to raise a ValueError.
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html

However there are limits for Outputs and Mappings that can be tracked the same way.  This is much cleaner than having AWS reject the template.

Incoming pull request with test cases.

apmclean:
Fixed.

PR #733, opened 2017-06-06
Raise ValueError for Outputs and Mappings - Fix Issue #732

apmclean:
Raises ValueError when CloudFormation template limits are reached for Outputs or Mappings.

Test cases included and passing.

Thanks!

### DIFF
--- a/tests/test_template.py
+++ b/tests/test_template.py
@@ -1,6 +1,6 @@
 import unittest
 
-from troposphere import Template, Parameter
+from troposphere import Template, Parameter, Output
 from troposphere.s3 import Bucket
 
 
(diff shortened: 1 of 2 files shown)


apmclean:
Ok let me tune the tests up.  I followed the existing test case paradigm for parameters and resources which have the maximums hard-coded for their ValueError.  Do you want me to include a tune-up for those in the pull request too?

markpeek:
Ah, got it. Guess we didn't catch it before. Yes, if you're so inclined please tune-up the other ones as well. Thanks!

apmclean:
Would it make sense to expose the template limit maximums as attributes of the template class?  Then the test cases could determine them programmatically.

Within the Template class init something like:

```python
self.limits={
    'outputs': 60,
    'resources': 200,
    'parameters': 100,
    'mappings': 100
}
```

Then any future changes will just require changing the class versus the class + test cases.

markpeek:
Interesting idea. But I'm not sure if those would be used except for the test cases. Probably best not to clutter the template class with it at this time.

apmclean:
Understood.  Cleaned up the test case and passed flakes and codestyle.

Thanks again for a great tool!  I'll try to contribute in more meaningful ways in the future.

markpeek:
Excellent. Thank you for the contribution and fixing those additional tests.

PR #738, opened 2017-06-08
Introduces AWS::EMR::SecurityConfiguration Support for Troposphere

### LINKED ISSUE #737: Add support for AWS::EMR::SecurityConfiguration Resource
AWS recently introduced the CloudFormation resource AWS::EMR::SecurityConfiguration. 

I'm working on adding that functionality to Troposphere

bmurtagh:
Will close #737 once merged.

### DIFF
--- a/examples/EMR_Cluster.py
+++ b/examples/EMR_Cluster.py
@@ -153,15 +153,17 @@ emr_instance_profile = template.add_resource(iam.InstanceProfile(
 
 # EMR Cluster Resource
 
+security_config = template.add_resource(emr.SecurityConfiguration(
+    'EMRSecurityConfiguration',
+    Name="EMRSampleClusterSecurityConfiguration",
+    SecurityConfiguration=security_configuration,
+))
+
 cluster = template.add_resource(emr.Cluster(
     "EMRSampleCluster",
     Name="EMR Sample Cluster",
     ReleaseLabel='emr-4.4.0',
-    SecurityConfiguration=emr.SecurityConfiguration(
-        'EMRSecurityConfiguration',
-        Name="EMRSampleClusterSecurityConfiguration",
-        SecurityConfiguration=security_configuration
-    ),
+    SecurityConfiguration=Ref(security_config),
     BootstrapActions=[emr.BootstrapActionConfig(
         Name='Dummy bootstrap action',
         ScriptBootstrapAction=emr.ScriptBootstrapActionConfig(
--- a/examples/EMR_Cluster.py
+++ b/examples/EMR_Cluster.py
@@ -11,6 +11,32 @@ scaling_policy = emr.SimpleScalingPolicyConfiguration(
                 )
 
 
+kms_key = 'arn:aws:kms:us-east-1:123456789012:key/1234-1234-1234-1234-1234'
+
+security_configuration = {
+    'EncryptionConfiguration': {
+        'EnableInTransitEncryption': 'true',
+        'InTransitEncryptionConfiguration': {
+            'TLSCertificateConfiguration': {
+                'CertificateProviderType': 'PEM',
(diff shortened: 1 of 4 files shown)


bmurtagh:
@phobologic @markpeek do you guys have any insight on https://travis-ci.org/cloudtools/troposphere/jobs/240855609 ?

This time around I've apparently made Python 3, but angered Python 2.7 :P

markpeek:
Sorry for the delay. I finally got a chance to look at this change. The SecurityConfiguration object looks good but the SecurityConfiguration property of Cluster should be a basestring (the name of the security configuration applied to the cluster), not a SecurityConfiguration. It would be nice if AWS had an example for using this property as I wonder whether this will require a DependOn attribute to ensure the SecurityConfiguration exists before the Cluster uses it.

bmurtagh:
For my understand, I need to change the `dict` type to `basestring` for `SecurityConfiguration` due to:

```SecurityConfiguration
The security configuration details in JSON format.

Required: Yes

Type: String

Update requires: Replacement
```

CloudFormation expecting the type to be a string even though it's formatted in JSON. I think the " in JSON format" confused me to use a `dict` vs a string. Making the change now.

bmurtagh:
I reverted back to the broken Python 2.7 code. @markpeek I thought I understood your request, but once I changed SecurityConfiguration from dict to basestring in troposphere/emr.py all build jobs failed: https://travis-ci.org/cloudtools/troposphere/builds/243246661

bmurtagh:
Regarding the dependency aspect of EMR Security Configurations. They are not required to deploy a cluster unless the cluster requires encryption (at rest and/or in transit).

markpeek:
First off, it still should be a basestring which should be the Name parameter from the SecurityConfiguration resource. Note: this requires rewriting the example and the template. However, when I tried to apply this simple fix the template generator is throwing some odd errors. Best thing to do is leave it as-is for now and let me find some time to look into the underlying issues. Thanks.

markpeek:
ok, the tests are passing now. Please review my change. The Cluster property takes either a string (the name of the SecurityConfiguration) or a Ref() of the SecurityConfiguration which returns the name as well. At least that's what I believe is the intended behavior.

bmurtagh:
Looks good Mark. Ty

One question I have to help me learn is what/why/how did you determine that the `SecurityConfiguration` property of the `AWS::EMR::Cluster` object needed to be a `basestring` vs referencing the `emr.SecurityConfiguration` class?

https://github.com/cloudtools/troposphere/pull/738/commits/ddaa43ae9befea8807ede78eb7dc08c8c41a4a6a#diff-f3aebd6ddbd9724805bf7c11394abafeL246

markpeek:
Sure, let me try to answer the questions.

1. The Cluster SecurityConfiguration [docs](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-emr-cluster.html#cfn-emr-cluster-securityconfiguration) listed the type as "String". Also, AWS::EMR::SecurityConfiguration is a resource so it needs to be instantiated separately (as opposed to a property).
2. Looking at the [docs](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-emr-securityconfiguration.html#w2ab2c19c12d512c13) for AWS::EMR::SecurityConfiguration we see there is a return value for Ref(). I ran a quick test and saw the Ref() on a SecurityConfiguration returns its name (either a constructed name or your own name through the Name property). I didn't need to do the Ref() as I could have hardcoded a name (String) but I thought it better to use Ref() in the example. (Note: the docs are wrong saying Ref() returns the cluster ID when it really returns the SecurityConfiguration name and I followed up with AWS about it.)

bmurtagh:
Awesome thank you very much

markpeek:
Thanks!

PR #736, opened 2017-06-08
Add autoscaling for Task Instance Groups

alexwen:
Also, allow for negative adjustments (downscaling) in ScalingAdjustments

PR #743, opened 2017-06-16
EMR AutoScaling Complex Validation and Introduction of an ignore validator type

bmurtagh:
This is my attempt to help button up #736. This should incorporate @alexwen's code with the addition of a `validate` method to perform the more complex `AdjustmentType` and `ScalingAdjustment` validations that @markpeek requested in the original PR. 

This PR also introduces an `ignore` method in validators.py that is implemented to "bypass" (for lack of a better term) so the simple property validation is "skipped' and the new `validate` method in `emr.py` is used.

The reasoning for the `ignore` method is that the `ScalingAdjustment` property in `SimpleScalingPolicyConfiguration` will actually take different types depending on the value of  `AdjustmentType`. 

As a result, we need to skip the initial checking in favor of the more complex validation.

**Old**:
```class SimpleScalingPolicyConfiguration(AWSProperty):
    props = {
        'AdjustmentType': (basestring, False),
        'CoolDown': (positive_integer, False),
        'ScalingAdjustment': (integer, True),
    }
```

**New**:
```
class SimpleScalingPolicyConfiguration(AWSProperty):
    props = {
        'AdjustmentType': (basestring, False),
        'CoolDown': (positive_integer, False),
        'ScalingAdjustment': (ignore, True),
    }
```

### DIFF
--- a/tests/test_emr.py
+++ b/tests/test_emr.py
@@ -6,7 +6,7 @@ import troposphere.emr as emr
 
 
 scaling_policy = emr.SimpleScalingPolicyConfiguration(
-                    AdjustmentType="EXACT_CAPACITY",
+                    AdjustmentType=emr.EXACT_CAPACITY,
                     ScalingAdjustment="1",
                     CoolDown="300"
                 )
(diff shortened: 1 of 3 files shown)


markpeek:
@bmurtagh I made a few tweaks if you want to verify. Thanks.

bmurtagh:
👀... if that's what you refer to as a few, I'd hate to see a PR where you claim a lot of tweaks 😎

+1 from me

markpeek:
Thanks @bmurtagh and @alexwen.

PR #745, opened 2017-06-20
removed forgotten & replaced ignore validator method

bmurtagh:
Forgot to remove the ignore method in `validators.py`when `defer` was introduced.

### DIFF
--- a/troposphere/validators.py
+++ b/troposphere/validators.py
@@ -60,11 +60,6 @@ def floatingpoint(x):
         return x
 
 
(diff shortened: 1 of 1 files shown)


markpeek:
@bmurtagh I'm fine with leaving ignore() in the validators as it might make sense at some point and was a good idea. For the case implemented I thought the name defer was more appropriate and reused the idea.

bmurtagh:
no worries, i thought we missed it during clean up

Issue #709, opened 2017-05-01
Depends on must be a string, not a Ref


jdc0589:
It looks like you can't actually use a `Ref` for `DependsOn` properties, despite it being in some of the examples. Right now, I'm having to pass in the literal string Name of the resource.

example:
```python
serices_public_subnet_1_nat_gateway = t.add_resource(NatGateway(
	"ServicesPublicSubnet1NatGateway",
	DependsOn = Ref(serices_public_subnet_1_nat_gateway_eip), # this won't work
	SubnetId = Ref(serices_public_subnet_1),
	AllocationId = GetAtt(serices_public_subnet_1_nat_gateway_eip, "AllocationId")
))
```

Error: `An error occurred (ValidationError) when calling the ValidateTemplate operation: Template  Format error: DependsOn must be a string or list of strings.`

jdc0589:
I just had a mental lapse and realized this might be an actual CloudFormation quirk... Either way, [some of the examples](https://github.com/cloudtools/troposphere/blob/6101e637ef80d83e23dfed0219f06637561e78f2/examples/EFS.py#L121) look like they won't actually work.

phobologic:
@jdc0589 you're right, DependsOn is supposed to point at the resource name itself, not a Ref.  We should fix the examples.  I'll try to take a look soon-ish, or if you want to submit a PR that'd be great.

bmurtagh:
Only looks like `examples/EFS.py` is using the incorrect method of `Ref()` with a `DependsOn`:

```
examples/EFS.py:    DependsOn=Ref(efs_mount_target)
```

```
~\troposphere$ grep -r "DependsOn" examples/*
examples/ApiGateway.py:    DependsOn='FoobarFunction',
examples/ApiGateway.py:    DependsOn="LambdaMethod",
examples/CloudTrail.py:    DependsOn=["BucketPolicy", "TopicPolicy"],
examples/EFS.py:    DependsOn=Ref(efs_mount_target)
examples/VPC_single_instance_in_subnet.py:        DependsOn='AttachGateway',
examples/VPC_single_instance_in_subnet.py:        DependsOn='AttachGateway',
```

phobologic:
This is fixed in #746, thanks @bmurtagh

PR #746, opened 2017-06-21
EFS DependsOn Ref to String fix

bmurtagh:
Closes low hanging fruit #709

As part of this change, I removed the `title` usage in favor of the positional argument for naming a resource. Using the positional seems to be the defacto standard looking at other examples and code within Troposphere.


phobologic:
Hey @bmurtagh - I updated your example to use a new method of referring to DependsOn that is more pythonic (and was just accepted into troposphere).  Let me know if you have any issue with it - @markpeek letting you know too.

markpeek:
Thanks!

Issue #550, opened 2016-08-18
ApiGateway::Method validation doesn't match documentation


ikben:
I don't have time right now to fix it myself, so I'm logging it here.

This is the current code, with in comment what the docs say about required (http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-method.html)

```
# Left out all methods that are defined in the right way

class Method(AWSObject):
    props = {
        "AuthorizationType": (str, False),  # Required
        "HttpMethod": (str, False),  # Required
        "ResourceId": (str, False),  # Required
        "RestApiId": (str, False)  # Required
    }
```

JaniszM:
+1, I need this property so much right now ;)
Link to doc: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apitgateway-method-integration.html#cfn-apigateway-method-integration-passthroughbehavior

markpeek:
@ikben @JaniszM I just did some quick edits for the missing prop and adjusting some of the required fields. Might be good for additional scrutiny on the names, types, and required fields.

ikben:
Looks great!

I quickly went through the docs, and this is what caught my eye, I haven't tested anything:

````
class StageDescription(AWSProperty):
    props = {
        # ThrottlingRateLimit is defined in the docs as a Number, while others are defined as Integer, so this may need to be a check on float.
        "ThrottlingRateLimit": (positive_integer, False),
     }

class MethodSetting(AWSProperty):
    props = {
         # Same remark as above
        "ThrottlingRateLimit": (positive_integer, False)
    }

class Method(AWSObject):
    props = {
        # We could add extra validation to authorizationType
        # (f you specify the AuthorizerId property, specify CUSTOM for this property.)
        "AuthorizationType": (basestring, True),
    }

class Integration(AWSProperty):
    props = {
        "IntegrationHttpMethod": (basestring, False),  # Conditional
    }

class Model(AWSObject):
    props = {
        # Schema is a JSON object, so I think this should be (dict, False)
        "Schema": (basestring, False)
    }

class RestApi(AWSObject):
    props = {
        # Body is a JSON object, so I think this should be (dict, False)
        "Body": (basestring, False),
        "FailOnWarnings": (basestring, False),  # Should be (bool, False)
        "Name": (basestring, False), # Conditional
    }
```

aarcro:
[...]

PR #749, opened 2017-06-22
API Gateway Property Updates and Improved Validation Coverage

bmurtagh:


bmurtagh:
@markpeek @phobologic i do have a question regarding the `validate` for the `RestApi(AWSObject)` in regards to the `Name` property. See http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-restapi.html#cfn-apigateway-restapi-name

The documentation for Name states, "Required: Conditional. Required if you don't specify a OpenAPI definition."

I assume this OpenAPI definition would be stored in the `Body` parameter, but the `Body` property's documentation states, "An OpenAPI specification that defines a set of RESTful APIs in the JSON format. For YAML templates, you can also specify the specification in the YAML format.

Required: No"

The terminology differences between "OpenAPI specification" and "OpenAPI definition" are confusing me. I assume they're interchangeable, but I do not have any experience with API Gateway so I'm guessing. Hopefully one of you guys have come across this before & can clarify.

bmurtagh:
This should close #550

markpeek:
The way I read it is, you can specify either Body or BodyS3Location to provide an OpenAPI spec which will then auto-name the API based on the title contents in the spec. (Using the API Gateway UI and playing with the Example API helps illustrate this). I'd say Name is required if neither of those are specified. It may also be possible (I haven't tested it) to use one of the Body/BodyS3Location *and* specify a Name. I would expect the Name property to take precedent but I don't know the answer to this question.

churnikorn:
Maybe this helps: if you create a rest api resource via the template and provide a name, that name will be overwritten once the AWS importer is done if the `title` in the swagger file is not the same as the name. I'm not entirely sure where in the flow the check happens for AWS and if specifying a title in the swagger doc would be picked up during the resource creation to the apply to its name.

PR #752, opened 2017-06-26
Accept Join type as parameter default value as it returns a string

stan3:
I'm getting a validation error with valid parameter:
Parameter default type mismatch: expecting type String got <class 'troposphere.Join'> with value <troposphere.Join object at 0x7f786b807d10>
due to using a Join as the default.

Changed validation check to allow Join type.

### DIFF
--- a/tests/test_parameters.py
+++ b/tests/test_parameters.py
@@ -1,6 +1,6 @@
 import unittest
 
-from troposphere import Parameter
+from troposphere import Join, Parameter
 
 
 class TestInitArguments(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


markpeek:
Instead of adding Join should this add AWSHelperFn? In other words, could other functions be used there as well (Select, Split, etc.)?

stan3:
I would guess functions that return a string would be ok (select, sub, ...) but not those that a return a list like split.

markpeek:
Thanks!

stan3:
Apologies but turns out this is actually a bogus change as AWS also only accepts the Default value as a plain string.  Please revert these changes.

PR #755, opened 2017-06-28
Fixed UsagePlan example to proper Ref

ntwaddell:
Tried the example, it failed before.  This fixes it so it will actually complete


markpeek:
Nice! Thanks!

PR #757, opened 2017-06-30
Check that FIFO Queues end with .fifo

Lowercases:
Add a validate() method in sqs.py that checks that FIFO queues
(*) have a QueueName parameter and (*) QueueName ends with '.fifo'
as required by AWS:

http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html

### DIFF
--- a/troposphere/sqs.py
+++ b/troposphere/sqs.py
@@ -34,6 +34,12 @@ class Queue(AWSObject):
         'VisibilityTimeout': (integer, False),
     }
 
+    def validate(self):
+        if self.properties.get('FifoQueue'):
+            if not self.properties.get('QueueName', '').endswith('.fifo'):
+                raise ValueError("SQS: FIFO queues need to provide a "
+                                 "QueueName that ends with '.fifo'")
+
 
 class QueuePolicy(AWSObject):
     resource_type = "AWS::SQS::QueuePolicy"


troposphere/sqs.py
```diff
@@ -34,6 +34,12 @@ class Queue(AWSObject):
         'VisibilityTimeout': (integer, False),
     }
 
+    def validate(self):
+        if self.properties.get('FifoQueue'):
+            if not self.properties.get('QueueName', '').endswith('.fifo'):
```
markpeek: The [docs](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sqs-queues.html#aws-sqs-queue-name) indicate QueueName is not required. But this code would raise an error if FifoQueue was specified without QueueName.

phobologic: Strangely, that's the first thing I thought too when we started using this internally.  But after testing, it turns out that if you specify FifoQueue, you *have* to specify a QueueName, and it has to end with `.fifo`.  It's annoying, but CF isn't smart enough to understand it has to add `.fifo`, which isn't all that surprising after all this time of messing w/ CF :)

markpeek:
Thanks!

PR #758, opened 2017-07-07
Add 'Auth' property to CodeBuild Source

gkrizek:
The Source property in the `AWS::CodeBuild::Project` type can optionally take an `Auth` parameter when the Source Type is 'GITHUB'. This tells CodeBuild to use the OAUTH method of authenticating to private repositories.

There is no direct documentation of this, but they describe it's use in this article:
http://docs.aws.amazon.com/codebuild/latest/userguide/create-project.html#create-project-cli

You can only specify 'Auth' in the Source object when the Type is GITHUB. If you specify Auth in the Source object, the Auth Type can only be 'OAUTH'

### DIFF
--- a/troposphere/codebuild.py
+++ b/troposphere/codebuild.py
@@ -6,6 +6,7 @@
 from . import AWSObject, AWSProperty, Tags
 from .validators import integer, boolean
 
+
 class Auth(AWSProperty):
     props = {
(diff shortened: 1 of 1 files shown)


markpeek:
@gkrizek thanks for the PR. Please note the additional change I made to rename the property.

PR #764, opened 2017-07-12
Ulimit's HardLimit and SoftLimit validator change

hyksos:
Ulimit's HardLimit and SoftLimit validator needs to be integer and not positive_integer.
http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_limits
For example memlock can be -1:-1 as required by:
https://www.elastic.co/guide/en/elasticsearch/reference/5.4/docker.html#_notes_for_production_use_and_defaults

### DIFF
--- a/troposphere/ecs.py
+++ b/troposphere/ecs.py
@@ -1,5 +1,5 @@
 from . import AWSObject, AWSProperty
-from .validators import boolean, network_port, positive_integer, integer
+from .validators import boolean, integer, network_port, positive_integer
 
 
 class Cluster(AWSObject):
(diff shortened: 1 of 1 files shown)


hyksos:
Relevant cloudformation doc indicating type is integer:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdefinitions-ulimit.html

troposphere/ecs.py
```diff
@@ -1,5 +1,5 @@
 from . import AWSObject, AWSProperty
-from .validators import boolean, network_port, positive_integer
+from .validators import boolean, network_port, positive_integer, integer
```
markpeek: Small change: can you alphabetize the imports? (put integer between boolean and network_port)

markpeek:
Thank you for the change and the quick additional edit.

PR #765, opened 2017-07-13
Adding EgressOnlyInternetGatewayId to EC2::Route

tigertoes:


troposphere/ec2.py
```diff
@@ -317,6 +317,7 @@ class Route(AWSObject):
     props = {
         'DestinationCidrBlock': (basestring, False),
         'DestinationIpv6CidrBlock': (basestring, False),
+        'EgressOnlyInternetGatewayId': (basestring, False),
```
markpeek: Would you like to add additional validation? You can follow the same pattern for the Destination* props in the validate method. Per the docs:
```
Required: Conditional. You must specify only one of the following properties: EgressOnlyInternetGatewayId, GatewayId, InstanceId, NatGatewayId, NetworkInterfaceId, or VpcPeeringConnectionId.
```

tigertoes: How does that look?

markpeek:
Excellent. Thank you for the PR and the extra validation.

PR #767, opened 2017-07-14
Allow passing in a dict into DashboardBody

BPScott:
The dict should be converted into a json encoded string

This logic is taken from the [ApiGateway Model](https://github.com/cloudtools/troposphere/blob/d583dfb78b8d207b26070ae1894f7e36a1b39017/troposphere/apigateway.py#L188), whose Schema property also returns a json encoded string.

An example of using this, where both dashboard resources produce the same Cloudformation:

```python
from troposphere.cloudwatch import Dashboard

t = Template()

t.add_resource(Dashboard(
    "dashboard",
    DashboardBody={"c": "d"}
))

t.add_resource(Dashboard(
    "dashboardstring",
    DashboardBody='{"c": "d"}'
))

print(t.to_json())
```

### DIFF
--- a/tests/test_cloudwatch.py
+++ b/tests/test_cloudwatch.py
@@ -1,0 +1,40 @@
+import unittest
+from troposphere.cloudwatch import Dashboard
+
+
+class TestModel(unittest.TestCase):
+    def test_dashboard(self):
+        # Check valid json dashboard string
+        dashboard = Dashboard(
+            "dashboard",
+            DashboardBody='{"a": "b"}',
+        )
+        dashboard.validate()
+
+        # Check invalid json dashboard string
+        dashboard = Dashboard(
+            "dashboard",
+            DashboardBody='{"a: "b"}',
+        )
+        with self.assertRaises(ValueError):
+            dashboard.validate()
+
+        # Check accepting dict and converting to string in validate
+        d = {"c": "d"}
+        dashboard = Dashboard(
+            "dashboard",
+            DashboardBody=d
+        )
+        dashboard.validate()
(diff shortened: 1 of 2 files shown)


markpeek:
I like this addition. It would be good to just refactor the common code into validators.py and change both to use it. Say something like:
```
def json_checker(name, prop):
    if isinstance(prop, basestring):
        # Verify it is a valid json string
        json.loads(prop)
        return prop
    elif isinstance(prop, dict):
        # Convert the dict to a basestring
        return json.dumps(schema)
    elif isinstance(prop, AWSHelperFn):
        return prop
    else:
        raise ValueError("%s must be a str or dict" % name)
```
By returning the prop it can simply be reassigned after the validator runs. Note: the AWSHelperFn still would make sense for Dashboard bodies in case someone wanted to do a Join (for example). Let me know if you want to make these changes. Thanks!

markpeek:
Thanks for the PR. I added the above refactoring.

BPScott:
I was away on holiday this weekend. Thank you for the feedback and making the changes in my absence.

I considered making a generic validator but I  was unsure if the AWSHelperFn part was specific to ApiGateway::Model so erred on the side of leaving it out for DashboardBody. Thanks for the confirmation that is should go in. Heck that general unfamiliarity with troposphere implementation is why I raised the dashboard stuff as an issue rather than a PR in the first place, [it turns out it was as simple as I thought it might be](https://github.com/cloudtools/troposphere/commit/122e9d33f2ef8e4d72104f84b2208618c7a98387), I now know for next time!

PR #768, opened 2017-07-15
Relax group name constraints to what they are in AWS.

jacob-meacham:
Currently, the Group titles are validated to the more strict alphanumeric regex. However, Groups in AWS can also contain any of +=,.@-_.

### DIFF
--- a/troposphere/iam.py
+++ b/troposphere/iam.py
@@ -2,6 +2,7 @@
 # All rights reserved.
 #
 # See LICENSE file for full license.
+import re
 
 from . import AWSObject, AWSProperty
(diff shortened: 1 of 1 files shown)


markpeek:
Just to follow up, the titles are required to be a stricter alphanumeric set. The IAM names (in properties) allow for the expanded regex through this [function](https://github.com/cloudtools/troposphere/blob/master/troposphere/validators.py#L134) called during property validation.

jacob-meacham:
Yeah, I meant to leave a comment - I realized it was my mistake, which is why I closed the PR. Thanks!

PR #785, opened 2017-08-15
Update PyPI Information

gkrizek:
I noticed that the MANIFEST.in file was referencing a Markdown README instead of a reStructuredText README.

Also, the PyPI website for this library has practically no information. I think it would be beneficial to copy the README.rst file over to the PyPI page as well.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,4 @@
 include LICENSE
-include README.md
+include README.rst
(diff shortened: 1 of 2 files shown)


gkrizek:
@markpeek  These are some housekeeping edits. Feel free to close if you think they aren't necessary.

markpeek:
Thanks!

PR #786, opened 2017-08-21
Adding IPv6 changes to AWS::EC2::Subnet

tigertoes:


tigertoes:
Oh, err, it appears my validation (despite not being complete) is annoying the linter, any suggestions on how I might better approach this?

ikben:
I haven't tested it, but something like this?

```
   def validate(self):
        if 'Ipv6CidrBlock' in self.properties:
            if not self.properties.get('AssignIpv6AddressOnCreation'):
                raise ValueError(
                    "If Ipv6CidrBlock is present, "
                    "AssignIpv6AddressOnCreation must be set to true"
                )
```

markpeek:
Thanks!

PR #787, opened 2017-08-22
NetworkACL Protocl Constants

thebenwaters:
Added some basic constants for adding NetworkACL Protocols from here https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml


markpeek:
Thanks!

PR #790, opened 2017-08-23
Add AWS::ApiGateway::GatewayResponse

marcindobry:
Adds support for [AWS::ApiGateway::GatewayResponse](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-gatewayresponse.html). 

Unfortunately, there seems to currently be a bug in CloudFormation which blocks creating GatewayResponses (https://forums.aws.amazon.com/thread.jspa?messageID=801223). However, I think it's still worth adding support for creating these objects according to the spec in the docs.

It's my first PR and I'm not sure about the conventions in the project so please let me know if anything needs to be changed accordingly.

### DIFF
--- a/tests/test_apigateway.py
+++ b/tests/test_apigateway.py
@@ -1,6 +1,6 @@
 import unittest
 from troposphere import Join
-from troposphere.apigateway import Model
+from troposphere.apigateway import GatewayResponse, Model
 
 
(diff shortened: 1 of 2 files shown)


markpeek:
Thank you for sending in a PR and adding the test. Everything looked good to me.

PR #791, opened 2017-08-23
Add support for aurora-postgresql as a valid DB engine

jcragg:
Aurora PostgreSQL is currently in preview and is accessible via CloudFormation. This requires aurora-postgresql to be a valid DB engine selection.


markpeek:
Thanks!

PR #793, opened 2017-08-26
adding sqs server side encryption

patrickpierson:
Based on https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sqs-queues.html#aws-sqs-queue-kmsmasterkeyid

Tested new example and SQS queue is created with the specified encryption key.


markpeek:
Thanks!

PR #800, opened 2017-09-05
Missing new cloudfront custom origin timeouts

alangarf:
New feature added to cloudfront to allow custom read and keepalive timeouts.


alangarf:
API seems to not like, even though documented: http://docs.aws.amazon.com/cloudfront/latest/APIReference/API_CustomOriginConfig.html

jacekszubert:
I've just tested it and seems to work fine. Can we merge it?

markpeek:
Those properties are still not [documented](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-customorigin.html). I'll have to follow up with AWS.

PR #808, opened 2017-09-12
Revert "Accept Join type as parameter default value as it returns a s…

stan3:
…tring (#752)"

This reverts commit 3b78644106368c195399477f2e47552f4497bd0d.

This was a incorrect change - cloudformation also rejects it with a validation error - "Template validation error: Template format error: Every Default member must be a string."


markpeek:
Thanks!

PR #809, opened 2017-09-12
Change Cognito UserPool SchemaAttribute "Required" value to boolean from basestring

ghTravis:
SchemaAttribute documentation mentions that the "Required" field is a boolean, similar to the "Mutable" field. Wasn't sure if it was just an oversight but to avoid confusion I thought it should also accept plain boolean values.

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cognito-userpool-schemaattribute.html


markpeek:
Looks correct. Thanks!

PR #810, opened 2017-09-13
Updating case of 'AssignIPv6AddressOnCreation'

tigertoes:
Turns out the AWS documentation was incorrect, and that "P" was supposed to be upper case (despite the fact the underlying API does not do this).


markpeek:
Thanks!

justjulian:
Hi @tigertoes it looks like AWS updated it, I get an error in CF when I try to create a subnet:
`Encountered unsupported property AssignIPv6AddressOnCreation`

PR #813, opened 2017-09-19
EFS example: SecurityGroupRule can't be referred to as a Ref

mlei:
Because the `SecurityGroupRule` isn't added into the template, `Ref` doesn't work here. The `SecurityGroup` constructor should refer to the `SecurityGroupRule` object directly.


markpeek:
Thanks!

PR #821, opened 2017-09-24
adding autoscaling example for http requests that closes #630

### LINKED ISSUE #630: Autoscaling example?
Has anyone written a example of autoscaling for ec2 instances based on http requests? The default out of the box from AWS revolves around CPU usage which isn't ideal for web servers.

patrickpierson:
@BardiaAfshin requested someone write an example that does autoscaling based on the number of requests an ELB gets. This is an example I wrote. Most of it was copied from the original autoscaling example but with scaling added that uses the Request Count metric from the ELB's cloudwatch metrics.


markpeek:
Excellent. Thanks!

PR #822, opened 2017-09-27
Update serverless.py

cb17:
Adding "DefinitionBody" to serverless::api. Not sure if this is the right type (dict) as the body can either be a YAML or JSON object. Also not sure how to do either/or (True/False) as you can specify either DefinitionBody _or_ DefinitionUri. Would appreciate getting this added to the next release ASAP, as this is a blocking defect. Thanks!

### DIFF
--- a/tests/test_serverless.py
+++ b/tests/test_serverless.py
@@ -15,11 +15,46 @@ class TestServerless(unittest.TestCase):
         t.add_resource(serverless_func)
         t.to_json()
 
-    def test_required_api(self):
+    def test_required_api_definitionuri(self):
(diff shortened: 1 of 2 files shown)


michael-robbins:
See https://github.com/cloudtools/troposphere/blob/master/troposphere/ec2.py#L278 on how to perform an OR validation!

michael-robbins:
As far as the type goes, can you test it yourself? Does CloudFormation expect an object, or a serialised object (string)?

markpeek:
Thanks! I did push validation logic and some tests onto your branch.

PR #823, opened 2017-09-28
Adding kinesis stream source to firehose

mathom:
The `KinesisStreamSourceConfiguration` is missing from `firehose.py`. Official docs are [here](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-kinesisfirehose-deliverystream.html).

One thing to note is `DeliveryStreamType` is missing from the docs but is present in the examples - it's required to make the source configuration work.

### DIFF
--- a/troposphere/firehose.py
+++ b/troposphere/firehose.py
@@ -136,13 +136,22 @@ class ExtendedS3DestinationConfiguration(AWSProperty):
     }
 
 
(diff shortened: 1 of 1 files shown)


markpeek:
Issue #831

markpeek:
Thanks!

Issue #832, opened 2017-09-30
Add AccelerateConfiguration to AWS::S3::Bucket


markpeek:
[AWS::S3::Bucket](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket.html)
Use the AccelerateConfiguration property to configure the transfer acceleration state for an Amazon S3 bucket.

danielpops:
Seems straightforward enough, I'll take a look. Looks like there are only two supported values, either `Enabled` or `Suspended`.

markpeek:
Closed via #833

PR #833, opened 2017-09-30
Add S3 Transfer Acceleration to AWS::S3::Bucket

danielpops:
This addresses #832

Example output matches the spec at http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-accelerateconfiguration.html

### DIFF
--- a/troposphere/s3.py
+++ b/troposphere/s3.py
@@ -46,6 +46,22 @@ class VersioningConfiguration(AWSProperty):
     }
 
 
+class AccelerateConfiguration(AWSProperty):
+    props = {
+        'AccelerationStatus': (basestring, True),
+    }
+    acceleration_status_values = [
+        'Enabled',
+        'Suspended',
+    ]
+
+    def validate(self):
+        if 'AccelerationStatus' in self.properties:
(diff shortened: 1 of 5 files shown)


troposphere/s3.py
```diff
@@ -46,6 +46,23 @@ class VersioningConfiguration(AWSProperty):
...
+class AccelerateConfiguration(AWSProperty):
+    props = {
+        'AccelerationStatus': (basestring, True),
+    }
+    acceleration_status_values = [
+        'Enabled',
+        'Suspended',
+    ]
+
+    def validate(self):
```
phobologic: I'd remove this, and instead write a custom validator here that does the `acceleration_status_values`:

```
def validate_acceleration_status(value):
    valid_status = ["Enabled", "Suspended"]
    if value not in valid_status:
        raise ValidationError("AccelerationStatus must be one of: %s", ", ".join(valid_status))
    return value
```

The benefit of this is that it will take into account non-string values, such as Functions like `Ref()` (say someone wanted to pass this as a Parameter, for example, that'd be a `Ref`).

Then you can just use that validator in place of `basestring`.

PR #844, opened 2017-10-07
Refactor s3.Bucket to remove custom __init__() and add tests

### LINKED ISSUE #841: S3 Bucket constructor
**Expected result:** troposphere.s3.Bucket constructor accepts `template` as a second parameter.
**Actual result:** troposphere.s3.Bucket constructor does not accept `template` as a second parameter.

Example of code that should work:
```python
template = Template()
Bucket(
    'PlaylistsFailoverS3Bucket',
    template,
    AccessControl=Private,
    BucketName="test-bucket",
)
```

markpeek:
Fixes #841

### DIFF
--- a/tests/s3.py
+++ b/tests/s3.py
@@ -1,0 +1,60 @@
+import unittest
+from troposphere import Template
+from troposphere.s3 import AccelerateConfiguration, Bucket, Private
+
+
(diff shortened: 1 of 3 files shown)


Tonkonozhenko:
@markpeek thanks!

PR #852, opened 2017-10-14
Add support for serverless S3Location CodeUri

### LINKED ISSUE #848: Specify CodeUri as a combination of bucket and key in SAM
Hi, is it possible to use `Bucket` and `Key` in the `CodeUri` in SAM transformation. Currently, it only support passing `CodeUri` as a string.     
Example: [Serverless_S3_Processor.py](https://github.com/cloudtools/troposphere/blob/master/examples/Serverless_S3_Processor.py)
```json
CodeUri='s3://<bucket>/s3_processor.zip'
```
Thanks

markpeek:
Fixes #848

### DIFF
--- a/tests/test_serverless.py
+++ b/tests/test_serverless.py
@@ -1,9 +1,23 @@
 import unittest
 from troposphere import Template
-from troposphere.serverless import Function, Api, SimpleTable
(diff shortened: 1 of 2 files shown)


markpeek:
Added via #853

PR #853, opened 2017-10-14
Add Tags, Tracing, KmsKeyArn, DLQ to serverless(SAM)

Laxenade:
SAM(Serverless) is missing some of the function properties, add following properties to the template:
* Tags
* Tracing
* KmsKeyArn
* DeadLetterQueue
* CodeUri, This has been covered in another PR [#852](https://github.com/cloudtools/troposphere/pull/852)

### DIFF
--- a/tests/test_serverless.py
+++ b/tests/test_serverless.py
@@ -1,9 +1,39 @@
 import unittest
 from troposphere import Template
-from troposphere.serverless import Function, Api, SimpleTable
+from troposphere.serverless import Function, Api, SimpleTable, DeadLetterQueue
 
 
 class TestServerless(unittest.TestCase):
+    def test_tags(self):
+        serverless_func = Function(
+            "SomeHandler",
+            Handler="index.handler",
+            Runtime="nodejs",
+            CodeUri="s3://bucket/handler.zip",
+            Tags={
+                'Tag1': 'TagValue1',
+                'Tag2': 'TagValue2'
+            }
+        )
+        t = Template()
+        t.add_resource(serverless_func)
+        t.to_json()
+
+    def test_DLQ(self):
+        serverless_func = Function(
+            "SomeHandler",
+            Handler="index.handler",
+            Runtime="nodejs",
+            CodeUri="s3://bucket/handler.zip",
+            DeadLetterQueue=DeadLetterQueue(
+                Type='SNS',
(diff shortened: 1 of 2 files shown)


tests/test_serverless.py
```diff
@@ -1,9 +1,39 @@
 import unittest
 from troposphere import Template
-from troposphere.serverless import Function, Api, SimpleTable
+from troposphere.serverless import Function, Api, SimpleTable, DeadLetterQueue
```
markpeek: This should be alphabetized (yeah, I see it wasn't originally either).

troposphere/serverless.py
```diff
@@ -31,6 +31,19 @@ def policy_validator(x):
...
 
+class DeadLetterQueue(AWSProperty):
+    props = {
+        'Type': (basestring, False),
+        'TargetArn': (basestring, False)
+    }
+
+    def validate(self):
+        valid_types = ['SQS', 'SNS']
+        if self.properties['Type'] not in valid_types:
```
markpeek: Type is not required so you need to check for existence before dereferencing it. 
```
if 'Type' in self.properties and \
    self.properties['Type'] not in valid_types:
...
```

troposphere/serverless.py
```diff
@@ -31,6 +31,19 @@ def policy_validator(x):
...
+            raise ValueError('Type must be either SQS or SNS')
+        return self.properties['Type']
```
markpeek: Property validators do return a type but object validators (like this one) don't return anything.

troposphere/serverless.py
```diff
@@ -45,7 +58,11 @@ class Function(AWSObject):
         'Policies': (policy_validator, False),
         'Environment': (Environment, False),
         'VpcConfig': (VPCConfig, False),
-        'Events': (dict, False)
+        'Events': (dict, False),
```
markpeek: The properties should be alphabetized (again, looks like they weren't to begin with).

Laxenade: This is in the same order as defined [here](https://github.com/awslabs/serverless-application-model/blob/7067ff112ce762d97838e824ec97b9b25e453498/versions/2016-10-31.md). I do prefer this order though, this way you could follow SAM's example more easily. (So you don't have to rearrange them in your mind to get something to work)

markpeek:
Nice job. Thanks!

Issue #850, opened 2017-10-12
Add new AWS::SSM resources


markpeek:
[AWS::SSM::MaintenanceWindow](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-maintenancewindow.html)
Use the AWS::SSM::MaintenanceWindow resource to create an Amazon EC2 Systems Manager Maintenance Window.

[AWS::SSM::MaintenanceWindowTarget](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-maintenancewindowtarget.html)
Use the AWS::SSM::MaintenanceWindowTarget resource to register a target with a Maintenance Window.

[AWS::SSM::MaintenanceWindowTask](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-maintenancewindowtask.html)
Use the AWS::SSM::MaintenanceWindowTask resource to define a Maintenance Window task.

[AWS::SSM::PatchBaseline](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-patchbaseline.html)
Use the AWS::SSM::PatchBaseline resource to define a Systems Manager patch baseline.

codevbus:
I can grab some.

codevbus:
See #851 

Assuming that looks good/is valid, I can send another PR for the remaining 3.

markpeek:
Looked great and merged. Feel free to PR the remaining 3.

Laxenade:
I can add one or two if no one is working on it.

codevbus:
@Laxenade I've changes for the additional 3 completed, just working on polishing up the unit tests. I'll be submitting my PR this afternoon.

markpeek:
Fixed via #851 and #854

PR #854, opened 2017-10-14
Aws ssm resource features

codevbus:
completes feature work on #850


markpeek:
Thank you for adding these new resources. I did a couple of merge and test tweaks before pulling it in so feel free to verify I got it right.

Fixes #850

Issue #861, opened 2017-10-19
Add support for VPCOptions in ElasticSearch


Hyunk3l:
hello 😄 
some days ago AWS announced support for VPCs in ElasticSearch service:
https://aws.amazon.com/es/blogs/aws/amazon-elasticsearch-service-now-supports-vpc/

Here is an example from official documentation:
```
POST /2015-01-01/es/domain
{
    "DomainName": "<DOMAIN_NAME>",
    "ElasticsearchVersion": "<VERSION>",
    "ElasticsearchClusterConfig": {
        "InstanceType": "<INSTANCE_TYPE>",
        "InstanceCount": <INSTANCE_COUNT>,
        "DedicatedMasterEnabled": "<TRUE|FALSE>",
        "DedicatedMasterCount": <INSTANCE_COUNT>,
        "DedicatedMasterType": "<INSTANCE_TYPE>",
        "ZoneAwarenessEnabled": "<TRUE|FALSE>"
    },
    ...
    "VPCOptions": {
        "SubnetIds": [
            "<SUBNET_ID>"
        ],
        "SecurityGroupIds": [
            "<SECURITY_GROUP_ID>"
        ]
    },
    ...
}
```
Is there any plan for adding support for VPCOptions?

Meanwhile, thanks for the fantastic job you are doing!

thanks!

craigbruce:
Unfortunately, AWS hasn't added this the CloudFormation API, see http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticsearch-domain.html, so it isn't accessible to CloudFormation and by extension Troposphere.

To encourage them, please submit a feature request (the more they get, hopefully the quicker they will implement it).

Hyunk3l:
I was looking in the wrong place http://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/es-configuration-api.html#es-configuration-api-datatypes-vpcoptions 😢 
thanks @craigbruce

cintiadr:
While this is not yet documented, I did get the impression that this should be already working (keep in mind I haven't tested).

stevengssns:
Fyi, the documentation has been updated, and it works.

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticsearch-domain.html

Hyunk3l:
PR updated :) thanks 4 the spot!

markpeek:
[...]

PR #862, opened 2017-10-19
Add support for VPCOptions in ElasticSearch

Hyunk3l:
Hello :)

I'm trying to solve this issue I opened some minutes ago https://github.com/cloudtools/troposphere/issues/861

The idea is to add support for VPCOptions in ElasticSearch (you will find more info in the issue).

Hope this works!

### DIFF
--- a/examples/ElasticsearchDomain.py
+++ b/examples/ElasticsearchDomain.py
@@ -2,7 +2,7 @@
 # http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticsearch-domain.html#d0e51519
 
 from troposphere import Template, constants
-from troposphere.elasticsearch import Domain, EBSOptions
+from troposphere.elasticsearch import Domain, EBSOptions, VPCOptions
 from troposphere.elasticsearch import ElasticsearchClusterConfig
 from troposphere.elasticsearch import SnapshotOptions
 
@@ -37,13 +37,10 @@ es_domain = templ.add_resource(Domain(
                         'Resource': '*'
                     }]},
     AdvancedOptions={"rest.action.multi.allow_explicit_index": "true"},
-    VPCOptions={'SubnetIds': [
-            {"Ref": "subnet-4f2bb123"}
-        ],
-        'SecurityGroupIds': [
-            {"Ref": "sg-04cf048c"}
-        ]
-    }
+    VPCOptions=VPCOptions(
+       SubnetIds=["subnet-4f2bb123"],
+       SecurityGroupIds=["sg-04cf048c"]
+    )
 ))
 
 print(templ.to_json())
--- a/examples/ElasticsearchDomain.py
+++ b/examples/ElasticsearchDomain.py
@@ -36,7 +36,14 @@ es_domain = templ.add_resource(Domain(
                         'Action': 'es:*',
                         'Resource': '*'
                     }]},
-    AdvancedOptions={"rest.action.multi.allow_explicit_index": "true"}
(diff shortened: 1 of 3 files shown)


Hyunk3l:
As @craigbruce pointed out, seems that CloudFormation does not support VCPOptions https://github.com/cloudtools/troposphere/issues/861#issuecomment-338049650
Should I close this PR or wait for them to implement the option?

markpeek:
I created a new label for it so we can leave it here in case it gets implemented soon in CF. Thanks.

markpeek:
@Hyunk3l I see this got implemented although it isn't a dict but rather a specific property type. Docs are [here](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticsearch-domain-vpcoptions.html). Can you update this PR? Thanks.

troposphere/elasticsearch.py
```diff
@@ -66,9 +66,15 @@ class Domain(AWSObject):
         'ElasticsearchClusterConfig': (ElasticsearchClusterConfig, False),
         'ElasticsearchVersion': (basestring, False),
         'SnapshotOptions': (SnapshotOptions, False),
-        'Tags': ((Tags, list), False)
+        'Tags': ((Tags, list), False),
+        'VPCOptions': (VPCOptions, False)
     }
 
+class VPCOptions(AWSProperty):
```
Hyunk3l: @markpeek `AWSProperty`, right?

troposphere/elasticsearch.py
```diff
@@ -55,6 +55,13 @@ class SnapshotOptions(AWSProperty):
     }
 
 
+class VPCOptions(AWSProperty):
+    props = {
+        "SecurityGroupIds": (dict, False),
```
markpeek: These should be list of strings:
    "SecurityGroupIds": ([basestring], False),
    "SubnetIds": ([basestring], False)

examples/ElasticsearchDomain.py
```diff
@@ -36,7 +36,14 @@
                         'Action': 'es:*',
                         'Resource': '*'
                     }]},
-    AdvancedOptions={"rest.action.multi.allow_explicit_index": "true"}
+    AdvancedOptions={"rest.action.multi.allow_explicit_index": "true"},
+    VPCOptions={'SubnetIds': [
```
markpeek: This should be (and you'll need an import at the top for VPCOptions):
VPCOptions=VPCOptions(
    SubnetIds=["subnet-4f2bb123"],
    SecurityGroupIds=["sg-04cf048c"],
),

tests/examples_output/ElasticsearchDomain.template
```diff
@@ -36,6 +36,14 @@
                 },
                 "SnapshotOptions": {
                     "AutomatedSnapshotStartHour": 0
+                },
+                "VPCOptions": {
```
markpeek: Fix this up given the change above.

examples/ElasticsearchDomain.py
markpeek: Also, the "Ref" is a bit of a misnomer given the subnets/sg's aren't defined in this example. Really you're accessing already defined subnets/security groups.

troposphere/elasticsearch.py
Hyunk3l: You are totally right!

Hyunk3l:
Thanks @markpeek! All changes are done and tests are green :)

markpeek:
Nice! Thank you for implementing this addition.

Hyunk3l:
Thank you for your support! 😄

PR #866, opened 2017-10-23
Set EC2 BlockDeviceMapping NoDevice property to type dict

### LINKED ISSUE #865: EC2 NoDevice error with boolean values.
It appears that the `NoDevice` field in the `BlockDeviceMapping` was updated in #858 to be a boolean.  Unfortunately now, the generated templates are failing to deploy, regardless of what value you use for the field.  When set to either True or False, I receive the following error when trying to deploy a generated template via CloudFormation: `Value of property NoDevice must be an object`

Looking at the [docs](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-blockdev-mapping.html#cfn-ec2-blockdev-mapping-nodevice)  the type is listed as boolean, however the last example on the page `Unmapping an AMI-defined Device` shows the use of an empty map, `{}`, and not a string or boolean value.

I saw that @tweakster noted he received a error stating this should be a string, but I have tried multiple regions, and always receive the same error if an empty map is not used.

We have been using this value successfully in previous versions, and I propose that the change be reverted.  I will work on submitting the required code change, later today, and can provide a sample of working code using this empty mapping as well.

johnctitus:
Reverts EC2->BlockDeviceMapping->NoDevice property to type dict, allowing for a value of `{}` as shown in the [documentation](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-blockdev-mapping.html#cfn-ec2-blockdev-mapping-nodevice) under the `Unmapping an AMI-defined Device` example.

Also adds an example template of unmapping ephemeral devices on an EC2 instance.

Fixes #865

### DIFF
--- a/examples/EC2_Remove_Ephemeral_Drive.py
+++ b/examples/EC2_Remove_Ephemeral_Drive.py
@@ -1,4 +1,6 @@
 from troposphere import FindInMap, GetAtt, Output, Parameter, Ref, Template
+from troposphere.ec2 import NO_DEVICE
+
 import troposphere.ec2 as ec2
 
 
@@ -37,10 +39,10 @@ ec2_instance = template.add_resource(ec2.Instance(
     BlockDeviceMappings=[
         ec2.BlockDeviceMapping(
             DeviceName='/dev/sdb',
-            NoDevice={}),
+            NoDevice=NO_DEVICE),
         ec2.BlockDeviceMapping(
             DeviceName='/dev/sdc',
-            NoDevice={}),
+            NoDevice=NO_DEVICE),
     ]
 ))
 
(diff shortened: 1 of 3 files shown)


examples/EC2_Remove_Ephemeral_Drive.py
```diff
@@ -0,0 +1,80 @@
...
+    InstanceType="t2.micro",
+    KeyName=Ref(keyname_param),
+    SecurityGroups=["default"],
+    BlockDeviceMappings=[
+        ec2.BlockDeviceMapping(
+            DeviceName='/dev/sdb',
+            NoDevice={}),
+        ec2.BlockDeviceMapping(
+            DeviceName='/dev/sdc',
+            NoDevice={}),
```
markpeek: Would it make sense to add a NO_DEVICE constant in ec2.py?

johnctitus: I think that's a great idea actually.  I'll push it up shortly.

johnctitus:
`NO_DEVICE` constant has also been added, and the example template updated.

markpeek:
Thanks!

PR #867, opened 2017-10-24
Change AWS::Batch::ComputeResources.Tags type to dict

flou:
AWS Batch seems to use a different Tag format. Instead of the usual
```
{
  'Key': 'TagKey',
  'Value': 'TagValue'
}
```
AWS Batch expects a regular map `{'TagKey': 'TagValue'}`.

This commit also adds an example template for AWS Batch.

### DIFF
--- a/troposphere/batch.py
+++ b/troposphere/batch.py
@@ -10,13 +10,13 @@ class ComputeResources(AWSProperty):
         "SecurityGroupIds": ([basestring], True),
         "BidPercentage": (positive_integer, False),
         "Type": (basestring, True),
-        "Subnets": ([basestring], True),
+        "Subnets": (list, True),
         "MinvCpus": (positive_integer, True),
         "ImageId": (basestring, False),
         "InstanceRole": (basestring, True),
         "InstanceTypes": ([basestring], True),
         "Ec2KeyPair": (basestring, False),
-        "Tags": ([basestring], False),
+        "Tags": (dict, False),
         "DesiredvCpus": (positive_integer, False)
     }
 
--- a/troposphere/batch.py
+++ b/troposphere/batch.py
@@ -10,7 +10,7 @@ class ComputeResources(AWSProperty):
         "SecurityGroupIds": ([basestring], True),
         "BidPercentage": (positive_integer, False),
         "Type": (basestring, True),
-        "Subnets": (list, True),
+        "Subnets": ([basestring], True),
         "MinvCpus": (positive_integer, True),
         "ImageId": (basestring, False),
         "InstanceRole": (basestring, True),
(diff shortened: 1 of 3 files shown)


troposphere/batch.py
```diff
@@ -10,13 +10,13 @@ class ComputeResources(AWSProperty):
         "SecurityGroupIds": ([basestring], True),
         "BidPercentage": (positive_integer, False),
         "Type": (basestring, True),
-        "Subnets": ([basestring], True),
+        "Subnets": (list, True),
```
markpeek: Any reason why this was changed? Using [basestring] is a list of basestrings which is more specific than just using list.

troposphere/batch.py
```diff
@@ -10,13 +10,13 @@ class ComputeResources(AWSProperty):
...
         "MinvCpus": (positive_integer, True),
         "ImageId": (basestring, False),
         "InstanceRole": (basestring, True),
         "InstanceTypes": ([basestring], True),
         "Ec2KeyPair": (basestring, False),
-        "Tags": ([basestring], False),
+        "Tags": (dict, False),
```
markpeek: The [documentation](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-computeenvironment-computeresources.html#cfn-batch-computeenvironment-computeresources-tags) is all over the map on this one as it is referring to the Tags as being normal [Tags](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-resource-tags.html). I'll have to follow up with my AWS contacts to understand why there is a discrepancy.

troposphere/batch.py
flou: In a previous version of this PR, the example template used a `CommaDelimitedList` for the `Subnets` parameter type and the build failed for Python 2.7 (see https://travis-ci.org/cloudtools/troposphere/jobs/292160894). Not sure why it failed though…

So I figured I'd use the `list` property type as in `AWS::RDS::DBSubnetGroup.SubnetIds` or `AWS::ElasticLoadBalancing::LoadBalancer.Subnets`.

troposphere/batch.py
flou: I understand. It took me a while to figure out how AWS Batch tags work, but when using normal Tags CloudFormation gives the following message:
```
Property validation failure: [Value of property {/ComputeResources/Tags} does not match type {Map}]
```
which is very weird when comparing with the docs.

flou: Do you have any news about this?

markpeek:
Thanks! (Note: I did revert the Subnet change)

PR #869, opened 2017-10-24
Allow AWSHelperFn for CodeCommit Trigger Event(s)

### LINKED ISSUE #868: Validation error in CodeCommit Trigger
It appears that the validation for the CodeCommit Trigger class does not currently allow the use of any intrinsic functions for the `Events` property, and only accepts a list of hard coded values.

https://github.com/cloudtools/troposphere/blob/45582eb1d21a6cc9cfa608f626d8acbf0317f37f/troposphere/codecommit.py#L18-L32

We are trying to allow for the dynamic selection of  the values with a  parameter, but encounter errors when attempting to use a `Ref`.  A snippet of our trigger definition, and the error encountered is below:

```
repo_trigger1 = codecommit.Trigger(
            Name = Ref(trigger_1_name),
            CustomData = Ref(trigger_1_custom_data),
            DestinationArn = Ref(trigger_1_destination_arn),
            Branches = Ref(trigger_1_branches),
            Events = Ref(trigger_1_events),
        )
```
We are able to successfully generate the template when changing the `Events` assignment to:

```Events = ["all"],```

I believe we just need to check if the value is one of the Helper functions before iterating through the events.  I will try to get a fix pushed up for review.

johnctitus:
Updates `Trigger` validation to also accept `AWSHelperFn` objects or a list of `AWSHelperFn` objects.

Fixes #868

### DIFF
--- a/troposphere/codecommit.py
+++ b/troposphere/codecommit.py
@@ -3,7 +3,7 @@
 #
 # See LICENSE file for full license.
 
-from . import AWSObject, AWSProperty
+from . import AWSHelperFn, AWSObject, AWSProperty
 
 
 class Trigger(AWSProperty):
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks!

PR #876, opened 2017-10-31
support ProcessingConfiguration for Elasticsearch and Redshift

ericblade:


troposphere/firehose.py
```diff
@@ -60,6 +60,26 @@ class CopyCommand(AWSProperty):
...
+    props = {
+        'ParameterName': (basestring, True),
+        'ParameterValue': (basestring, True),
+    }
+
+
+class Processor(AWSProperty):
+    props = {
+        'Parameters': ([ProcessorParameter], True),
+        'Type': (basestring, True),
```
phobologic: it might be worth a custom validator here, since the only valid type it appears is `Lambda`:

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-kinesisfirehose-deliverystream-processor.html

ericblade: I went ahead and added validator functions for all of the string type enums that I noticed in the docs files for these APIs.

PR #877, opened 2017-11-02
Fixes incorrect class definition.

stevengorrell:
Hello,

The template created by this class fails on AWS deployment:

```
Property validation failure: [Encountered unsupported properties in {/ApprovalRules/PatchRules/0/PatchFilterGroup}: [Values, Key]]
```

The expected property from the AWS side is the group property. This update fixes and creates a valid CloudFormation template.


markpeek:
Nice catch. Thanks!

PR #894, opened 2017-11-04
Added support for VpnTunnelOptionsSpecifications

cigan1:
Support for VpnTunnelOptionsSpecifications allows specification of PreSharedKey and TunnelInsideCidr.

### DIFF
--- a/troposphere/ec2.py
+++ b/troposphere/ec2.py
@@ -6,7 +6,7 @@
 from . import AWSHelperFn, AWSObject, AWSProperty, Tags
 from .validators import (
     boolean, exactly_one, integer, integer_range,
-    network_port, positive_integer
+    network_port, positive_integer, tunnel_inside_cidr
 )
 
 try:
@@ -566,7 +566,7 @@ class VPCGatewayAttachment(AWSObject):
 class VpnTunnelOptionsSpecification(AWSProperty):
     props = {
         'PreSharedKey': (basestring, False),
-        'TunnelInsideCidr': (basestring, False),
+        'TunnelInsideCidr': (tunnel_inside_cidr, False),
     }
 
 
--- a/troposphere/ec2.py
+++ b/troposphere/ec2.py
@@ -563,6 +563,13 @@ class VPCGatewayAttachment(AWSObject):
     }
 
 
+class VpnTunnelOptionsSpecification(AWSProperty):
+    props = {
+        'PreSharedKey': (basestring, False),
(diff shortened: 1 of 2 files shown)


troposphere/ec2.py
```diff
@@ -563,6 +563,13 @@ class VPCGatewayAttachment(AWSObject):
     }
 
 
+class VpnTunnelOptionsSpecification(AWSProperty):
+    props = {
+        'PreSharedKey': (basestring, False),
+        'TunnelInsideCidr': (basestring, False)
```
markpeek: Best to add trailing commas.

troposphere/ec2.py
```diff
@@ -572,6 +579,9 @@ class VPNConnection(AWSObject):
         'StaticRoutesOnly': (boolean, False),
         'Tags': ((Tags, list), False),
         'VpnGatewayId': (basestring, True),
+        'VpnTunnelOptionsSpecifications': (
+            [VpnTunnelOptionsSpecification], False
+        )
```
markpeek: Best to add trailing commas.

troposphere/ec2.py
```diff
@@ -563,6 +563,13 @@ class VPCGatewayAttachment(AWSObject):
     }
 
 
+class VpnTunnelOptionsSpecification(AWSProperty):
+    props = {
```
markpeek: Would you want to add validators for these two properties as well?

PreSharedKey
Constraints: Allowed characters are alphanumeric characters and ._. Must be between 8 and 64 characters in length and cannot start with zero (0).

TunnelInsideCidr
Constraints: A size /30 CIDR block from the 169.254.0.0/16 range. The following CIDR blocks are reserved and cannot be used:

169.254.0.0/30
169.254.1.0/30
169.254.2.0/30
169.254.3.0/30
169.254.4.0/30
169.254.5.0/30
169.254.169.252/30

troposphere/ec2.py
cigan1: Added.

troposphere/ec2.py
cigan1: Added.

troposphere/ec2.py
cigan1: I will work on this.

cigan1:
Added validator for TunnelInsideCidr.

troposphere/ec2.py
cigan1: @markpeek validator added.

Issue #884, opened 2017-11-03
Update CodeDeploy resource


markpeek:
[AWS::CodeDeploy::DeploymentGroup](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-codedeploy-deploymentgroup.html)

In the [`LoadBalancerInfo`](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codedeploy-deploymentgroup-loadbalancerinfo.html) property type, use the `TargetGroupInfo` property to specify information about a target group in Elastic Load Balancing to use in a deployment.

markpeek:
Fixed via #895

PR #895, opened 2017-11-04
add TargetGroupInfo to DeploymentGroup #884

ScOut3R:
Hi,

my Python is sysadmin-like so this PR might got the solution totally wrong. The reason I'm raising is I just hit this missing feature today and thought I'd give it a try to implement it.

### DIFF
--- a/troposphere/codedeploy.py
+++ b/troposphere/codedeploy.py
@@ -4,7 +4,7 @@
 # See LICENSE file for full license.
 
 from . import AWSObject, AWSProperty
-from .validators import boolean, positive_integer
+from .validators import boolean, positive_integer, exactly_one
 
 
 KEY_ONLY = "KEY_ONLY"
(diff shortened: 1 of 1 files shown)


troposphere/codedeploy.py
```diff
@@ -4,7 +4,7 @@
 # See LICENSE file for full license.
 
 from . import AWSObject, AWSProperty
-from .validators import boolean, positive_integer
+from .validators import boolean, positive_integer, exactly_one
```
markpeek: Please alphabetize.

troposphere/codedeploy.py
```diff
@@ -80,11 +80,25 @@ class ElbInfoList(AWSProperty):
...
+    props = {
+        'Name': (basestring, False)
+    }
+
+
 class LoadBalancerInfo(AWSProperty):
     props = {
-        'ElbInfoList': ([ElbInfoList], False)
+        'ElbInfoList': ([ElbInfoList], False),
+        'TargetGroupInfoList': ([TargetGroupInfoList], False)
```
markpeek: I usually always add commas at the end of these lists to make future additions a 1 line change.

troposphere/codedeploy.py
ScOut3R: Done.

troposphere/codedeploy.py
ScOut3R: Thank you, I'm always happy to learn best practices.

markpeek:
Thanks! (Fixes #884)

Issue #889, opened 2017-11-03
Update KMS::Key


markpeek:
[AWS::KMS::Key](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-kms-key.html)

Use the `Tags` property to specify an arbitrary set of tags (key--value pairs) to associate with a custom master key (CMS).

qassim:
Raised PR https://github.com/cloudtools/troposphere/pull/900

markpeek:
Fixed via #900

PR #900, opened 2017-11-10
Add support for tags in AWS::KMS::Key.

qassim:
AWS::KMS::Key doesn't currently support the Tags property. This change add this functionality. 

Fixes issue: #889


markpeek:
Thanks!

markpeek:
@cigan1 I also added your suggested change in a subsequent commit. Thanks.

PR #905, opened 2017-11-15
Allow ref on Parameter

jappievw:
Parameters also allow being used in a reference. Right now it's only possible by using the string reference. Being able to use the `ref()` function makes it a bit more convenient.


markpeek:
Cool...thanks!

PR #906, opened 2017-11-15
Adds Tags to Cloudfront Distribution

taoistmath:
Allows Tags to be added to cloudfront distribution as outlined here:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution.html#cfn-cloudfront-distribution-tags


markpeek:
Thanks!

Issue #885, opened 2017-11-03
Update SecurityGroup resource


markpeek:
[AWS::EC2::SecurityGroup](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group.html)

In the [`EC2 Security Group Rule`](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group-rule.html) property type, use the `Description` property to specify the description of a security group rule.

markpeek:
Fixed via #899

PR #910, opened 2017-11-21
Add Description property for security group ingress and egress

bwhaley:
#885 added `Description` for `SecurityGroupRule` but not for `SecurityGroupIngress` nor `SecurityGroupEgress`.

### DIFF
--- a/troposphere/ec2.py
+++ b/troposphere/ec2.py
@@ -376,6 +376,7 @@ class SecurityGroupEgress(AWSObject):
     props = {
         'CidrIp': (basestring, False),
         'CidrIpv6': (basestring, False),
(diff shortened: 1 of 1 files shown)


markpeek:
Ahhh, good catch. Thanks!

statik:
Thank you! I'm very eager for this to be released, having descriptions our security group rules will be incredibly helpful.

bwhaley:
Note that there appears to be a limitation with the Description property which I've mentioned in this thread on the AWS forums: https://forums.aws.amazon.com/thread.jspa?threadID=267939

Basically the description doesn't work when the source is another security group. It only works when the source is a CIDR range. Hoping for some clarification from AWS on this.

PR #925, opened 2017-11-28
Updated for YAML template

praveensinghraghav:
This pull request will also inform users that we can create the AWS CloudFormation template in YAML also,
In order to do this we have to use print(template.to_yaml()) instead of print(template.to_json()).
Hence we can create YAML template from the troposphere.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -77,6 +77,27 @@ A simple example to create an instance would look like this:
     }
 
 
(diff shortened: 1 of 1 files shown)


markpeek:
Nice. Thanks!

praveensinghraghav:
My Pleasure

PR #928, opened 2017-11-30
Removed Logging info from SSM

Caprowni:
Hey,

Recently I found a bug when using Cloudformation and SSM whilst building an automated patching function.  The bug broke the MaintenanceWindowTask Class, this was caused by the LoggingInfo Class being deprecated by AWS - it has been confirmed by them in a support case I raised.  Whilst they worked to remove the API itself and amend the documentation I am updating the code here so that anyone who tries to use this in the foreseeable future won't run into these issues.

The workaround/proper method of enabling logging for Maintenance Window Tasks is to use "TaskInvocationParameters" inside the MaintWindowTask Class and then inside the TaskInvocationParameters, invoke the MaintenanceWindowRunCommandParameters Class to enabled bucket logs and a logging prefix.

### DIFF
--- a/troposphere/ssm.py
+++ b/troposphere/ssm.py
@@ -17,14 +17,6 @@ class NotificationConfig(AWSProperty):
     }
 
 
-class LoggingInfo(AWSProperty):
-    props = {
-        'Region': (basestring, True),
-        'S3Bucket': (s3_bucket_name, True),
-        'S3Prefix': (basestring, False),
-    }
-
-
 class MaintenanceWindowAutomationParameters(AWSProperty):
     props = {
         'DocumentVersion': (basestring, False),
@@ -162,7 +154,6 @@ class MaintenanceWindowTask(AWSObject):
 
     props = {
         'Description': (basestring, False),
-        'LoggingInfo': (LoggingInfo, False),
         'MaxConcurrency': (integer, False),
         'MaxErrors': (integer, True),
         'Name': (basestring, False),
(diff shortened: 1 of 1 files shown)


vsimakhin:
I've also opened a case with AWS Support and they confirmed it's a bug and they will fix it. Might be it's better just to update the fields as it's in boto api and wait for fixing it?

vsimakhin:
For instance, if you use the logging info, cloudformation will raise the error
```
Unrecognized field "S3Bucket" (class com.amazonaws.services.ssm.model.LoggingInfo), not marked as ignorable (3 known properties: "S3KeyPrefix", "S3Region", "S3BucketName"]) at [Source: N/A; line: -1, column: -1] (through reference chain: com.amazonaws.services.ssm.model.UpdateMaintenanceWindowTaskRequest["LoggingInfo"]->com.amazonaws.services.ssm.model.LoggingInfo["S3Bucket"])
```
If you try to change the class, you will get another error:
```
Property validation failure: [Encountered unsupported properties in {/LoggingInfo}: [S3BucketName]]
```

Caprowni:
Interesting, my interactions with AWS was that they mentioned the API's themselves for LoggingInfo will be deprecated in near future and therefore avoid using them altogether and just use the Task invocation parameters. 

I can indeed amend the LoggingInfo properties inside Troposphere to accept S3BucketName as a workaround just now then, if you would prefer?

Caprowni:
I updated the LoggingInfo Class as per your comments @vsimakhin

PR #929, opened 2017-11-30
Add Fargate support to ECS types

ejholmes:
This updates the ECS Service type to support the new properties for enabling VPC networking and Fargate support.

### DIFF
--- a/troposphere/ecs.py
+++ b/troposphere/ecs.py
@@ -2,6 +2,10 @@ from . import AWSObject, AWSProperty
 from .validators import boolean, integer, network_port, positive_integer
 
 
+LAUNCH_TYPE_EC2 = 'EC2'
+LAUNCH_TYPE_FARGATE = 'FARGATE'
+
+
 class Cluster(AWSObject):
     resource_type = "AWS::ECS::Cluster"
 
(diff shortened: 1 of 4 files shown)


shyamvala:
Thanks for the update @ejholmes. To complete Fargate support I think some changes needed to be done on the Task Definition side as well especially for supporting CPU and Memory fields. 

http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_TaskDefinition.html

ejholmes:
Good call! I've updated, and add an example for launching nginx using Fargate (tested in CloudFormation).

troposphere/ecs.py
```diff
@@ -197,9 +226,13 @@ class TaskDefinition(AWSObject):
 
     props = {
         'ContainerDefinitions': ([ContainerDefinition], True),
+        'Cpu': (basestring, False),
+        'ExecutionRoleArn': (basestring, False),
         'Family': (basestring, False),
+        'Memory': (basestring, False),
         'NetworkMode': (basestring, False),
         'PlacementConstraints': ([PlacementConstraint], False),
+        'RequiresCompatibilities': (list, False),
```
markpeek: Instead of "list" this should be "[basestring]" which will do checks on the elements as well.

ejholmes:
@markpeek fixed

markpeek:
Thanks @ejholmes for the Fargate changes. And appreciate the input from @shyamvala as well.

PR #931, opened 2017-12-01
Fix validation failure for fifo queues in the case that no QueueName …

slukes:
…is specified.  Note that Cloudformation *does* now automatically give a QueueName ending with .fifo in this case, contrary to the comment on PR #757.


markpeek:
Thanks!

slukes:
Thanks also to @PyvesB who paired with me on this :D

Issue #933, opened 2017-12-01
elasticloadbalancingv2 LoadBalancer validation bug.


johnctitus:
I encountered a glitch with the validation of the  LoadBalancer class for elasticloadbalancingv2.  The error occurred when attempting to allow for the creation of either aninternal or internet-facing NLB from the same template.  When trying to generate the template, I received an error instead:

`ValueError: LoadBalancer: only one of the following can be specified: SubnetMappings, Subnets`

The NLB definition causing the error is:

```
    NetworkLB = template.add_resource(elb.LoadBalancer(
        "NetworkLB",
        Name="NetworkLB",
        Scheme="internet-facing",
        Subnets=If(
            "isInternal", 
            [Ref(subnetA), Ref(subnetB)], 
            Ref(AWS_NO_VALUE)),
        SubnetMappings=If("isInternal", Ref(AWS_NO_VALUE), [
            elb.SubnetMapping(
                AllocationId=GetAtt(eipA, 'AllocationId'),
                SubnetId=Ref(subnetA)
            ),
            elb.SubnetMapping(
                AllocationId=GetAtt(eipB, 'AllocationId'),
                SubnetId=Ref(subnetB)
            )
        ]),
        Type='network'
    ))
```
Since this is a valid template, I propose we remove the [validation](https://github.com/cloudtools/troposphere/blob/master/troposphere/elasticloadbalancingv2.py#L143-L148) from this class, and leave it to the template creator to ensure they are not setting the properties incorrectly.  I will also get a PR submitted shortly.

markpeek:
@johnctitus I know the example above is a common case but leaving validation in does help for the other cases. I was thinking about it and perhaps something like this would allow for both instead of removing the validation altogether:
```
    def validate(self):
        conds = [
            'SubnetMappings',
            'Subnets',
        ]

        def check_if(names, props):
            l = []
            for name in names:
                l.append(name in props and isinstance(props[name], If))
            return all(l)

        if check_if(conds, self.properties):
            return

        exactly_one(self.__class__.__name__, self.properties, conds)
```
In other words, if both conditions exist and are If objects, ignore the additional validation. Otherwise do the exactly_one() check. Thoughts?

fawssupeng-cfntemplates:
Hey @markpeek,

That's sounds like a good solution.  I will work on updating the PR.

johnctitus:
Hey @markpeek,

That sounds like a good solution to me.  I will work on updating the PR.

PR #934, opened 2017-12-01
remove elasticloadbalancingv2 LoadBalancer Validation

johnctitus:
This change removes the validation step from the LoadBalancer class to allow for conditionally setting one of these properties, while setting the other to `AWS_NO_VALUE`

Fixes cloudtools/troposphere#933

### DIFF
--- a/troposphere/elasticloadbalancingv2.py
+++ b/troposphere/elasticloadbalancingv2.py
@@ -3,7 +3,7 @@
 #
 # See LICENSE file for full license.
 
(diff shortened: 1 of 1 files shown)


johnctitus:
PR now change as follows:

elasticloadbalancingv2 LoadBalancer validation now checks if all mutually exclusive conditions are instances of the `If` class, and if so returns.  Then proceeds to verifying only a single property is configured.

markpeek:
Thanks for the fix and reworking the PR.

PR #941, opened 2017-12-07
Added support for AutoPublishAlias to AWS::Serverless::Function

floodfx:
Hi there -

Thanks for troposphere!

I was trying to use it to build a Serverless Application Model (SAM) definition and discovered that the `AutoPublishAlias` attribute was not supported.  
Based on the [SAM spec](https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md), `AutoPublishAlias` should be allowed as an optional attribute.

I've added support for it and a test.  LMK if all looks right as this is my first PR here.

Best,
-Donnie

### DIFF
--- a/tests/test_serverless.py
+++ b/tests/test_serverless.py
@@ -61,6 +61,18 @@ class TestServerless(unittest.TestCase):
         t.add_resource(serverless_func)
         t.to_json()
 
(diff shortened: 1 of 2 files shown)


markpeek:
Nice job...thanks!

Issue #944, opened 2017-12-11
guardduty AWSObjects missing resource_type variables


rtkjbillo:
I'm attempting to generate a template with troposphere 2.1.2 to add GuardDuty, and it appears that the `Detector`, `IPSet` and `ThreatIntelSet` objects do not define the `resource_type` variable in `guardduty.py`. 

As a result, the JSON that gets output does not include a `Type` section, and fails to validate when being uploaded to CloudFormation.

Example code:

```
>>> import troposphere.guardduty
>>> from troposphere import Template
>>> t = Template()
>>> g = troposphere.guardduty.Detector('example', Enable=True)
>>> t.add_resource(g)
<troposphere.guardduty.Detector object at 0x1012701d0>
>>> print t.to_json()
{
    "Resources": {
        "example": {
            "Properties": {
                "Enable": "true"
            }
        }
    }
}
```

I will attempt to submit a pull request shortly unless this is already on someone's radar.

PR #945, opened 2017-12-11
Add resource_type value and unit tests for guardduty `AWSObject`s

rtkjbillo:
Unit tests confirm construction of these objects and ensure that the
`Type` field is set properly in the output dict to prevent regressions.

See: https://github.com/cloudtools/troposphere/issues/944


markpeek:
Nice catch! Thanks!

PR #946, opened 2017-12-11
Recursively validate JSON while preserving helper function types

thejohnlin:
Currently, using a Ref inside the Schema parameter to create a new apigateway.Model resource will result in an Exception from attempting to validate the dictionary via json.dumps(). This PR is an attempt to enable this use-case by only attempting json.dumps() on the values of type basestring in the dictionary.


thejohnlin:
Additional tests will be coming soon.

Issue #871, opened 2017-10-25
awslambda.Function does not support Cloudformation Package CLI command


marty-sullivan:
As the title suggests, troposphere.awslambda.Function resource does not support a String argument for the Code parameter for supplying a local path. This is for the `aws cloudformation package` command in the CLI. It is likely that other resources are also missing this support as outlined in the CLI documentation below:

http://docs.aws.amazon.com/cli/latest/reference/cloudformation/package.html

The error thrown is:

TypeError: <class 'troposphere.awslambda.Function'>: MyFunction.Code is <type 'str'>, expected <class 'troposphere.awslambda.Code'>

marty-sullivan:
Not totally familiar with the inner workings of Troposphere and AWSObject, so not sure if it is possible to allow two parameter types in a resource. However, in my own code as a temp workaround, I created my own PackageFunction Resource that looks like the following. This seems to work for now, but would be nice if Troposphere added this support.

```
class PackageFunction(AWSObject):
    resource_type = "AWS::Lambda::Function"

    props = {
        'Code': (basestring, True),
        'Description': (basestring, False),
        'DeadLetterConfig': (DeadLetterConfig, False),
        'Environment': (Environment, False),
        'FunctionName': (basestring, False),
        'Handler': (basestring, True),
        'KmsKeyArn': (basestring, False),
        'MemorySize': (validate_memory_size, False),
        'Role': (basestring, True),
        'Runtime': (basestring, True),
        'Tags': (Tags, False),
        'Timeout': (positive_integer, False),
        'TracingConfig': (TracingConfig, False),
        'VpcConfig': (VPCConfig, False),
    }
```

markpeek:
I don't believe that will do what you want it to do. The package command appears to take your code, uploads it to S3, and then rewrites your template to use that uploaded code. To do this you would need to use something like boto3 to upload your code and then use the Code property to specify the S3 location.

marty-sullivan:
Yes, I am running `aws cloudformation package`  externally to Troposphere after generating the template json.

markpeek:
I need to think about how best to handle this issue as adding basestring would reduce the verification of Code. For example, the fix would look like this (allowing Code and basestring):
```
'Code': ((Code, basestring), True),
```
Note: you could easily override this in your code by adding a workaround like this:
```
from troposphere.awslambda import Function

# Allow Code to also be a basestring for "aws cloudformation package"
Function.props['Code'] = ((Code, basestring), True)
```

[...]

PR #947, opened 2017-12-11
make Serverless::Function CodeUri optional

troyready:
CodeUri is required in a final CFN template, but should be omitted in
templates that will be processed with `sam package` (it will upload the
function to S3 and create a final template with a generated CodeUri)

Example at
http://docs.aws.amazon.com/lambda/latest/dg/with-api-example-use-app-spec.html#api-tutorial-spec

~Another approach might be to make another class for this?~ EDIT: Alternate approach proposed in #952

### DIFF
--- a/troposphere/serverless.py
+++ b/troposphere/serverless.py
@@ -58,7 +58,7 @@ class Function(AWSObject):
     props = {
         'Handler': (basestring, True),
         'Runtime': (basestring, True),
-        'CodeUri': ((S3Location, basestring), True),
+        'CodeUri': ((S3Location, basestring), False),
         'FunctionName': (basestring, False),
         'Description': (basestring, False),
         'MemorySize': (validate_memory_size, False),


markpeek:
This came up in another issue and I offered a workaround. Please take a look at #871. I'm on the fence about disabling the validation. Be good to hear your thoughts.

troyready:
Maybe it should be a new class? Something like
```
diff --git a/troposphere/serverless.py b/troposphere/serverless.py
index e60f53c..cf2a25f 100644
--- a/troposphere/serverless.py
+++ b/troposphere/serverless.py
@@ -76,6 +76,12 @@ class Function(AWSObject):
     }
 
 
+class FunctionForPackaging(Function):
+
+    props = Function.props
+    props['CodeUri'] = ((S3Location, basestring), False)
+
+
 class Api(AWSObject):
     resource_type = "AWS::Serverless::Api"
 

```

It's not hard to override the properties locally, but it "feels" like there should be a built-in way to make this type of template.

markpeek:
@troyready can you resubmit your PR with your new class?

markpeek:
Merged in #952 instead. Closing.

PR #952, opened 2017-12-13
add FunctionForPackaging class to serverless

troyready:
CodeUri is required in a final CFN template, but should be omitted in
templates that will be processed with sam package (it will upload the
function to S3 and create a final template with a generated CodeUri)

Example at
http://docs.aws.amazon.com/lambda/latest/dg/with-api-example-use-app-spec.html#api-tutorial-spec

This class will allow Functions to be specified that can omit CodeUri

(This is an alternate solution to #947 )


markpeek:
I needed to tweak the fix so the underlying props dict didn't change for Function. I tried to push through your branch but failed due to conflicts. I ended up cherry-picking your change and merging in my change plus a test. Please take a look and verify you agree with it. Thanks for the change.

troyready:
Perfect, thanks!

PR #957, opened 2018-01-02
Add dictionary param support to resources that have AWSHelperFn props.

LyndonHook:
Adds support to resources that use AWSHelperFn as properties so they can be instantiated by using a python dictionary type instead of having to pass the instantiated helper function. This is necessary if your resource creation is driven by a generic YAML config setup.
e.g.
`YAML -> Dictionary -> Troposphere Resource`


LyndonHook:
Any feedback on this? I'd love to get it merged.

PR #960, opened 2018-01-07
Adding support for Block Device Mapping V2

eirriel:
Hi,
I found that Server object doesn't have this property so maybe it's a good idead to add it.


markpeek:
Thanks!

PR #961, opened 2018-01-08
Add support for Policy Document in SAM template.

davemasino:
Changed policy_validator function to allow for an IAM Policy Document value for Policies attribute.


markpeek:
Perhaps this should use the awacs Policy as well. Take a look at the sqs.py [PolicyDocument](https://github.com/cloudtools/troposphere/blob/master/troposphere/sqs.py#L53).

davemasino:
Thanks, will do. I noticed in the docs that awacs.aws.Policy will be deprecated in the future. Should I use awacs.aws.PolicyDocument instead?

markpeek:
Excellent. Thanks!

Issue #650, opened 2017-01-21
API documentation


boompig:
Is there any documentation outside the source code? [Read the docs](https://readthedocs.org/projects/troposphere/) doesn't have any useful information.

phobologic:
Largely there isn't a ton of documentation since we try to keep troposphere a one to one copy of the cloudformation templating language, so you can usually just use that to describe what fields/attributes various resources/properties have.

What kind of documentation would you like to see though?

boompig:
Auto-generated documentation of methods and classes would be nice.

phobologic:
Yeah - the classes themselves are pretty bare of docs, unfortunately.  It'd just be signatures.

*that said* I know @markpeek is working occasionally on integrating the Cloudformation schema that comes directly from AWS, and that has links to the actual cloudformation docs for each resource, so maybe we could do something cool with that at some point.

Anyway, thanks for the clarification.

ScriptAutomate:
I recently was thinking about this problem with Troposphere, and saw that this issue was opened a few years ago. It's still an existing problem, and I though I'd add some thoughts here in case others wanted to contribute.

Some initial Sphinx documentation has been slowly added since this issue discussion ended, with #1311 and #1314, but Classes are pretty much just as-is (source code).

I would love to help with this, actually, but may need some help. I'm thinking something like the following could be used to automate and expand the documentation with Sphinx and the CFN resource spec documentation links:

### To Start

Some priority items:

- Travis CI should be re-compiling via Sphinx with [Troposphere readthedocs](https://troposphere.readthedocs.io/en/latest/) integration always using the latest compiled release version. It's currently only reflecting when Sphinx was initially added to this repo back with `v2.4.1`
  - This would need to also involve `release` in `./docs/conf.py` updating to the current version each time a new Troposphere release occurs
- Basic directions/documentation added for how to run Sphinx, and contribute to the documentation (compiling with Python 2.7, what make commands are used like `make html`? etc.)

### Furthering Automated API Documentation Ideas

- See if there is a way to run Sphinx to compile without needing to use Python 2.7.x to do so, as currently it fails due to hangups around `basestring` when it runs in Python 3.6+. With 2.7 being EOL, would be good to get Sphinx compiling in Python3
- Add `sphinx.ext.napolean` to the `extensions` list in `./docs/conf.py` so that documentation can be done with Google Style Docstrings, for better readability
[...]

PR #963, opened 2018-01-10
Stab at documenting Troposphere basics

bgdnlp:
It seems like it's not very clear at first sight how Troposphere should be used, people are looking for docs (ex. #650). I did too. This is an attempt to maybe get people starting. It's a README.md file in 'docs' directory, which means that http://troposphere.readthedocs.io/en/latest/ should pick it up, if I understand correctly.

Do you think it would be useful?


markpeek:
Thanks! Always great to get more docs.

PR #966, opened 2018-01-11
Adding HealthCheckGracePeriodSeconds into ECS Service

andydoTech:
https://aws.amazon.com/about-aws/whats-new/2017/12/amazon-ecs-adds-elb-health-check-grace-period/

### DIFF
--- a/troposphere/ecs.py
+++ b/troposphere/ecs.py
@@ -98,6 +98,7 @@ class Service(AWSObject):
         'PlatformVersion': (basestring, False),
         'ServiceName': (basestring, False),
         'TaskDefinition': (basestring, True),
+        'HealthCheckGracePeriodSeconds': (positive_integer, False),
     }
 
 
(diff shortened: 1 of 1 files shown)


phobologic:
I don't see this mentioned in the Cloudformation docs yet:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-service.html

Are you sure this is accepted by Cloudformation?

phobologic:
It's also not currently listed in the resource schema specification:

https://d1uauaxba7bl26.cloudfront.net/latest/gzip/CloudFormationResourceSpecification.json

andydoTech:
Ah, you are right. Let me close and submit it when they are ready.

phobologic:
Looks like this was added, and we have another PR for it: #971 :)

phobologic:
Re-opening since this should be good now, thanks @andydoTech !

phobologic:
Ok, made a minor ordering change, but I'm going to accept this now.  Thanks @andydoTech !

PR #968, opened 2018-01-18
Add AllowedPattern to Parameter

khornberg:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-parameter.html#cfn-ssm-parameter-allowedpattern


markpeek:
Thanks!

PR #971, opened 2018-01-24
Add support for HealthCheckGracePeriodSeconds to ECS::Service

acmcelwee:
As of Jan 23, the `ECS::Service` resource has support for the new `HealthCheckGracePeriodSeconds` config.

[Docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-service.html#cfn-ecs-service-healthcheckgraceperiodseconds)

### DIFF
--- a/troposphere/ecs.py
+++ b/troposphere/ecs.py
@@ -89,6 +89,7 @@ class Service(AWSObject):
         'Cluster': (basestring, False),
         'DeploymentConfiguration': (DeploymentConfiguration, False),
         'DesiredCount': (positive_integer, False),
+        'HealthCheckGracePeriodSeconds': (integer, False),
         'LaunchType': (launch_type_validator, False),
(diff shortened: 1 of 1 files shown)


troposphere/ecs.py
```diff
@@ -89,6 +89,7 @@ class Service(AWSObject):
         'Cluster': (basestring, False),
         'DeploymentConfiguration': (DeploymentConfiguration, False),
         'DesiredCount': (positive_integer, False),
+        'HealthCheckGracePeriodSeconds': (integer, False),
```
phobologic: Should the validator here be `positive_integer`?

https://github.com/cloudtools/troposphere/blob/master/troposphere/validators.py#L27

phobologic:
Actually, I'm going to go ahead and accept #966 since it should do what we want here.  Thanks @acmcelwee !

PR #972, opened 2018-01-25
Add long form parameter to to_yaml

fredd369:
currently, troposphere to_yaml function will by default use the short form.  This causes tag issue with yaml parsers.  cfn_flip already has the option to turn on the long form, so this should be passed through so that it can be used in troposphere.


markpeek:
Thanks!

PR #995, opened 2018-03-05
"JobDefinitionName" property in JobDefinition class is not required

SHLo:
According to cloudformation document https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-batch-jobdefinition.html#cfn-batch-jobdefinition-jobdefinitionname


markpeek:
Thanks!

PR #998, opened 2018-03-07
Add support for Aurora 5.7

RasmusWernerLarsen:
- This adds aurora-mysql, which is the appropriate identifier for Aurora
5.7, as a valid engine type for rds instances and clusters.

See: https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBCluster.html under "Engine".


markpeek:
Thanks  for the PR. This was already added via https://github.com/cloudtools/troposphere/commit/0226a0f478c054347ed54171eee4373e7cf2eadc

Issue #999, opened 2018-03-09
Tags support for SSM documents


Stvad:
SSM documents now support tags: https://docs-aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-document.html#cfn-ssm-document-tags 
It'd be great to have that support implemented in troposphere

markpeek:
Fixed via #1000

PR #1000, opened 2018-03-09
Tags support for SSM documents #999

Stvad:
SSM Documents now support tags (#999)

Tested by adding tag information to some documents I've generated via Troposphere - when I updated the stack with new template - tags were successfully added to the documents.


markpeek:
Thanks!

PR #1003, opened 2018-03-12
Update to cfn-flip 1.0.2

stilvoid:
cfn-flip 1.0.2 fixes the issue where zero-prefixed strings are not quoted which confuses the CloudFormation service


phobologic:
Thanks @stilvoid

PR #1005, opened 2018-03-14
Corrected the description of NatGateway

atambol:
Nat gateway is a different resource from a Nat instance. I have updated the description to reflect the same as per my understanding.


phobologic:
Thanks @atambol  - I fixed the tests that were breaking due to the change, I'm going to go ahead and merge this.

PR #1007, opened 2018-03-15
Update deprecated modules in TemplateGenerator

noelmccrory:
When using the TemplateGenerator, the following warning message appears:

UserWarning: This module has replaced by troposphere.dynamodb. Please import that module instead, as troposphere.dynamodb2 will be removed soon.


phobologic:
Thanks @noelmccrory !

PR #1010, opened 2018-03-18
Correct Spot Fleet TagSpecifications

calebmarcus:
The new support for adding tags to Spot Fleets created in CloudFormation [works a bit differently](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-spotfleet-spotfleetrequestconfigdata-launchspecifications.html#cfn-ec2-spotfleet-spotfleetlaunchspecification-tagspecifications) than was added in c4ee115e -- the property is TagSpecifications (vs. TagSpecification) and it must be a list.

The SpotFleetTagSpecification property ResourceType [claims it's not required](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-spotfleet-spotfleetrequestconfigdata-launchspecifications-tagspecifications.html#cfn-ec2-spotfleet-spotfleettagspecification-resourcetype) but I get "Tag specification resource type must have a value" in CloudFormation when I omit it, so I marked it required. (There is only one valid value, so it's a bit silly, but oh well.)

### DIFF
--- a/troposphere/ec2.py
+++ b/troposphere/ec2.py
@@ -706,7 +706,7 @@ class LaunchSpecifications(AWSProperty):
         'SecurityGroups': ([SecurityGroups], False),
         'SpotPrice': (basestring, False),
         'SubnetId': (basestring, False),
-        'TagSpecification': (SpotFleetTagSpecification, False),
+        'TagSpecifications': ([SpotFleetTagSpecification], False),
         'UserData': (basestring, False),
         'WeightedCapacity': (positive_integer, False),
     }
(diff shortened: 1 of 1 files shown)


markpeek:
Awesome. Thanks for the PR!

PeteE:
@markpeek any idea when the next troposphere release will happen?  I'd like to have this change included so I don't have to fetch the egg from github.  Thanks for fixing @calebmarcus!

PR #1011, opened 2018-03-19
Allow properties supported by `cloudformation package` to be strings

benbridts:
When uploading local artifacts using the `cloudformation package`
command, you can specify paths relative to the template instead
of the normal properties.

See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-cli-package.html


ikben:
The tests fail, because now the TemplateGenerator can't be sure how to map those properties from json. I'm not sure how hard it is to fix that.

PR #1012, opened 2018-03-25
Simplify using parameter groups and labels

azatoth:
Basing of the example in #367 I've made an implementation that will
allow you to easy call `grp_template = template.add_group('Group Name')`
and subsequently work on that `grp_template` for that group of items
like it was the top level template


simplesteph:
that would be amazing to get this merged! Is it still WIP and how do we get this in?

PR #1015, opened 2018-03-28
Adding SUSE to list of operating systems

kwkeefer:
SUSE is a supported operating system for SSM. 

See BOTO3 docs:
http://boto3.readthedocs.io/en/latest/reference/services/ssm.html#SSM.Client.create_patch_baseline


markpeek:
Thanks!

PR #1016, opened 2018-04-02
Eliminate infinite loop when pickle loads BaseAWSObject and objects derived from it.

satvidh:
If pickle loads this object, then __getattr__ will cause an infinite loop when pickle invokes this object to look for __setstate__ before attributes is "loaded" into this object. Therefore, short circuit the rest of this call if attributes is not loaded yet. See http://grokbase.com/t/python/python-bugs-list/092sczhjwk/issue5370-unpickling-vs-getattr for source of this change.

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -130,6 +130,14 @@ class BaseAWSObject(object):
             self.template.add_resource(self)
 
     def __getattr__(self, name):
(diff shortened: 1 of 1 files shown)


phobologic:
Hey @satvidh - thanks for the PR.  I've gone ahead and added a test for the issue this has fixed, and I'm going to merge this.  Thanks again, and sorry for the delay!

PR #1017, opened 2018-04-03
Update Codebuild Sources to include Github Enterprise

himalayanelixir:
Added support for Github Enterprise as a source for Codebuild. Support for 'GitCloneDepth' and 'InsecureSsl' were added in the last release but the source types were not updated.


phobologic:
Thanks @himalayanelixir !

Issue #1027, opened 2018-04-20
codepipeline.ActionTypeId has the wrong case


chizou:
https://github.com/cloudtools/troposphere/blob/master/troposphere/codepipeline.py#L10

As referenced from the CloudFormation documentation, the D in ID should be lowercase:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codepipeline-pipeline-stages-actions.html#cfn-codepipeline-pipeline-stages-actions-actiontypeid

markpeek:
Fixed via rename.

PR #1028, opened 2018-04-20
Fix misspelling of codepipeline.ActionTypeId to match CF documentation

chizou:
This fixes the issue reported in Issue #1027 while maintaining backwards compatibility for people that used the old spelling.

### DIFF
--- a/troposphere/codepipeline.py
+++ b/troposphere/codepipeline.py
@@ -16,6 +16,10 @@ class ActionTypeID(AWSProperty):
     }
 
 
+class ActionTypeId(ActionTypeID):
+    pass
+
+
 class ArtifactDetails(AWSProperty):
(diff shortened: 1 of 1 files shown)


markpeek:
@phobologic I'm inclined to just rename ActionTypeID to ActionTypeId instead of carrying both forward (even though it would be a breaking change). Thoughts?

phobologic:
I'm honestly fine either way - breaking changes like this don't tend to cause issues, other than folks having to update a few templates.  If we were really concerned about it we could throw a warning whenever folks used `ActionTypeID` for a release, then remove it.  Probably not necessary though.

markpeek:
This has been fixed by renaming instead of an extra class. Thanks for the issue and PR.

PR #1035, opened 2018-05-02
add "ServiceRegistries" attribute in AWS::ECS::Service

SHLo:
this attribute is for Service Discovery setting which is currently working for cloudformation
https://docs.aws.amazon.com/AmazonECS/latest/developerguide/service-discovery.html

### DIFF
--- a/troposphere/ecs.py
+++ b/troposphere/ecs.py
@@ -99,6 +99,7 @@ class Service(AWSObject):
         'PlatformVersion': (basestring, False),
         'ServiceName': (basestring, False),
         'TaskDefinition': (basestring, True),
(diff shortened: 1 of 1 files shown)


markpeek:
According to the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-service.html) this is not yet exposed in CloudFormation. We'll hold off merging for now. Also, likely would be good to have the properties alphabetized.

markpeek:
This was added recently but the implementation needed a ServiceRegistry property. See this commit for details: 2b2b70dd00d5192fcc7a0f1833cada72359c7e7a

PR #1040, opened 2018-05-09
Updated appsync apikey expires to be an int: Fixes #1039

### LINKED ISSUE #1039: Appsync ApiKey Expires needs to be an int
The expires prop on the appsync ApiKey class needs to be an int otherwise the decimal place causes cloudformation to throw a `Can not deserialize value of type java.lang.Long from String` error.

jimox:
Fixes issue https://github.com/cloudtools/troposphere/issues/1039

The expires prop on the appsync ApiKey class needs to be an int otherwise the decimal place causes cloudformation to throw a `Can not deserialize value of type java.lang.Long from String` error.

### DIFF
--- a/troposphere/appsync.py
+++ b/troposphere/appsync.py
@@ -4,7 +4,7 @@
 # See LICENSE file for full license.
 
 from . import AWSObject, AWSProperty
-from .validators import boolean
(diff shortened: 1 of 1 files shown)


phobologic:
Thanks!

PR #1041, opened 2018-05-09
Fix mappings

lrascao:
Multiple `template.add_mapping` calls were squashing the previous key, this patch fixes that

### DIFF
--- a/tests/test_mappings.py
+++ b/tests/test_mappings.py
@@ -1,0 +1,47 @@
+import unittest
+
+from troposphere import Template
+
+single_mapping = """\
+{
+    "Mappings": {
+        "map": {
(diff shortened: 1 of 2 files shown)


phobologic:
Hey @lrascao thanks for the PR, the code looks good - I'm a little worried that this is a change in behavior which might be used by other users of troposphere (the code for add_mapping is quite old, so there's a decent chance that someone is doing so).  Also the term `add_mapping` might not necessarily mean `update existing mapping` at first.  @markpeek any thoughts?

lrascao:
ping?

lrascao:
bump

markpeek:
Thanks! Sorry for the delay.

Issue #1045, opened 2018-05-16
Cloudformation error on tags type validation when provisioning dax resource


jswoods:
When provisioning a dax resource, cloudformation shows the error:

```
Property validation failure: [Value of property {/Tags} does not match type {Map}]
```

The dax object should use a dict instead of the Tags object for the Tags property.

ARMAAN7139:
I am getting similar error when provisioning AWS PinPoint App using CFN.

markpeek:
@ARMAAN7139 since this issue is closed, can you open a new issue and include a small test case with the error you are getting?

PR #1046, opened 2018-05-16
Use a dict instead of the Tags object for the Tags prop for Dax

jswoods:
Cloudformation expects a dict rather than a list of key value pairs which the Tags object provides. See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dax-cluster.html.

This fixes https://github.com/cloudtools/troposphere/issues/1045


markpeek:
@cmmeyer would you happen to know why dax appears to be using a totally different Tags format? This sounds like more of a CloudFormation defect to me.

jswoods:
@markpeek there appears to be two ways to handle tags in cloudformation. If you look at https://github.com/cloudtools/troposphere/blob/master/troposphere/batch.py#L19 for example, it uses a dict. I'm sure there are others as well.

cmmeyer:
This may be an artifact of the new self-service model for CloudFormation support.  When in doubt, the resource specification is your best friend for confirming/validating types:

[https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html
)
```
    "AWS::DAX::Cluster": {
...
        "Tags": {
          "Required": false,
          "Documentation": "http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dax-cluster.html#cfn-dax-cluster-tags",
          "PrimitiveType": "Json",
          "UpdateType": "Mutable"
        }
```

Doing some quick sampling, there's definitely inconsistency here -- most are "List", a couple are "JSON" and (interestingly) some are "Tag".  

Lemme check in with the team and see if I can understand why some are difference, might just be something we can call out in on-boarding.

jswoods:
@cmmeyer - for my own clarity, are you saying this fix is valid or is not valid? If not, what would be the proposed solution to https://github.com/cloudtools/troposphere/issues/1045?

cmmeyer:
If this fixes tags for DAX then it should be the right call.  I don't think you're going to see anything in the API/CloudFormation to normalize tags anytime soon.

adamdavis40208:
Can this be merged? Tagging DAX resources would be a nice-to-have

r631269:
+1, could really use this soon

adamdavis40208:
@markpeek This PR has been around for ages, and it'd be nice to have it pulled in. Is there a better spot to discuss a merging this? (google group/IRC/etc)

Thanks!

adamdavis40208:
Hello 2019. Any updates on this?

markpeek:
Thanks. Sorry for the delay and thanks for the constant reminders. I was hoping the CloudFormation team would "fix" this incompatibility but it is better to move forward with what works.

adamdavis40208:
Awesome! Now of course amazon will fix it, which will cause another bug :)

That's how the universe works

cmmeyer:
Sorry, folks! The distributed nature of the services means it's almost impossible to normalize this sort of thing across the platform.  I'll keep this bookmarked and try to give you a heads up if I hear anything.

PR #1048, opened 2018-05-22
Allow Ref's to test equality against their data

phobologic:
This is largely used by the Pseudo Parameters, so that these three
properties are all the same:

```
from troposphere import NoValue, Ref

s = "AWS::NoValue"
r = Ref(s)

s == r
s == NoValue
r == NoValue
```

This is in preparation for another PR to fix the validators for
`mutually_exlusive` and `exactly_one` - which should both allow
multiple values provided the values are set to `AWS::NoValue`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 sudo: false
 python:
   - "2.7"
-  - "3.3"
   - "3.4"
   - "3.5"
   - "3.6"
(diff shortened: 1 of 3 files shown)


happyraul:
Refs used to be hashable before this change. Now that a `__eq__` method has been defined, they are no longer hashable. Should a `__hash__` also be added to revert to the previous behavior?

PR #1050, opened 2018-05-23
Allow multiple NoValue properties in mutually_exclusive

phobologic:
mutually_exclusive ignores the value of properties, only looking at
whether or not they are defined. This causes a bug where you cannot
have two mutually exclusive properties, even if the value of those
properties is AWS::NoValue, which is allowed by cloudformation.


phobologic:
Not sure what is going on with the tests - when I add `from . import NoValue` to `troposphere/validators.py` it causes random tests to break.

PR #1053, opened 2018-05-25
Allow Refs to be hashable using their data

happyraul:
With the addition of the `__eq__` method (#1048), Refs are no longer hashable. This PR should revert to the previous behavior.


markpeek:
Nice! Thanks!

phobologic:
Thanks for fixing this error I introduced, @happyraul !

PR #1054, opened 2018-05-26
Handle bool types in the template generator

### LINKED ISSUE #1044: TemplateGenerator emits TypeError when importing bools that should be strings
Gist: https://gist.github.com/ThatGerber/1fdb34ca4d293f8c34c1fbca09c5252f

This same error affects other resources like `troposphere.ecs.Environment` https://github.com/cloudtools/troposphere/blob/master/troposphere/ecs.py#L105 where the value being imported can be any type, but should be cast as String. The Json/Yaml marshaller attempts to determine the type of each object when creating the object, and the TemplateGenerator requires them to be passed in a certain way.

markpeek:
Fixes #1044

### DIFF
--- a/troposphere/template_generator.py
+++ b/troposphere/template_generator.py
@@ -263,6 +263,14 @@ class TemplateGenerator(Template):
                         expected_type in self.inspect_members):
                     kwargs[prop_name] = self._create_instance(
                         expected_type, kwargs[prop_name], prop_name)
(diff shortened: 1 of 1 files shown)


markpeek:
Fixed via ad7c5f2a562e865e89d346dbc349ec4ff69411f9 and db117248dfb0fc500ae9d10db34c42608240bb8d

PR #1055, opened 2018-05-28
Fix to_yaml long_form parameter

Barker404:
Since the update to use cfn_flip 1.0.2, the `long_form` parameter of `to_yaml` was being incorrectly passed to the `clean_up` parameter of `cfn_flip.to_yaml`. This fixes that, and also add `clean_up` as an actual parameter.


markpeek:
Thanks for the fix! Sorry for the regression, I updated the tests (7c9bee845153babb0330960e443590144ab6b10c) to catch any future changes.

PR #1063, opened 2018-06-06
Removing validation for parameters that don't require them

JulianGindi:
I encountered this when attempting to use the "PrimaryClusterId" property of the Elasticache ReplicationGroup type. It is validating on fields that are actually not allowed to be specified when PrimaryClusterId is specified. [Here is a link to the relevent documentation.](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticache-replicationgroup.html#cfn-elasticache-replicationgroup-primaryclusterid)

The change below generates a valid json template which accomplishes my original goal of creating a ReplicationGroup from an existing cache instance.


JulianGindi:
Hey there, wondering if we can merge this?

markpeek:
Thanks for the reminder. My main issue was removing validation. I opened #1328 to track adding a class validator for this resource.

PR #1068, opened 2018-06-12
fix casing of CpuCredits

PeteE:
This fixes the following error:
```
Property validation failure: [Encountered unsupported properties in {/LaunchTemplateData/CreditSpecification}: [CPUCredits]]
```


markpeek:
Thanks!

PR #1070, opened 2018-06-14
Add 'Name' property for AWS::Serverless::Api

Iamz:
Add the missing 'Name' property for AWS::Serverless::Api resource type. (See the document [here](https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessapi))


markpeek:
Thanks!

PR #1072, opened 2018-06-15
Add equality methods to Template

saysjonathan:
This change adds support for `__eq__()`, `__ne__()`, and `__hash__()` to `Template` objects. `to_json()` is used to test for equivalency and to generate the hash.

The following will be available to `Templates`:

- equality testing
```
>>> t1 = Template(Description='foo')
>>> t2 = Template(Description='foo')
>>> t1 == t2
True
```

- inequality testing
```
>>> t1 = Template(Description='foo')
>>> t2 = Template(Description='bar')
>>> t1 == t2
False
```

- usage in sets
```
>>> t1 = Template(Description='foo')
>>> t2 = Template(Description='foo')
>>> len(set([t1, t2]))
1
>>> t1 = Template(Description='foo')
>>> t2 = Template(Description='bar')
>>> len(set([t1, t2]))
2
```

### DIFF
--- a/tests/test_template.py
+++ b/tests/test_template.py
@@ -66,5 +66,49 @@ class TestValidate(unittest.TestCase):
             template.add_mapping("mapping", {"n": "v"})
 
 
+class TestEquality(unittest.TestCase):
+    def test_eq(self):
+        metadata = 'foo'
+        description = 'bar'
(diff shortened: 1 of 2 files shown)


markpeek:
Nice...thanks!

PR #1075, opened 2018-06-21
Fix validator for SSM MaintenanceWindowTask TaskType

haylix-edward:
It appears the documentation (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-maintenancewindowtask.html#cfn-ssm-maintenancewindowtask-tasktype) is not correct, and that SSM actually requires the string STEP_FUNCTIONS.

e.g.
image

### DIFF
--- a/troposphere/validators.py
+++ b/troposphere/validators.py
@@ -266,7 +266,7 @@ def notification_event(events):
 
 
 def task_type(task):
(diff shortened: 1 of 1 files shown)


haylix-edward:
I pushed a second unrelated commit to master, which caused the Travis test to fail. Will open a new PR.

PR #1079, opened 2018-06-23
Generate class stubs necessary for autocompletion

DrLuke:
Troposphere dynamically generates the property-fields for its classes. IDEs thus have no way of knowing these fields beforehand and can't autocomplete them. This can be remedied by providing stubs that outline the fields and their types of all classes. See [PEP 484#stub-files](https://www.python.org/dev/peps/pep-0484/#stub-files) for more information.

This Pullrequest adds scripts to generate stub-files for python2 and python3, which can then be shipped alongside the python sources.


DrLuke:
Run this script exactly as you would run gen.py. The output code should be put in *.pyi files alongside the *.py files with the same name.

DrLuke:
I have further improved stub generation by also creating a stubs for __init__. The python2 stubs aren't quite as elaborate as the python3 stubs, which additionally contain the types for each argument.

DrLuke:
Python 2 stub generator was removed because autocompletors are to honor python 3 stubs for both versions of python.

Python 3 stub generator was merges into gen.py, which will generate regular code when called as before, and will generate stub code when the flag `--stub` is set.

markpeek:
How are you running this to get each individual file output? I'll likely accept this change and then fixup the script a little bit to output the files aligned with the troposphere file naming. Also, I did see a KeyError that needs to be tracked down which occurred when running it on spec/CloudFormationResourceSpecification.json. Appreciate you making the changes.

markpeek:
Thanks!

PR #1088, opened 2018-06-27
Allow github enterprise to use OAuth in codebuild

trillchewy:


phobologic:
Hey @trillchewy - thanks for the PR.  I went digging into the docs, and they don't actually mention anything about Auth being allowed for `GITHUB_ENTERPRISE` but they do mention `BITBUCKET` so I'm a little concerned this won't be valid.  Have you spoken with AWS support about this/tried this yourself?

https://docs.aws.amazon.com/codebuild/latest/APIReference/API_ProjectSource.html

trillchewy:
I have tried it and it deploys as valid through cloudformation, however you may be correct as that is in not needed as it deploys as well without the auth set.  I will close this out for now.  Thanks!

PR #1095, opened 2018-06-29
Added Fn:Base64 function

### LINKED ISSUE #1094: Add support for Fn::Base64
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-base64.html

ror6ax:
Fixes #1094

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -4,6 +4,7 @@
 # See LICENSE file for full license.
 
 
+import base64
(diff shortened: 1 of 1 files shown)


markpeek:
This already exists [here](https://github.com/cloudtools/troposphere/blob/master/troposphere/__init__.py#L407). Note: since the function itself returns a base64 blob, the passed in value should not be encoded.

PR #1098, opened 2018-07-04
Add basestring to allowed types for Code property of Function class

haylix-edward:
In order to use the "aws package" command, we need to be able to provide a file/folder path to the Function resource as a simple string value.

### DIFF
--- a/troposphere/awslambda.py
+++ b/troposphere/awslambda.py
@@ -90,6 +90,9 @@ class Code(AWSProperty):
             )
 
 
+codetypes = Code, basestring
+
+
 class VPCConfig(AWSProperty):
 
     props = {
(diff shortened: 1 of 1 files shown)


JordonPhillips:
@haylix-edward I believe what you're looking for is [`aws.serverless.Function`](https://github.com/cloudtools/troposphere/blob/master/troposphere/serverless.py#L50) ([aws docs](https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction)).

haylix-edward:
@JordonPhillips We did look at using those, but they don't work inside nested cloudformation stacks (it has been long enough that I don't recall exactly what the issue was).

PR #1100, opened 2018-07-06
Fix CPUCredits casing and create LaunchTemplateCreditSpecification class

### LINKED ISSUE #1099: Unsupported property CpuCredits used for EC2 Instances.
It looks like #1068 caused a breaking change to EC2 instances using `CreditSpecification` property, resulting in the following error when trying to deploy the generated templates:

```
unsupported property CpuCredits
```

It appears the change was implemented to correct an error when deploying LaunchTemplates, however the object type for [LaunchTemplate CreditSpecification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-creditspecification.html) uses a different case than for [EC2 Instance CreditSpecification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance-creditspecification.html).  Looks like separate classes will need to be defined for each use.

johnctitus:
Change corrects the casing for the `CreditSpecification` class's `CPUCredits` object as documented at https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance-creditspecification.html

Change also implements a `LaunchTemplateCreditSpecification` class with the correct case for that object.

Fixes #1099

### DIFF
--- a/troposphere/ec2.py
+++ b/troposphere/ec2.py
@@ -147,7 +147,7 @@ class Placement(AWSProperty):
 
 class CreditSpecification(AWSProperty):
     props = {
-        'CpuCredits': (basestring, False),
+        'CPUCredits': (basestring, False),
     }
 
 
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks @johnctitus! And thank you @cmmeyer for the clarifications with the team.

PR #1104, opened 2018-07-13
Add UsernameAttributes to Cognito

gkrizek:
I was trying to setup a Cognito User Pool and noticed that the UserPool object was missing the `UsernameAttributes` field.

From the docs:

> Specifies whether email addresses or phone numbers can be specified as usernames when a user signs up. Possible values: phone_number or email.
> Required: No
> Type: List of String values

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpool.html#cfn-cognito-userpool-usernameattributes


markpeek:
Thanks!

PR #1106, opened 2018-07-22
Changed policy to AmazonDynamoDBFullAccess for delete and put

davemasino:
AmazonDynamoDBFullAccess is required for put or delete, and is consistent with AWS Labs example


markpeek:
Thanks!

PR #1108, opened 2018-08-01
SimpleTable has more attributes

thebenwaters:
https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlesssimpletable


markpeek:
Thanks!

Issue #1052, opened 2018-05-24
Add ScheduleExpression validation


Jezorko:
The field `troposphere.events.Rule::ScheduleExpression` is a `basestring`. The validity of it is confirmed during the stack creation of CloudWatch.
The documentation for that field can be found [here](https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html).

Would it be possible to validate the cron when JSON is being generated?

phobologic:
Totally possible - it'd likely have to be regex's though.  PRs welcome!

PR #1114, opened 2018-08-12
Schedule expression validation

ChristopherHackett:
This is the enhancement described in the "help wanted" ticket #1052

The behaviour of the validator has been tested agains the values that should be expected.

I'm not entirely sure of the style but this validation will save many users time debugging the cron/rate expressions when using troposphere.

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -220,59 +220,63 @@ class TestValidators(unittest.TestCase):
                 event_schedule_expression(s)
 
         for s in [
-            'cron(* * * * * *)', 'cron(*   *   *   *   *   *)',
+            'cron(* * ? * * *)', 'cron(*   *   ?   *   *   *)',
             # Minutes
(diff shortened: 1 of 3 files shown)


markpeek:
Nice! Thanks!

jonikarppinen:
Hello! It appears that this change broke some things. Valid cron expressions 
that take 6 parameters (as used in [CloudWatch Events](https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html#CronExpressions)) are now reported as invalid:

> ValueError: The following is not a valid expression: "cron(0 1/3 * * ? *)"

Quick fix for anyone else whose build got broken by this change in troposphere 2.3.2:

    pip install -I troposphere==2.3.1

@markpeek Should I open a new issue too?

markpeek:
This change was backed out per #1139 and troposphere 2.3.3 was released to address the regression. Will address adding this in a future release pending more test cases.

PR #1118, opened 2018-08-15
Add get_or_add parameter method

slukes:
When writing reuable methods to add resources and parameters to a template to abstract a common pattern it is often the case that we want to get a parameter that we know may already be inside the template, depending on what the end user has already called.  This PR adds a method which will add a parameter if it's not already there and otherwise return it.


markpeek:
Thanks!

PR #1120, opened 2018-08-16
Added HealthCheckCustomConfig to ServiceDiscovery Service

akerbergen:
Documentation for this property is available here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-servicediscovery-service.html
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-servicediscovery-service-healthcheckcustomconfig.html

It is needed to solve issues with CloudFormation and ServiceDiscovery in ECS, ref: https://forums.aws.amazon.com/thread.jspa?messageID=858305

### DIFF
--- a/troposphere/servicediscovery.py
+++ b/troposphere/servicediscovery.py
@@ -43,6 +43,12 @@ class HealthCheckConfig(AWSProperty):
     }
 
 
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks!

PR #1121, opened 2018-08-16
Tags support for SQS queues

flou:
SQS queues now support tags (cf. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sqs-queues.html#cfn-sqs-queue-tags)


gavihs:
We are looking for this feature, do we have a schedule when this will be promoted

markpeek:
Thanks!

Issue #1124, opened 2018-08-17
ThroughputMode and ProvisionedThroughputInMibps properties missing from efs.FileSystem


tnaroska:
[Two new properties in Cloudformation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-efs-filesystem.html?shortFooter=true#aws-resource-efs-filesystem-properties) not supported by troposphere:
- ThroughputMode
- ProvisionedThroughputInMibps

PR #1126, opened 2018-08-21
Add ThroughputMode and ProvisionedThroughputInMibps to EFS (#1124)

flou:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-efs-filesystem.html#cfn-elasticfilesystem-filesystem-provisionedthroughputinmibps

### DIFF
--- a/troposphere/efs.py
+++ b/troposphere/efs.py
@@ -1,6 +1,26 @@
 from . import AWSObject, Tags
 from .validators import boolean
 
+Bursting = 'bursting'
+Provisioned = 'provisioned'
+
+
+def throughput_mode_validator(mode):
+    valid_modes = [Bursting, Provisioned]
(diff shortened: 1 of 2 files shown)


flou:
Closes #1124

troposphere/efs.py
```diff
@@ -10,6 +30,8 @@ class FileSystem(AWSObject):
         'FileSystemTags': (Tags, False),
         'KmsKeyId': (basestring, False),
         'PerformanceMode': (basestring, False),
+        'ProvisionedThroughputInMibps': (float, False),
```
markpeek: Was the intent to use provisioned_throughput_validator here as a validator?

markpeek:
Thanks! I thought it best to add this PR and you can follow up with the validator if needed.

PR #1129, opened 2018-08-23
Added Webhook,AuthenticationConfiguration and WebhookFilterRule for creating Webhooks

arnemohr:
Added the necessary classes for creating Webhooks via Troposphere/Cloudformation.
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-codepipeline-webhook.html

### DIFF
--- a/troposphere/codebuild.py
+++ b/troposphere/codebuild.py
@@ -161,6 +161,35 @@ class VpcConfig(AWSProperty):
     }
 
 
+class WebhookFilterRule(AWSProperty):
+    props = {
+        'JsonPath': (str, True),
+        'MatchEquals': (str, False)
+    }
+
+
+class AuthenticationConfiguration(AWSProperty):
+    props = {
+        'AllowedIPRange': (str, False),
+        'SecretToken': (str, False),
+    }
+
+
+class Webhook(AWSObject):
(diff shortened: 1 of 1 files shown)


troposphere/codebuild.py
```diff
@@ -161,6 +161,35 @@ class VpcConfig(AWSProperty):
...
+    resource_type = "AWS::CodePipeline::Webhook"
+
+    props = {
+        "Authentication": (str, True),
+        "AuthenticationConfiguration": (AuthenticationConfiguration, True),
+        "Filters": ([WebhookFilterRule], True),
+        "TargetPipeline": (str, True),
+        "TargetAction": (str, True),
+        "Name": (str, False),
+        "TargetPipelineVersion": (int, True),
```
flou: `TargetPipelineVersion` should be `integer` instead of `int`.

troposphere/codebuild.py
```diff
@@ -161,6 +161,35 @@ class VpcConfig(AWSProperty):
     }
 
 
+class WebhookFilterRule(AWSProperty):
+    props = {
+        'JsonPath': (str, True),
```
flou: All string properties should be defined with the type `basestring` instead of `str`.

PR #1111, opened 2018-08-07
AccountAggregationSources must be a list

martinkaberg:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-config-configurationaggregator.html#cfn-config-configurationaggregator-accountaggregationsources

PR #1131, opened 2018-08-24
working aggregation

winjer:
this code generates a valid template that runs.


markpeek:
Thanks! Added via #1111.

PR #1132, opened 2018-08-25
closes #735 Add cloud-config.yaml UserData to basic EC2 example

### LINKED ISSUE #735: We should add an example of using cloud-init
(Other search terms - cloud-init, cloud-config, yaml, yml)

----

I will submit a PR with such an example if it makes sense - was just wondering if there is a design-choice not to have this example, or if it is just by chance.

ciarancourtney:
* Assumes cloud-config.yaml is in same dir as script
* Example substitutes hostname and FQDN
* yaml file also uses cf-style template strings for dual usage
* Modify Test runner to pass __file__ global to exec

### DIFF
--- a/examples/EC2InstanceSample.py
+++ b/examples/EC2InstanceSample.py
@@ -1,9 +1,23 @@
 # Converted from EC2InstanceSample.template located at:
 # http://aws.amazon.com/cloudformation/aws-cloudformation-templates/
+from io import open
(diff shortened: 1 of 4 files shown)


ciarancourtney:
Example test is only failing on Linux & Python 3, the `__file__` global needs to be added to `awacs` package?

PR #1133, opened 2018-08-29
AMAZON_LINUX_2 is now supported by SSM

axelpavageau:
as per https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-patchbaseline.html#cfn-ssm-patchbaseline-operatingsystem , `AMAZON_LINUX_2` is now a valid value


markpeek:
Thanks!

PR #1134, opened 2018-08-30
[codebuild] source_type - use value comparison instead of identity

jaymecd:
Eventually I see error for `CODECOMMIT` source type:

```
ValueError: Source Location: must be defined when type is CODEPIPELINE
```

That was caused by identity check (`is`) of `source_type` string variable.

```python
>>> source_type = 'CODECOMMIT'
>>> source_type is not 'CODEPIPELINE'
True   # <== What?

>>> id(source_type), id('CODEPIPELINE')
(4430947472, 4432006256)
```

Cheers!


markpeek:
Thanks!

PR #1135, opened 2018-08-30
Instance ImageId is no longer required, specifically if using Launch …

DoNotMicrowave:
…Templates


DoNotMicrowave:
Closed -- I will figure out what's up with the unit tests and submit again.

PR #1136, opened 2018-08-31
Return real booleans in the output

ikben:
cfn-lint (https://github.com/awslabs/cfn-python-lint) was giving me an error on booleans (E3012 Property Resources/KmsKey/Properties/Enabled should be of type Boolean), because troposhere renders them to strings.

This makes them also booleans in JSON.

If we want to limit to number of templates that would have their output changed, we may want to do this only if the input was `True` or `False` and not on strings or `1`/`0`.

### DIFF
--- a/tests/examples_output/Autoscaling.template
+++ b/tests/examples_output/Autoscaling.template
@@ -129,13 +129,13 @@
             "Type": "AWS::AutoScaling::AutoScalingGroup",
             "UpdatePolicy": {
                 "AutoScalingReplacingUpdate": {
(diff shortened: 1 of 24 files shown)


ikben:
The failing checks are expected because the output changes, I haven't had the time yet to fix them.

minhio:
any chance of getting this moving forward?

knikolov-epam:
+1

michael-k:
I've picked up the work done here by @ikben in cloudtools/troposphere#1409.  Should be ready for a review.

PR #1143, opened 2018-09-13
Fix amazonmq missing properties

neovenex:


markpeek:
Thanks! Note: I'm going to update the use of "list" to be "[basestring]" which is more precise for type matching.

PR #1145, opened 2018-09-14
Update the OnPremisesInstanceTagFilters parameter for AWS::CodeDeploy::DeploymentGroup

Bazze:
CloudFormation expects the `OnPremisesInstanceTagFilters` parameter of the `AWS::CodeDeploy::DeploymentGroup` resource to be a list of objects and not a single object, [as seen in the documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-codedeploy-deploymentgroup.html#cfn-codedeploy-deploymentgroup-onpremisesinstancetagfilters). This PR fixes that.

I would also like to change the `codedeploy.OnPremisesInstanceTagFilters` to `codedeploy.OnPremisesInstanceTagFilter` (singularis), but not sure how you usually go about doing things like that since it's breaking. If you think it's a good idea, let me know how you usually do it and I will fix it.

### DIFF
--- a/examples/CodeDeploy.py
+++ b/examples/CodeDeploy.py
@@ -1,6 +1,7 @@
 from troposphere import Template
 from troposphere.codedeploy import AutoRollbackConfiguration, \
-    DeploymentStyle, DeploymentGroup, ElbInfoList, LoadBalancerInfo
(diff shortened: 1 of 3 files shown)


markpeek:
Thanks! Actually the name should be "TagFilter". I wish I could have better historical documentation to show why I named things like I did at the time. lol

PR #1148, opened 2018-09-18
Fix AutoScalingRollingUpdate validation failure

calebmarcus:
The validator for AutoScalingGroups assumed that if an AutoScalingRollingUpdate was specified, it would contain the MinInstancesInService attribute, which caused an AttributeError when that argument wasn't passed. However, MinInstancesInService is [optional](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html).

This fixes the assumption that MinInstancesInService is present by adding a default value, and adds tests for the existing intended behavior (to make sure I didn't break it) and to confirm the fix works.

### DIFF
--- a/tests/test_asg.py
+++ b/tests/test_asg.py
@@ -107,6 +107,37 @@ class TestAutoScalingGroup(unittest.TestCase):
         )
         self.assertTrue(group.validate())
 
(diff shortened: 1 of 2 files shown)


markpeek:
Thanks!

PR #1149, opened 2018-09-21
adding UseOnlineResharding policy, per 09/20/2018 update

axelpavageau:
Sept 20th update added a new UpdatePolicy attribute for Elasticache replication groups. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/ReleaseHistory.html


markpeek:
Thanks!

PR #1150, opened 2018-09-21
Add SchedulingStrategy as a prop to ecs.Service

rtkseagray:
This adds support for the newly released `SchedulingStrategy` option in ecs.Service. See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-service.html#cfn-ecs-service-schedulingstrategy for more information on this parameter.

### DIFF
--- a/tests/test_ecs.py
+++ b/tests/test_ecs.py
@@ -39,6 +39,30 @@ class TestECS(unittest.TestCase):
 
         ecs_service.to_dict()
 
+    def test_allow_scheduling_strategy(self):
+        task_definition = ecs.TaskDefinition(
+            "mytaskdef",
+            ContainerDefinitions=[
+                ecs.ContainerDefinition(
+                    Image="myimage",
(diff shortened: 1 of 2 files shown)


yrsurya:
Any tentative date when new version is going to release?

rtkseagray:
@yrsurya I'm not sure what the schedule is for merging PRs into troposphere, but as a workaround, you can monkey patch the `ecs.Service` props to allow the new value:

```
from troposphere import ecs

ecs.Service.props['SchedulingStrategy'] = (basestring, False)
``` 

You'll need to do this before you go to define your template. This also isn't a terribly robust solution, so I'd recommend removing this workaround once a new version of troposphere is available.

markpeek:
Thanks!

dgouldin:
Thanks for merging, @markpeek. Is it possible to get a release cut with this merge commit in it?

PR #1151, opened 2018-09-23
Use basestring to allow percentage definition in MaintenanceWindowTask

ScOut3R:
According to the reference MaxErrors and MaxConcurrency accepts a String so that both a number and a percentage can be defined for these options.


phobologic:
Thanks!

PR #1152, opened 2018-09-24
Add Launch Template capability to Spot Fleets

calebmarcus:
EC2 Spot Fleets can now launch instances from Launch Templates, not just inline Launch Specifications. This adds that functionality to Troposphere.

I ran into a number of cases where the CF documentation didn't make sense, or when tested, didn't agree with CF's behavior -- I've submitted pull requests to their documentation in those cases, and I've written this PR to agree with CF's behavior, rather than the documentation. I'll point those instances out in the PR comments.

### DIFF
--- a/troposphere/ec2.py
+++ b/troposphere/ec2.py
@@ -757,7 +757,7 @@ class LaunchSpecifications(AWSProperty):
     }
 
 
-class SpotFleetLaunchTemplateSpecification(AWSProperty):
+class FleetLaunchTemplateSpecification(AWSProperty):
     props = {
         'LaunchTemplateId': (basestring, False),
         'LaunchTemplateName': (basestring, False),
@@ -784,7 +784,7 @@ class LaunchTemplateOverrides(AWSProperty):
 
 class LaunchTemplateConfig(AWSProperty):
     props = {
-        'LaunchTemplateSpecification': (SpotFleetLaunchTemplateSpecification,
+        'LaunchTemplateSpecification': (FleetLaunchTemplateSpecification,
                                         True),
         'Overrides': ([LaunchTemplateOverrides], False)
     }
--- a/troposphere/ec2.py
+++ b/troposphere/ec2.py
@@ -758,12 +758,18 @@ class LaunchSpecifications(AWSProperty):
 
 
 class SpotFleetLaunchTemplateSpecification(AWSProperty):
-    # Only difference vs. LaunchTemplateSpecification is that Version is optional here
     props = {
         'LaunchTemplateId': (basestring, False),
         'LaunchTemplateName': (basestring, False),
-        'Version': (basestring, False),
+        'Version': (basestring, True),
     }
+
+    def validate(self):
+        template_identifiers = [
+            'LaunchTemplateId',
+            'LaunchTemplateName'
+        ]
+        exactly_one(self.__class__.__name__, self.properties, template_identifiers)
 
 
 class LaunchTemplateOverrides(AWSProperty):
@@ -776,10 +782,9 @@ class LaunchTemplateOverrides(AWSProperty):
     }
 
 
-class LaunchTemplateConfigs(AWSProperty):
-    # Not sure why LaunchTemplateSpecification is not required
+class LaunchTemplateConfig(AWSProperty):
     props = {
-        'LaunchTemplateSpecification': (SpotFleetLaunchTemplateSpecification, False),
+        'LaunchTemplateSpecification': (SpotFleetLaunchTemplateSpecification, True),
         'Overrides': ([LaunchTemplateOverrides], False)
     }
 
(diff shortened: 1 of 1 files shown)


troposphere/ec2.py
```diff
@@ -757,13 +757,47 @@ class LaunchSpecifications(AWSProperty):
     }
 
 
+class SpotFleetLaunchTemplateSpecification(AWSProperty):
```
calebmarcus: This is the same as LaunchTemplateSpecification in the same file -- I wasn't sure whether it's better to create a new one, or use the existing one in a new context (in the CF documentation, all these sub-properties have their own identity in each context, even if otherwise identical.)

troposphere/ec2.py
```diff
@@ -757,13 +757,47 @@ class LaunchSpecifications(AWSProperty):
...
+    props = {
+        'LaunchTemplateId': (basestring, False),
+        'LaunchTemplateName': (basestring, False),
+        'Version': (basestring, True),
```
calebmarcus: Docs say not required, but CF gives an error when not specified, and Version is required in both Instances and ASGs when referencing a Launch Template. See https://github.com/awsdocs/aws-cloudformation-user-guide/pull/162

troposphere/ec2.py
```diff
@@ -757,13 +757,47 @@ class LaunchSpecifications(AWSProperty):
...
+
 class SpotFleetRequestConfigData(AWSProperty):
     props = {
         'AllocationStrategy': (basestring, False),
         'ExcessCapacityTerminationPolicy': (basestring, False),
         'IamFleetRole': (basestring, True),
+        'InstanceInterruptionBehavior': (basestring, False),
         'ReplaceUnhealthyInstances': (boolean, False),
-        'LaunchSpecifications': ([LaunchSpecifications], True),
+        'LaunchSpecifications': ([LaunchSpecifications], False),
```
calebmarcus: Docs still say LaunchSpecifications is required -- but it's not required when LaunchTemplateConfigs is supplied. PR submitted here: https://github.com/awsdocs/aws-cloudformation-user-guide/pull/161

troposphere/ec2.py
```diff
@@ -757,13 +757,47 @@ class LaunchSpecifications(AWSProperty):
...
+        'InstanceType': (basestring, False),
+        'SpotPrice': (basestring, False),
+        'SubnetId': (basestring, False),
+        'WeightedCapacity': (basestring, False)
+    }
+
+
+class LaunchTemplateConfig(AWSProperty):
+    props = {
+        'LaunchTemplateSpecification': (SpotFleetLaunchTemplateSpecification, True),
```
calebmarcus: Docs say LaunchTemplateSpecification is not required. But... it really is. It's meaningless (and doesn't run!) without it.
https://github.com/awsdocs/aws-cloudformation-user-guide/pull/163

troposphere/ec2.py
markpeek: I would suggest using SpotFleetLaunchSpecification and then changing the LaunchSpecifications entry to use it. I'm saying this based on what is in the "specification" files.

calebmarcus: Hey Mark,

Not sure I understand (part of the difficulty in discussing these things is the inconsistencies in naming all across CloudFormation, particularly with plurals!)

The CloudFormation change this PR is intended to support adds a new [LaunchTemplateConfigs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-spotfleet-spotfleetrequestconfigdata.html#cfn-ec2-spotfleet-spotfleetrequestconfigdata-launchtemplateconfigs) property to SpotFleetRequestConfigData, in addition to the LaunchSpecifications property that was already there. So the change doesn't actually relate to LaunchSpecifications.

Each [LaunchTemplateConfig](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-spotfleet-launchtemplateconfig.html) contains a [LaunchTemplateSpecification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-spotfleet-fleetlaunchtemplatespecification.html). Since the properties of that are identical to the ones from [Autoscaling](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-autoscaling-autoscalinggroup-launchtemplatespecification.html) and [EC2](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance-launchtemplatespecification.html), my question was asking whether I should create this new class, or just have LaunchTemplateConfiguration use the existing one for EC2 instances.

Looking at the specification file, it looks like they actually call this a FleetLaunchTemplateSpecification, so perhaps I should call it that.

Does that make more sense?

Thanks!

calebmarcus: Thanks for pointing me to that spec document -- that's pretty useful. Some of the errors I found in the CF documentation are not present in the spec, which makes me wonder how the docs are generated...

markpeek: Yeah, the docs are not always correct. But I've been trying to rely on them for naming guidance. For the above naming try this:
```
make spec
python scripts/gen.py spec/Ec2SpotFleetSpecification.json
```
which is how I got the line:
```
        'LaunchSpecifications': ([SpotFleetLaunchSpecification], False),
```

calebmarcus: Ah, thanks. The LaunchSpecifications line is already in troposphere prior to this PR -- I'll use its suggestion for the new functionality, though.

phobologic:
Sorry, going through the PRs and missed this one - another was merged with this in it however.  Please let me know if you see any issues, and thanks for the PR (and sorry once again for missing it).

calebmarcus:
@phobologic No worries, glad the functionality is in.

PR #1153, opened 2018-09-24
Added ConnectionId and ConnectionType to API GW method integration

akerbergen:
Docs: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apitgateway-method-integration.html#cfn-apigateway-method-integration-connectionid


markpeek:
Thanks!

PR #1154, opened 2018-09-24
Add support for extensible resource definitions in template generator

tylersouthwick:
In macros, a resource type might be defined that needs to be parsed
to be processed. Those resource types might be within the macro transform
or from a different library


markpeek:
Thanks!

PR #1156, opened 2018-10-01
Use dict for RestApi Parameters

dkavanagh:
By supplying value that the previous "[basestring]" accepted, aws would always respond:
"Value of property Parameters must be an object".
The documentation here https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-restapi.html
shows the proper form to be:
"Parameters" : { String:String, ... },

This change allows such values to be set.


markpeek:
Thanks!

PR #1157, opened 2018-10-02
updates rds vpc example and closes #985

### LINKED ISSUE #985: RDS VPC example instance classes are out of date
https://github.com/cloudtools/troposphere/blob/565247a50bd27a8f45c3e408b466b193473cb9a6/examples/RDS_VPC.py#L67-L76

The current list of instance classes is documented here: https://aws.amazon.com/rds/instance-types/

patrickpierson:


phobologic:
Thanks!

Issue #1155, opened 2018-09-25
Add support for serverless Aurora


akamel001:
Cloudformation just added new parameters to [AWS::RDS::DBCluster](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbcluster.html) that allows you to deploy a serverless cluster using the following parameters 

[EngineMode](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbcluster.html#cfn-rds-dbcluster-enginemode) 
The DB engine mode of the DB cluster. Valid values include provisioned or serverless.

[ScalingConfiguration](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbcluster.html#cfn-rds-dbcluster-scalingconfiguration)
For DB clusters in serverless DB engine mode, the scaling properties of the DB cluster.

PR #1164, opened 2018-10-09
Update rds.py

mtryhoen:
take into account the new serverless feature

### DIFF
--- a/troposphere/rds.py
+++ b/troposphere/rds.py
@@ -20,6 +20,7 @@ VALID_LICENSE_MODELS = ('license-included', 'bring-your-own-license',
                         'general-public-license', 'postgresql-license')
 VALID_ENGINEMODE = ('provisioned', 'serverless')
 
+
 def validate_iops(iops):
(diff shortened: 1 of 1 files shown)


troposphere/rds.py
```diff
@@ -18,7 +18,7 @@
                     'aurora-mysql', 'aurora-postgresql', 'mariadb')
 VALID_LICENSE_MODELS = ('license-included', 'bring-your-own-license',
                         'general-public-license', 'postgresql-license')
-
+VALID_ENGINEMODE = ('provisioned', 'serverless')
```
markpeek: I believe you need to add another blank line here to pass the commit tests. You can run "make test" locally to verify.

seddarj:
Didn't see this PR because it doesn't reference the issue #1155...

So I implemented my own version of this which also implements the ScalingConfiguration property (see #1166)

phobologic:
@mtryhoen thanks for this - I'm going to go ahead and close this in favor of #1166 which adds more of the serverless settings.

PR #1169, opened 2018-10-10
Add better method names for Troposphere objects.

nielslaukens:
Some method names are misleading: template.add_metadata() actually
replaces the metadata with the newly given data.

This PR renames these methods from `add_` to `set_`, but keeps the old names for backward compatibility. Optionally, these can be marked with a deprecation warning.

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -571,11 +571,17 @@ class Template(object):
         self.version = None
         self.transform = None
 
-    def add_description(self, description):
+    def set_description(self, description):
         self.description = description
 
-    def add_metadata(self, metadata):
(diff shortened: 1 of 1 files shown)


phobologic:
@nielslaukens spoke with @markpeek - he's good with these changes and adding the deprecation warning, so go ahead whenever you have a chance.  Thanks!

nielslaukens:
I've added deprecation warnings to the old usage. Maybe in a few releases, these can be "upgraded" to a `raise DeprecationWarning()` to hard-fail in these cases.

DrLuke:
Great PR. What further review is required?

PR #1194, opened 2018-11-05
Fix DockerVolumeConfiguration Labels and DriverOpts definition

AntoineMonnet:
AWS DockerVolumeConfiguration expects objects for Labels and DriverOpts.
The current definition leads to a failure when creating the stack : Value of property DriverOpts must be an object,  Value of property Labels must be an object

### DIFF
--- a/tests/test_ecs.py
+++ b/tests/test_ecs.py
@@ -245,7 +245,9 @@ class TestECS(unittest.TestCase):
     def test_docker_volume_configuration(self):
         docker_volume_configuration = ecs.DockerVolumeConfiguration(
             Autoprovision=True,
-            Scope="task"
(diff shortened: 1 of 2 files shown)


markpeek:
This is one of those situations where the [spec files](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) are saying "dict" (map) and the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-dockervolumeconfiguration.html) show list of strings. Be good to get some clarification from AWS on it.

markpeek:
Thanks!

PR #1198, opened 2018-11-08
Fix EC2 SpotFleet properties #1195

### LINKED ISSUE #1195: EC2 SpotFleet unable to create
The release 2.3.4 introduced a bug in the file `troposphere/ec2.py`, class `SpotFleetRequestConfigData`:
- Both `LaunchTemplateConfigs` and `LaunchSpecifications` are set as required, but as per the cloudformation documentation they are conditional (you can only set one of them)

The `validate` function is correctly implementing the conditional logic but in the properties they should be set to false, not true.

maikelpenz:
**Change Summary:**
Fixing the `LaunchTemplateConfigs` and `LaunchSpecifications` properties to not required as per cloudformation documentation you can only specify one of them

**File Changed:**
ec2.py

Fixes #1195


markpeek:
Thanks!

PR #1212, opened 2018-11-19
Fix CodeDeploy EC2TagSet definition

Mallear:
Redefine Ec2TagSet class using `Ec2TagSetList` as a property instead of `Ec2TagSet` as the current definition leads to a failure when creating the stack with status `Encountered unsupported property Ec2TagSet`.


markpeek:
Thanks!

PR #1216, opened 2018-11-20
adding PublicAccessBlockConfiguration to s3 buckets, per Nov 19 2018 …

axelpavageau:
adding PublicAccessBlockConfiguration to s3 buckets, per Nov 19 2018 update


markpeek:
Thanks!

PR #1223, opened 2018-11-21
adding new route53 resources, per Nov 20 2018 update

axelpavageau:


troposphere/route53.py
```diff
@@ -7,6 +7,18 @@
 from .validators import integer, positive_integer, network_port, boolean
 
 
+VALID_RULETYPES = ('SYSTEM', 'FORWARD')
+
+
+def validate_ruletype(ruletype):
+    """Validate database Engine for DBInstance """
```
markpeek: This comment should be updated since it appears to be a copy/paste.

axelpavageau: Fixed, thanks for your review

markpeek:
Thanks!

PR #1225, opened 2018-11-21
adding new EC2Fleet resource, per Nov 20 2018 update

axelpavageau:


troposphere/ec2.py
```diff
@@ -916,3 +916,73 @@ class LaunchTemplate(AWSObject):
...
+    }
+
+
+class FleetLaunchTemplateConfigRequest(AWSProperty):
+    props = {
+        'LaunchTemplateSpecification': (
+            FleetLaunchTemplateSpecificationRequest,
+            False
+        ),
+        'Overrides': (FleetLaunchTemplateOverridesRequest, False),
```
chizou: The docs say this should be a list: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-ec2fleet-fleetlaunchtemplateconfigrequest.html#cfn-ec2-ec2fleet-fleetlaunchtemplateconfigrequest-overrides

troposphere/ec2.py
```diff
@@ -916,3 +916,73 @@ class LaunchTemplate(AWSObject):
...
+
+class EC2Fleet(AWSObject):
+    resource_type = "AWS::EC2::EC2Fleet"
+    props = {
+        'ExcessCapacityTerminationPolicy': (basestring, False),
+        'LaunchTemplateConfigs': (FleetLaunchTemplateConfigRequest, True),
+        'OnDemandOptions': (OnDemandOptionsRequest, False),
+        'ReplaceUnhealthyInstances': (boolean, False),
+        'SpotOptions': (SpotOptionsRequest, False),
+        'TagSpecifications': [(TagSpecifications), False],
```
chizou: I believe the syntax here should be `'TagSpecifications': ([TagSpecifications], False),`

troposphere/ec2.py
axelpavageau: you're right, thanks for your review.

troposphere/ec2.py
axelpavageau: right as well !

markpeek:
Thanks!

PR #1239, opened 2018-11-29
adding billing mode property to DDB tables, per Nov 28 2018 update

axelpavageau:


axelpavageau:
You're right. I've updated both properties to non-mandatory. I'll try to add a better check whenever I can.

petergaultney:
#1243 seems to have some logic around requiring one or the other of BillingMode vs ProvisionedThroughput.

acmcelwee:
LGTM. I tested the branch out to convert a provisioned capacity table to an on demand one, and everything worked mostly as expected. Interestingly, CF allowed the new `BillingMode` set to `PAY_PER_REQUEST` with something still set for `ProvisionedThroughput`, and the table didn't actually the billing model until a second update to actually remove the `ProvisionedThroughput` property. That's obviously just something strange on the CF side, but i figured it was at least worth sharing my observations.

I'd love to see this get merged in, so we can start converting some existing tables.

acmcelwee:
re #1243 vs this one's validation, I generally prefer to leave the conditional logic and validation up to the cloudformation engine, since those things change from time to time.

djk29a:
Agreed that conditional validation should be delegated to CloudFormation and succinctness like this PR is better. Was confused about expectations. However, I made a note in the other PR that Troposphere will fail to validate when using conditionals for the ProvisionedThriughput property (perhaps others affected?) and provided a sample template that deploys successfully but which doesn’t pass Troposphere tests. For my use case, I need a bug fix to Troposphere as well as this PR it seems.

markpeek:
With respect to validation, troposphere attempts to do "best effort" to validate where possible and when someone writes the validation functions to support it. Things like mutual exclusivity of properties is only available from documentation and not part of the spec files. So it takes more work especially as new resources/properties are rolled out.

With respect to this change, likely I will prefer #1243 from @djk29a since it includes more thorough validation and an example test case. I'll try to review and merge these later pending my free time.

axelpavageau:
I'd rather have @djk29a 's PR merged as well. More validation can't hurt.

ntwaddell:
awesome!

PR #1242, opened 2018-11-30
adding new Lambda resources and props, per Nov 29 2018 update

axelpavageau:


markpeek:
Thanks!

troposphere/awslambda.py
```diff
@@ -201,3 +202,34 @@ class Version(AWSObject):
...
+        'S3Key': (basestring, True),
+        'S3ObjectVersion': (basestring, False),
+    }
+
+
+class LayerVersion(AWSObject):
+    resource_type = "AWS::Lambda::LayerVersion"
+
+    props = {
+        'CompatibleRuntimes': (basestring, False),
```
dkavanagh: This should be an list of basestring per https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-layerversion.html

markpeek: @dkavanagh thank you for the review. I just pushed a fix.

PR #1243, opened 2018-12-04
Add on-demand billing for DynamoDB tables

djk29a:
I saw that #1239 is also open for the same feature and that there were a couple suggestions made for adding more validation.

I would appreciate feedback on whether my validation logic can be improved since I couldn't find any other set of properties resembling DynamoDB's billing logic now. I looked at the ELBv2 code and was surprised that it doesn't do more deep validation to model the numerous conditionally required properties / mutually exclusive properties.

Although the `ProvisionedThroughput` AWSProperty has become optional unless the billing mode is `PAY_PER_REQUEST` this does not seem to be consistent when using Global Secondary Indexes because that still requires `ProvisionedThroughput` settings. You can confirm this currently in the console because you can't specify on-demand for the GSI. In a possibly related behavior, updates to the GSI `ProvisionedThroughput` property seem to be getting ignored and failing internally in CloudFormation with a 400 status `One or more parameter values were invalid: List of GlobalSecondaryIndexUpdates is empty`. All I really checked was that CloudFormation validates and creates the stack correctly.

In a slightly unrelated fix, I had a great deal of trouble getting a new use case accepted by the tests and added a fix. A complex object returned by a Conditional didn't validate in Troposphere but is definitely accepted by CloudFormation (the example template deploys nicely). Consequently, I added some extra logic to the template generator code to be a little more lenient. Should I split this PR into a separate bug fix? In self-interest, I'd prefer to keep this PR together but I'll do whatever is desired by maintainers.

Edit: referenced wrong PR

### DIFF
--- a/examples/DynamoDB_Tables_OnDemand.py
+++ b/examples/DynamoDB_Tables_OnDemand.py
@@ -1,0 +1,229 @@
+#!/usr/bin/python
+
+from troposphere import (
+    Template,
+    If,
+    NoValue,
+    Equals,
+    Ref,
+    Output,
+    Parameter
+)
+from troposphere.dynamodb import (
+    KeySchema,
+    AttributeDefinition,
+    Projection,
+    ProvisionedThroughput,
+    Table,
+    GlobalSecondaryIndex
+)
+
+template = Template()
+
+template.add_description("Create two dynamodb tables with "
+                         "conditional on-demand billing. One "
+                         "with global secondary index and one without")
+
+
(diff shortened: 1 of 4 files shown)


petergaultney:
Would love to see this get into Troposphere. We're blocked on adopting the new OnDemand billing structure for some of our environments until Troposphere allows it.

ntwaddell:
yes please!  👍

markpeek:
Nice! Thanks!

rajexcited:
When this change will be released ?

PR #1247, opened 2018-12-07
Correct RoleArn case for OrganizationAggregationSource

akerbergen:
Documentation: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-config-configurationaggregator-organizationaggregationsource.html


markpeek:
Thanks!

PR #1255, opened 2018-12-14
adding new IoTAnalytics resources, per Dec 13 2018 update

axelpavageau:


troposphere/iotanalytics.py
```diff
@@ -0,0 +1,248 @@
+from . import AWSObject, AWSProperty, Tags
```
markpeek: Could you add a copyright and license comment to this file?

axelpavageau: sure, it's done.

markpeek:
Thanks!

PR #1241, opened 2018-11-29
Adding new properties to elbv2 Listener Action

dgouldin:
* AuthenticateCognitoConfig
 * AuthenticateOidcConfig
 * FixedResponseConfig
 * Order
 * RedirectConfig

Resolves #1240

PR #1258, opened 2018-12-19
Added new aws properties in elbv2

wesleywh:
A series of new additions have been made by aws to elbv2 properties. These are the new listener actions available:
* AuthenticateCognitoConfig
* AuthenticateOidcConfig
* FixedResponseConfig
* RedirectConfig
* Order

These series of changes add support for these. For more information refer to their documentation here:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-listenerrule-actions.html#cfn-elasticloadbalancingv2-listenerrule-action-authenticatecognitoconfig

One slight change I have made is that TargetGroupARN is required in their docs but I made it conditional because you can't include that when you use certain actions like the AuthenticateCognitoConfig.

I tried to keep this in the same format that I have been seeing in this repository.

I don't know if these have been added in the elb as well because I haven't looked due to the fact that I am not using it.

### DIFF
--- a/troposphere/elasticloadbalancingv2.py
+++ b/troposphere/elasticloadbalancingv2.py
@@ -23,12 +23,60 @@ class Certificate(AWSProperty):
     }
 
 
+class AuthenticateCognitoConfig(AWSProperty):
+    props = {
+        'AuthenticationRequestExtraParams': (dict,False),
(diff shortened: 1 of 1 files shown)


wesleywh:
Oh this is a duplicate of #1241 . So I am going to close this.

PR #1263, opened 2018-12-30
Add DeleteAutomatedBackups to RDS DBInstance

rouge8:
This was part of the Nov. 19 update:

> Use the DeleteAutomatedBackups property to indicate whether automated backups should be deleted (true) or retained (false) when you delete a DB instance. The default is true.


markpeek:
Thanks!

PR #1266, opened 2019-01-03
Fix EC2Fleet class definition to match functional correctness of Clou…

Canuteson:
The LaunchTemplateConfigs property of AWS::EC2::EC2Fleet is required to be a list of FleetLaunchConfigRequest objects
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-ec2fleet.html

While the CloudFormation documentation currently states that ValidFrom and ValidUntil type is `Integer`, the details property description states:
```
date and time of the request, in UTC format (for example, YYYY-MM-DDTHH:MM:SSZ)
```
Troposphere incorrectly diverges from the CloudFormation documentation in the preceeding case.

CloudFormation actually requires `ValidUntil` and `ValidFrom` to be an ISO 8601 formatted string, similar to the corresponding properties in AWS::EC2::SpotFleet property `SpotFleetRequestConfigData`
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-spotfleet-spotfleetrequestconfigdata.html

Troposphere currently follows the CloudFormation documentation in this case, which is actually incorrect.

### DIFF
--- a/troposphere/ec2.py
+++ b/troposphere/ec2.py
@@ -1066,7 +1066,7 @@ class EC2Fleet(AWSObject):
     resource_type = "AWS::EC2::EC2Fleet"
     props = {
         'ExcessCapacityTerminationPolicy': (basestring, False),
-        'LaunchTemplateConfigs': (FleetLaunchTemplateConfigRequest, True),
+        'LaunchTemplateConfigs': ([FleetLaunchTemplateConfigRequest], True),
         'OnDemandOptions': (OnDemandOptionsRequest, False),
         'ReplaceUnhealthyInstances': (boolean, False),
         'SpotOptions': (SpotOptionsRequest, False),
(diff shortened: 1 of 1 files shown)


cmmeyer:
CloudFormation should get this fixed in the docs as well (https://github.com/awsdocs/aws-cloudformation-user-guide) -- has anyone issued a PR for this over there yet?

Canuteson:
> Thanks!
> 
> BTW, for a future issue, should there be an ISO 8601 validator for ValidUntil/ValidFrom?

That sounds like a good idea.

Canuteson:
> CloudFormation should get this fixed in the docs as well (https://github.com/awsdocs/aws-cloudformation-user-guide) -- has anyone issued a PR for this over there yet?

https://github.com/awsdocs/aws-cloudformation-user-guide/pull/230

Issue #977, opened 2018-02-05
TemplateGenerator is broken


skantzidis:
When attempting to use the TemplateGenerator on a well defined input:

	{
		"AWSTemplateFormatVersion": "2010-09-09",
		"Resources": {
			"AccessPolicy": {
				"Type": "AWS::IAM::ManagedPolicy",
				"Properties": {
					"Description": "Access to data",
					"Path": "/",
					"PolicyDocument": {
						"Version": "2012-10-17",
						"Statement": [
							{
								"Action": [
									"s3:*"
								],
								"Effect": "Allow",
								"Resource": [
									"arn:aws:s3:::myTestBucket",
									"arn:aws:s3:::myTestBucket*"
								]
							}
						]
					}
				}
			}
		}
	}

I get the following error:

	Traceback (most recent call last):
	  File "C:\Users\AppData\Local\Programs\Python\Python35\lib\site-packages\IPython\core\interactiveshell.py", line 2910, in run_code
		exec(code_obj, self.user_global_ns, self.user_ns)
	  File "<ipython-input-7-9d2021fc9486>", line 1, in <module>
		TemplateGenerator(test5)
	  File "C:\Users\AppData\Local\Programs\Python\Python35\lib\site-packages\troposphere\template_generator.py", line 63, in __init__
		self._get_resource_type_cls(v)
	  File "C:\Users\AppData\Local\Programs\Python\Python35\lib\site-packages\troposphere\template_generator.py", line 161, in _convert_definition
		return self._create_instance(expected_type, args, ref)
	  File "C:\Users\AppData\Local\Programs\Python\Python35\lib\site-packages\troposphere\template_generator.py", line 274, in _create_instance
		return cls(title=ref, **args)
	  File "C:\Users\AppData\Local\Programs\Python\Python35\lib\site-packages\troposphere\__init__.py", line 124, in __init__
		self.__setattr__(k, v)
	  File "C:\Users\AppData\Local\Programs\Python\Python35\lib\site-packages\troposphere\__init__.py", line 201, in __setattr__
		self._raise_type(name, value, expected_type)
	  File "C:\Users\AppData\Local\Programs\Python\Python35\lib\site-packages\troposphere\__init__.py", line 219, in _raise_type
		expected_type))
	TypeError: <class 'troposphere.iam.ManagedPolicy'>: AccessPolicy.PolicyDocument is <class 'list'>, expected (<class 'dict'>,)

[...]

PR #1270, opened 2019-01-04
Fix nested 'Name' sections in Output import

saineju:
TemplateGenerator seems to generate nested 'Name' sections to Exports in Output section as mentioned in issue #977, this PR should fix the nesting


markpeek:
Thanks!

PR #1273, opened 2019-01-07
adding a S3OriginConfig object to distinguish between Distribution and StreamingDistribution properties

### LINKED ISSUE #1272: S3Origin doesn't have a DomainName property
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-s3originconfig.html doesn't list the `DomainName` property, but it was added as mandatory by #1235 , introducing a regression.

axelpavageau:
should fix #1272


markpeek:
This one isn't as straightforward. There are two S3Origin in the doc, [Distribution S3Origin](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-s3originconfig.html) and [StreamingDistribution S3Origin](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-streamingdistribution-s3origin.html). I believe the correct one is the latter StreamingDistribution S3Origin from looking at the specification files where the Distribution one is named S3OriginConfig (also the url for that one shows aws-properties-cloudfront-distribution-s3originconfig.html).

Changes needed:
- Add both S3Origin and S3OriginConfig
- Use the appropriate one in StreamingDistribution and Distribution
- The example appears to be using Distribution which then would not have DomainName
- Likely be good to add a StreamingDistribution example.

Adding @stevengorrell (author of PR #1235) for his opinion and comments. Looks like this is another change needed to the doc repo, @cmmeyer if this looks correct I can file a doc Issue on it.

axelpavageau:
You're right. I didn't see that as I don't use StreamingDistributions and wasn't familiar with its properties. I've updated the PR and will add a StreamingDistribution example later on

markpeek:
Nice! Thanks for the quick turnaround on this issue.

jhooyberghs:
Is it possible to do a bugfix release with this fix included? This same regression is breaking some things for me at the moment with the 2.4.0 release

markpeek:
I'm waiting for a review/merge for PR #1277 and then if no other regressions come in I'll do a 2.4.1 release.

markpeek:
@jhooyberghs just released [2.4.1](https://github.com/cloudtools/troposphere/releases/tag/2.4.1)

PR #1274, opened 2019-01-08
SSM Example.

Caprowni:
This change will aim to add an example for SSM and it's patch baseline/filter capabilities.  Currently there isn't a template for this so adding this for future reference incase anyone needs to use this feature within Cloudformation.


Caprowni:
Not sure how the build failed with an assertion on expected vs actual output since the template I provided in template outputs is printed from the troposphere code.

markpeek:
Thank you for adding this new example.

PR #1275, opened 2019-01-08
cloudfront - remove unsupported property

vsimakhin:
This property (DomainName) is not supported for S3Origin in CloudFormation https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-s3originconfig.html
neither in boto3 api https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudfront.html#CloudFront.Client.create_distribution
```
'S3OriginConfig': {
                        'OriginAccessIdentity': 'string'
                    },
```

The stack is rolled back with error 
```
Property validation failure: [Encountered unsupported properties in {/DistributionConfig/Origins/0/S3OriginConfig}: [DomainName]]
```

### DIFF
--- a/examples/CloudFront_StreamingDistribution_S3.py
+++ b/examples/CloudFront_StreamingDistribution_S3.py
@@ -31,7 +31,7 @@ myDistribution = t.add_resource(StreamingDistribution(
     StreamingDistributionConfig=StreamingDistributionConfig(
         Comment="Streaming distribution",
         Enabled=True,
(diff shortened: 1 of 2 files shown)


markpeek:
Thank you for the PR. However this was just fixed in master via #1273. Note: you might want to look at the discussion as there was a naming collision due to incorrect AWS docs.

PR #1276, opened 2019-01-08
adding better validation for AWS::CloudWatch::Alarm properties

axelpavageau:


markpeek:
@girishkhadke what you seem to be implying is Metric, Statistic, and ExtendedStatistic are all mutually exclusive. So should the validate be implemented as:
```
def validate(self):
        conds = [
            'ExtendedStatistic',
            'Metric',
            'Statistic',
        ]
        exactly_one(self.__class__.__name__, self.properties, conds)
```

axelpavageau:
I've updated the PR, but I believe the cloudformation attribute is `MetricName` not `Metric` ?

girishkhadke:
According to AWS documentation, the name of attribute used to specify a metric based on math expression is : "Metrics" 

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cw-alarm.html#cfn-cloudwatch-alarm-metrics

girishkhadke:
This is the error I was getting on Cloudformation Stack create/update:

```
Statistic should not be set if list of Metrics is set. (Service: AmazonCloudWatch; Status Code: 400; Error Code: ValidationError; Request ID: *****)
```

axelpavageau:
allright. I've updated the PR.

markpeek:
Thanks for the quick turnaround on a fix.

girishkhadke:
Thank you.

Issue #1245, opened 2018-12-06
ValueError: Resource FromPort required in type AWS::EC2::SecurityGroupEgress


medwig:
Attempting to add a SecurityGroupEgress rule across all IpProtocol fails validation if no ports are specified:

```
from troposphere import Ref, Template, Tags
from troposphere.ec2 import VPC, SecurityGroupEgress, SecurityGroup

t = Template()

vpc = t.add_resource(VPC(
    "VirtualPrivateCloud",
    CidrBlock="10.0.0.0/16",
    InstanceTenancy="default",
    EnableDnsSupport=True,
    EnableDnsHostnames=False,
))

defaultVpcSecurityGroup = t.add_resource(
    SecurityGroup(
        'DefaultVpcSecurityGroup',
        GroupDescription="Default VPC security group",
        VpcId=Ref(vpc),
        Tags=Tags(
            Name="DefaultVpcSecurityGroup"
        )
    )
)

defaultVpcSecurityGroupEgressRule = t.add_resource(
    SecurityGroupEgress(
        "defaultVpcSecurityGroupEgressRule",
        Description="foobar",
        GroupId=Ref(defaultVpcSecurityGroup),
        CidrIp="0.0.0.0/0",
        IpProtocol='-1',
        # FromPort='-1',
        # ToPort='-1',
        DependsOn=defaultVpcSecurityGroup,
    )
)

print(t.to_yaml())

-----------------------------------------------------------------------------
ValueError: Resource FromPort required in type AWS::EC2::SecurityGroupEgress (title: defaultVpcSecurityGroupEgressRule)
```

In reality it works just fine without those attributes, if you open the resulting cloudformation template and delete `FromPort` & `ToPort` the stack will succeed. In fact, if `FromPort` & `Toport` are specified and `IpProtocol: '-1'` for stack creation then updates to the rule (for example adding a description) will fail stack updates unless those two lines are removed. This seems to be an issue with the EC2 docs.

markpeek:
Fixed via #1277

Issue #1208, opened 2018-11-15
Cloudformation Drift and Security Group Ports


daugustus:
I appreciate the hard work that has gone into Troposphere- I really like it and use it in my daily work.

It appears that with the latest offering of Drift Detection deals with the port numbers. In my case, I see that when the ip protocol is specified as '-1', CF does NOT want port numbers specified. However, Troposphere requires the ports. The result of this difference appears in the drift detection console in AWS. :(

Thanks,
Dave

orenmazor:
Ha. just got nailed by this.

markpeek:
Fixed via #1277

PR #1277, opened 2019-01-09
Allow empty To/From port ranges for SG's for certain IP protocols

markpeek:
Here is a proposed change to allow ToPort and FromPort to not be required. This should address #1208 and #1245.

Any one want to code review or test?
@daugustus @orenmazor @medwig @phobologic @axelpavageau


markpeek:
Fixes #1208 
Fixes #1245

PR #1288, opened 2019-01-15
Add missing region information.

clokep:
This adds some missing constants, not a huge deal, but seemed good to add so that they're all there. Thanks for making troposphere!

Adds the ap-south-1, ca-central-1, eu-west-3, eu-north-1, and us-east-2. Also adds some missing availability zones for existing regions.

See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html

I used their CLI tool to check the availability AZs in each region:

```bash
aws ec2 describe-availability-zones --region region-name
```

There's also an "ap-northeast-3 / Asia Pacific (Osaka-Local)" region, but you have to enable that on your account first (or something like that), regardless it wasn't available for me.


markpeek:
Thanks!

clokep:
Awesome! Thank you. 🥇

PR #1292, opened 2019-01-18
ResourceGroupArn is no longer mandatory for AWS::Inspector::AssessmentTarget, per 2019 Jan 17 update

axelpavageau:
…tTarget, per 2019 Jan 17 update


markpeek:
Thanks!

PR #1293, opened 2019-01-18
fix creating RotationSchedule for SecretsManager

mmenard-bbc:


troposphere/secretsmanager.py
```diff
@@ -46,7 +46,7 @@ class RotationSchedule(AWSObject):
     props = {
         'SecretId': (basestring, True),
         'RotationLambdaARN': (basestring, True),
-        'RotationRules': ([RotationRule], False)
+        'RotationRules': (RotationRule, False)
```
markpeek: Looking at the docs and the resource spec files, the RotationRule property should be names RotationRules (not the plural). Could you make this change as well?

markpeek:
Thanks!

PR #1302, opened 2019-01-28
Add Python 3.7 to travis testing

markpeek:


michael-k:
FYI: `AttributeError: 'functools.partial' object has no attribute '__qualname__'` from the test run will be fixed in Python 3.7.3, see https://bugs.python.org/issue35767

markpeek:
Thanks @michael-k. Yes, I am waiting for that fix to be released to make 3.7 part of the travis run.

.travis.yml
```diff
@@ -1,10 +1,12 @@
 language: python
-sudo: false
+sudo: required
```
michael-k: This is no longer necessary and the `sudo` keyword is deprecated: https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

.travis.yml
```diff
@@ -1,10 +1,12 @@
...
+dist: xenial
 python:
   - "2.7"
   - "3.4"
   - "3.5"
   - "3.6"
+  - "3.7"
```
michael-k: This still refers to *3.7.1*. You can replace it with either `3.7.3` (released [2019-03-25](https://pythoninsider.blogspot.com/2019/03/python-373-is-now-available.html)) or `3.7-dev` (*3.7.3+* at the moment).

.travis.yml
markpeek: Applied. Thanks!

.travis.yml
markpeek: Applied. Thanks!

PR #1306, opened 2019-02-01
Fix for AWS Lambda reserved environment variables

cedricATBQ:
Regarding AWS doc (https://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html#lambda-environment-variables) 'LANG', 'LD_LIBRARY_PATH', 'NODE_PATH', 'PATH' and 'PYTHONPATH' are not reserved and can be overrode.


markpeek:
Thanks!

PR #1307, opened 2019-02-01
[events] add SqsParamaters support to Rule Target

jaymecd:
Contains the message group ID to use when the target is a FIFO queue. 
See for more info https://docs.aws.amazon.com/AmazonCloudWatchEvents/latest/APIReference/API_Target.html#CWE-Type-Target-SqsParameters

### DIFF
--- a/troposphere/events.py
+++ b/troposphere/events.py
@@ -39,6 +39,12 @@ class RunCommandParameters(AWSProperty):
     }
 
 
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks! This does not appear in the documentation ([bug filed](https://github.com/awsdocs/aws-cloudformation-user-guide/issues/264)). It does show up in the resource specification files.

PR #1308, opened 2019-02-01
Sphinx docs

lowcloudnine:
## Sphinx Documentation
Since I've become a regular consumer and reader of the docs, particularly the examples I wanted to sphinx-ify them.  This also allows the project to integrate with read-the-docs in the future.  

The changes can be broken down into 3 categories:

1.  docs - large changes and where 95-99% of the changes are.  Used sphinx-quickstart and the added the needed files from there.  Uses sphinx and sphinx-rtd-theme, so this is an added dependency.  I also added docs/_build to the gitignore file as the build directory is only used for generated product not code.

2.  examples - to get the examples to play with sphinx an __init__ file is needed (and added) and I changed the ApiGateway to a function with a name call at the end.  I only did one, so if this is not method the project likes I didn't waste but 5 mins.

3. helpers/usersdata - I added some new lines to the docstring in from_file so sphinx would stop warning me about it.

You can generate the html for the docs by cd into the docs dir and running `make html`.  Then point a web browser to the index.html in the docs/_build/html folder.

Feedback is welcome and I understand if you don't want to go to using sphinx.


markpeek:
I was looking forward to seeing it pass the tests and be included. Did you mean to close this? Were there issues you needed help with?

lowcloudnine:
Mark, cloudtools, et al.,

Yes, I did mean to close this.  Every time I would fix one test problem another would rear it's head, so I closed it until I understand the issues better.  No need to run Travis every time I commit.

I do have one question, when Travis runs does it use something like 2to3 to convert the code to Python 3?  I ask because when I run in a Python 3 env it fails locally due to the basestring not existing and my research showed this to be a Python 2/3 compatibility issue, i.e. basestring doesn't exists in Python 3.  What am I missing?

This is first time I'll be contributing to a OS project in over a year and a half so please excuse any faux pas.

Thank you,
Jon

________________________________

lowcloudnine:
Never-mind the Travis question, found the answer, in the travis.yaml.  Like I said, please forgive the silliness.
________________________________

PR #1311, opened 2019-02-05
Sphinx docs

lowcloudnine:
# Sphinx Docs

This is an attempt to Sphinx-ify the documentation, to consolidate and make it so it could be automated to integrate with read-the-docs in the future.  Feedback is welcome.  It's very basic, no logo, etc. and much work remains this is just a starting point.

Round 2 -- I removed the examples from the docs until I can figure out how to make tests pass when the example is converted to a function.  There's a placeholder file for the examples, but it can be removed if needed.

Changes not in the docs directory:

1.  Moved docs/requirements.txt file to the main directory as requirements-dev.txt.
2.  Added the docs/_build directory to .gitignore.
3.  Changed the docstring in troposphere/helpers/userdata.py so Sphinx wasn't warning on it.  Just 3 newlines.


docs/contribute.rst
```diff
@@ -0,0 +1,45 @@
...
+
+The latest test scripts can be found at https://travis-ci.org/cloudtools/troposphere.
+If you look at the details of a job, you can see what automated tests
+will be run against any commits to the project.
+
+1. Create a virtualenv (e.g. `virtualenv ~/virtualenv/troposphere`)
+#. Activate it: `source ~/virtualenv/troposphere/bin/activate`
+#. `pip install --upgrade pip setuptools wheel`
+#. `pip install -r requirements-dev.txt`
+#. Run tests:
```
markpeek: I run "make test" instead of these individual commands.

docs/index.rst
```diff
@@ -0,0 +1,376 @@
+.. troposphere documentation master file, created by
+   sphinx-quickstart on Fri Feb  1 14:27:25 2019.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+===========
```
markpeek: This appears to be a copy of README.rst from the top-level. Rather than keeping this in sync, perhaps it could be changed to just do:
```
.. include:: ../README.rst
```
This also begs the question as to whether the TOC should come first.

docs/troposphere_iz.rst
```diff
@@ -0,0 +1,20 @@
+troposphere modules i - z
```
markpeek: This seems to be missing quite a few modules (rds, s3 to name a couple).

Also, should these files be auto-generated?

markpeek: BTW, also noticed it was missing "h" between "a-g" and "i-z". And, as an example of autogenerating the list, something like this is quick and dirty:
```
ls ../troposphere/[a-i]*.py | sed 's#../troposphere/\(.*\).py#\1#' | xargs printf ".. automodule:: troposphere.%s\n    :members:\n\n"
```

docs/index.rst
lowcloudnine: Concur with the include it is changed/done.  The TOC is now first, this to me is a matter of taste as the TOC is also in the sidebar.

docs/troposphere_iz.rst
lowcloudnine: Agree with the these should be autogenerated but I wasn't sold on my original organization of the files alphabetically.  After some thought, it might be better to organize the modules as AWS does, i.e. by Compute, Storage, Database, Networking, etc.  While more work it would be more native and possibly flow better for first time users.

An autogenerated alphabetical file could be added to ensure complete module coverage and provide an additional way to browse the libraries.

docs/index.rst
markpeek: Yeah, I wasn't sure which one looked better. I'll let you decide on where the TOC goes. :-) Feel free to push the "include" change and I'll merge.

docs/troposphere_iz.rst
markpeek: The alphabetical does mimic the CloudFormation docs.

lowcloudnine: Then autogenerate it is!  I'm assuming you'll put your snippet in the Travis yaml?

docs/index.rst
lowcloudnine: After looking at each several times, I went with TOC at the bottom by index.  It changed be changed easily, especially since the README is now included, thanks for that tidbit.

markpeek:
Thank you very much! Now available on [readthedocs](https://troposphere.readthedocs.io/en/latest/).

PR #1314, opened 2019-02-11
Docs update

lowcloudnine:
I was inspired by Mark's use of the Makefile and figured there must be more Sphinx like way to automate api doc creation.  Found sphinx-apidoc and used it to build out the examples, troposphere and tests packages.  Added the needed commands to the Makefile.  In doing so I noticed GitHub complained about the Contribute file not being available so using the include method I linked the various docs in the main directory to the sphinx docs.

There is one change outside of the rST and docs/ locations.  I updated setup.py to dynamically read in the version.  I also added the ability to read in the requirements from requirements.txt and added specific version of the various libraries in that file.  I believe it will make dependency management easier.

As before feedback is welcome and hope this helps.


CODE_OF_CONDUCT.rst
```diff
@@ -0,0 +1,85 @@
+Code of Conduct
```
markpeek: I'll use this as a general comment. Given markdown is the norm for github docs and rst is the norm for sphinx docs, did you look at or consider integrating a markdown extension into sphinx? Such as, https://www.sphinx-doc.org/en/master/usage/markdown.html . I'm asking if this is an alternative to moving all the .md to .rst but note I'm not requiring you use this but wanted to have a conversation on alternatives. Note: the README.rst file was changed from .md awhile back to have been rendering inside of PyPI.

troposphere/__init__.py
```diff
@@ -39,6 +39,9 @@
 
 valid_names = re.compile(r'^[a-zA-Z0-9]+$')
 
+if sys.version_info[0] == "3":
```
markpeek: There are a few instances of this added to files. The use of 2to3 should not require these changes. Could you elaborate on why this was required?

CODE_OF_CONDUCT.rst
lowcloudnine: To be frank, no I didn't consider it and I'm not opposed to doing it that way.  I tend to be a purist and as such generally avoid mixing of different markdowns, yaml and json or Python and bash, etc, this is for various reasons; some valid, some not.  While not native rST is "well" supported on both GitHub and GitLab and the purist in me would prefer to keep to one markdown language and to stick to the Sphinx way of doing things.  It also avoids yet one more dependency.  That said rST is not as well known as md and it might be easier for new contributors to work with.  So, I'll yield to your preference.  Let me know your decision and I'll resolve any issues.

troposphere/__init__.py
lowcloudnine: Not happy this is in the commit.  I had looked at converting the code base to be 2 and 3 compatible without the need for 2to3.  In my mind this would be a first refactor on the road to Python 3 native and I was going to comment in the appropriate issues at a later time.  These were some minor experiments and I'll remove and re-commit later today/tonight when I have time.  My apologies.

CODE_OF_CONDUCT.rst
markpeek: I chatted with @phobologic and he was a +1 on converting to rst. Let's keep that part of your commit and I'll await your commit later to revert the basestring changes. Thanks!

markpeek:
Nice! Thanks!

michael-k:
Are you aware that with this PR, `awacs` is no longer a soft, but a hard dependency?

It's listed in requirements.txt which is used for `install_requires` in `setup.py`.

https://github.com/cloudtools/troposphere/blob/42041ba3b7111ee9b7d6b9abac7af43d341c3971/requirements.txt#L6

https://github.com/cloudtools/troposphere/blob/42041ba3b7111ee9b7d6b9abac7af43d341c3971/setup.py#L86

Another thing I observed is, that `tests_require` and `extras_require` do not reflect the introduced minimal version boundary (>=0.8) for awacs.

https://github.com/cloudtools/troposphere/blob/42041ba3b7111ee9b7d6b9abac7af43d341c3971/setup.py#L88-L89

PR #1322, opened 2019-02-18
adding new AWS::FSx::FileSystem resource, per Feb 2019 15 update

axelpavageau:


troposphere/fsx.py
```diff
@@ -0,0 +1,43 @@
...
+
+    props = {
+        'BackupId': (basestring, False),
+        'FileSystemType': (basestring, False),
+        'KmsKeyId': (basestring, False),
+        'LustreConfiguration': (LustreConfiguration, False),
+        'SecurityGroupIds': ([basestring], False),
+        'StorageCapacity': (integer, False),
+        'SubnetIds': ([basestring], False),
+        'Tags': ((Tags, list), False),
```
markpeek: I have been deprecating the use of a list in Tags so the accpeted type should should be ```Tags``` and not ```(Tags, list)```.

axelpavageau: good catch, that was copied from an older resource.

markpeek:
Thanks!

PR #1326, opened 2019-02-19
Drop support for Python 3.4

michael-k:
It will reach its EOL in March 2019, see https://www.python.org/dev/peps/pep-0429/#release-schedule

It has aleady been proposed in cloudtools/troposphere#1295 that Python 3.5 should be the minimal required version.  This PR addresses a low hanging fruit on the road towards this desirable state. :)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -82,7 +82,7 @@ setup(
         'scripts/cfn2py'
     ],
 
-    python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*",
+    python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -83,6 +82,7 @@ def get_version():
         'scripts/cfn2py'
     ],
 
+    python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*",
```
markpeek: I'm fine with removing testing and mention of 3.4 but I'm not sure I want to explicitly prevent use on 3.4...or at least until it is officially EOL. Could you adjust this to remove 3.4 for now? Thanks.

michael-k: Done

markpeek:
Thanks!

PR #1331, opened 2019-02-22
Add new Apikey property

oussemos:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-apikey.html#cfn-apigateway-apikey-value


markpeek:
Thanks!

PR #1334, opened 2019-02-27
Add psuedo-parameter Ref for AWS::Partition

schlarpc:
This change just adds a pre-made Ref object for `AWS::Partition`, an unloved and comparatively recent addition to the list of psuedo-parameters.


markpeek:
Thank you for giving AWS::Partition some love.

PR #1338, opened 2019-03-05
Prefer awacs.aws.PolicyDocument over awacs.aws.Policy

michael-k:
`awacs.aws.PolicyDocument` was introduced in awacs 0.7.1 to prevent confusions with `troposphere.iam.Policy`.  troposphere requires awacs >= 0.8.

### DIFF
--- a/troposphere/ec2.py
+++ b/troposphere/ec2.py
@@ -4,19 +4,12 @@
 # See LICENSE file for full license.
 
 from . import AWSHelperFn, AWSObject, AWSProperty, Tags
+from .compat import policytypes
 from .validators import (
     boolean, exactly_one, integer, integer_range, double,
     network_port, positive_integer, vpn_pre_shared_key, vpn_tunnel_inside_cidr,
     vpc_endpoint_type
 )
-
-try:
-    from awacs.aws import Policy
-
-    policytypes = (dict, Policy)
-except ImportError:
-    policytypes = dict,
-
 
 VALID_ELASTICINFERENCEACCELERATOR_TYPES = ('eia1.medium', 'eia1.large',
                                            'eia1.xlarge')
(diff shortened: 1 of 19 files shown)


troposphere/ec2.py
```diff
@@ -11,9 +11,9 @@
 )
 
 try:
-    from awacs.aws import Policy
+    from awacs.aws import PolicyDocument
```
markpeek: I'd been holding off on this PR in order to verify backward compatibility. I think it would be best to change the troposphere/*.py files to use both Policy and PolicyDocument.
```
try:
    from awacs.aws import Policy, PolicyDocument
    policytypes = (dict, Policy, PolicyDocument)
except ImportError:
    policytypes = dict,
```
This allows existing code to still function while supporting PolicyDocument and promoting its use in the examples.

michael-k: Yeah, you're right.  Since PolicyDocument is a subclass of Policy, Policy would no longer be valid.

But forward compatibility should also be considered.  Let's assume that awacs 1.0 drops Policy.  The current release of troposphere would not be compatible with it.  Therefore I propose the following code:

```python
try:
    from awacs.aws import Policy, PolicyDocument
    policytypes = (dict, Policy, PolicyDocument)
except ImportError:
    try:
        from awacs.aws import PolicyDocument
        policytypes = (dict, PolicyDocument)
    except ImportError:
        policytypes = dict,
```

That are imho too many lines of code that would be duplicated over many files. I therefore also propose to introduce a new module `troposphere.compat` (like eg. [django rest framework](https://github.com/encode/django-rest-framework/blob/master/rest_framework/compat.py)) and import `policytypes` from there.

markpeek: Yes, adding it via a compat library would be good. Thanks!

markpeek:
Thanks @michael-k!

PR #1352, opened 2019-03-21
Add authenticate-cognito and authenticate-oidc to elb v2 Action's "type" validator

bangfalse:
Supporting metadata types were added, but can't be used properly without changing the "Type" parameter.


markpeek:
Thanks!

PR #1353, opened 2019-03-22
Update the instance types in constants.

clokep:
I was using a couple instance types that weren't included, so went through and added a bunch. You can see https://aws.amazon.com/ec2/instance-types/ for reference.

The `u-xtb1.metal` ones I wasn't really sure how to format, so I turned the `-` into a `_`, I also debated just removing the `-`.

Note that there were some duplicates in here, which I removed. I checked for more with:

```bash
sort troposphere/constants.py | uniq -c | sort
```


troposphere/constants.py
```diff
@@ -158,7 +187,9 @@
 G2_2XLARGE = 'g2.2xlarge'
 G2_8XLARGE = 'g2.8xlarge'
 
+G3S_XLARGE = 'g3s.xlarge'
 G3_2XLARGE = 'g3.2xlarge'
```
markpeek: While you didn't add it, it looks like G3_2XLARGE isn't a valid option. Could you verify and remove as well?

clokep: I was wondering about this since I couldn't find it anywhere. If you Google it there are some references to it, but none seem to be from AWS documentation. https://www.ec2instances.info/ also does not list it. I went through the "launch instance" workflow on us-east-1 and didn't see this as an option either.

I've pushed a commit to remove it.

markpeek: Thanks for the quick fix.

markpeek:
Thanks!

clokep:
Thanks for the quick merge!

PR #1363, opened 2019-04-09
EngineAttributes should take list

wynnsusilo:
EngineAttributes variable in the Server class is currently taking a single EngineAttribute object, but Cloudformation allows for a list of attributes, as is necessary when using puppet R10K. See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-opsworkscm-server.html#cfn-opsworkscm-server-engineattributes


markpeek:
Thanks!

PR #1364, opened 2019-04-11
Fix the scripts for Python3

allyunion:
Running the scripts through pylint gives a lot of errors...  It seems like these scripts are out of date and need a little TLC...

This is just a basic patch to get the code working in Python3, though I'm not sure if the basestring hack is the best approach below.


markpeek:
Thanks! I'm going to tweak gen.py to add a #! header and the print_function import.

PR #1372, opened 2019-04-17
Fixed: Codebuild Webhook Filters are to be a list of list of WebhookFilter

JohnPreston:
According to documentation, Web Hooks should be written in the form of List of List of WebHooks
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codebuild-project-projecttriggers.html

API call confirms the structure:
```
"webhook": {
                "url": "https://api.github.com/repos/<OMIT>",
                "payloadUrl": "https://codebuild.eu-west-1.amazonaws.com/webhooks?t=<OMIT>",
                "filterGroups": [
                    [
                        {
                            "type": "EVENT",
                            "pattern": "PULL_REQUEST_CREATED, PULL_REQUEST_UPDATED, PULL_REQUEST_REOPENED",
                            "excludeMatchedPattern": false
                        },
                        {
                            "type": "ACTOR_ACCOUNT_ID",
                            "pattern": "ausername",
                            "excludeMatchedPattern": false
                        }
                    ],
                    [
                        {
                            "type": "EVENT",
                            "pattern": "PULL_REQUEST_CREATED, PULL_REQUEST_UPDATED, PULL_REQUEST_REOPENED",
                            "excludeMatchedPattern": false
                        },
                        {
                            "type": "BASE_REF",
                            "pattern": "ref/heads/master$",
                            "excludeMatchedPattern": false
                        },
                        {
                            "type": "HEAD_REF",
                            "pattern": "ref/heads/branch1$",
                            "excludeMatchedPattern": false
                        }
                    ]
                ]
            },
```
Hope I used the class appropriately.

### DIFF
--- a/troposphere/codebuild.py
+++ b/troposphere/codebuild.py
@@ -239,19 +239,26 @@ class ProjectTriggers(AWSProperty):
         'FilterGroups': (list, False)
     }
 
-    def __init__(self, title=None, **kwargs):
-        super(AWSProperty, self).__init__(title, **kwargs)
+    def validate(self):
+        if 'FilterGroups' not in self.properties:
+            raise KeyError('FilterGroups is required when creating triggers')
         if not isinstance(self.FilterGroups, list):
             self._raise_type('FilterGroups', self.FilterGroups, list)
-        if not isinstance(self.FilterGroups[0], list):
-            self._raise_type('FilterGroups[0]', self.FilterGroups[0], list)
-        count = 0
-        for hook in self.FilterGroups[0]:
-            if not isinstance(hook, WebhookFilter):
+        counti = 0
+        for elem in self.properties.get('FilterGroups'):
+            if not isinstance(elem, list):
                 self._raise_type(
-                    'FilterGroups[0][{}]'.format(count), hook, WebhookFilter
+                    'FilterGroups[{}]'.format(counti),
+                    self.FilterGroups[counti], list
                 )
-            count += 1
+            countj = 0
+            for hook in self.FilterGroups[counti]:
+                if not isinstance(hook, WebhookFilter):
+                    self._raise_type(
+                        'FilterGroups[{}][{}]'.format(counti, countj),
+                        hook, WebhookFilter
+                    )
+                countj += 1
 
 
 def validate_status(status):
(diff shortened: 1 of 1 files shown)


JohnPreston:
Force-pushed to squash 2 commits which were to fix pep-8 issues.

markpeek:
Oh interesting. list of lists...likely what you want to do to implement is not to override \_\_init\_\_() but create a `validate()` function instead and a test case to verify it. Something like this (untested):
```
def validate(self):
    if 'FilterGroups' not in self.properties:
        return
    for filter_list in self.properties.get('FilterGroups'):
        if not isinstance(filter_list, list):
            raise ....
        for webhook in filterlist:
            if not isinstance(webhook, WebhookFilter):
                raise ...
````

**Comment by JohnPreston** [comment]
Force-pushed to squash 2 commits which were to fix pep-8 issues.

**Comment by markpeek** [comment]
Thanks! I did see an issue with counti not getting incremented so I switched to using enumeration and added some tests. Please take a look at https://github.com/cloudtools/troposphere/commit/8259f0a44d47b77bf9a2aa37a1c8d0ae7b73091c to verify.

**Comment by phobologic** [comment]
@JohnPreston @markpeek - sorry to revive this, but we just ran into an issue with this.

https://github.com/cloudtools/troposphere/blame/66cea0002c55c3f667ace816ce4c8f457940bea2/troposphere/codebuild.py#L244

Effectively that makes `FilterGroups` required, but 

https://github.com/cloudtools/troposphere/blame/66cea0002c55c3f667ace816ce4c8f457940bea2/troposphere/codebuild.py#L240

has it required as False, as the documentation dictates.  Was this intended?  We can work around it by setting `FilterGroups` to a blank list, I think, but I don't believe this is required (it didn't used to be, and the docs don't list it as required).  Thanks!

**Comment by markpeek** [comment]
Agreed. Looks like ProjectTriggers is not required and Webhook/FilterGroups are not required. Likely that raise line gets changed to a return. Feel free to PR and adding a test is always encouraged. :-)

**Comment by phobologic** [comment]
Cool, I'll get to a PR probably tonight.

**Comment by phobologic** [comment]
Yeah, we'll need a PR - my workaround (an empty list of lists) fails:

```
Failed to call UpdateWebhook, reason: Filter group cannot be empty (Service: AWSCodeBuild; Status Code: 400; Error Code: InvalidInputException; Request ID: a9f097fb-8c9a-11e9-9c26-5d3f240d6956)
```

PR #1376, opened 2019-04-20
Added support for lambda in TargetGroup with additional validation

leewinder:
Added 'lambda' as a valid target type in a TargetGroup, changing the previously required types (which are only required by 'ip' and 'instance' types) to optional.

Added validation to ensure the now-optional properties are present when needed, and not present when the target type is a lambda.

Apologies for the slightly curt formatting, I was struggling to keep some of the lines under 79 characters :)


markpeek:
Thanks!

PR #1381, opened 2019-04-25
Look for any cloudformation variables in userdata

allyunion:
Look for any cloudformation variables in userdata and use Sub around them to force substitution as needed


allyunion:
@markpeek : troposphere had difficulty doing multiple Sub lines for a userdata script.  This was the only reasonable solution that I found works.

markpeek:
@allyunion thank you for the PR. Likely it would be good if you could supply a couple of unit tests to exhibit expected execution and any error cases.

markpeek:
Closing due to no response for feedback.

Issue #1019, opened 2018-04-06
In DMS endpoint, Username is required ....


emayssat-snaplogic:
Hello,

When I look at the source code, I see that the Username in the endpoint definition is set as required.
There is no username in DynamoDB, therefore this requirement seems incorrect.
To bypass the validation issue, I provide a Username but let's see if the stack gets configured correctly.

--
Emmanuel

emayssat-snaplogic:
The code below fixes the issue
`
class Endpoint(AWSObject):
    resource_type = "AWS::DMS::Endpoint"

    props = {
...
        'Username': (basestring, False),
    }
`

markpeek:
I believe what you're implying is when using DMS on a DynamoDB database certain fields are invalid and, in the case of username, not required. However, for other database types username is required. If that is the case then the username should be set to False (not required) and an validation function added which will enforce the username requirement based on the EngineName.

kylemurph:
This also prevents S3 Endpoints from being provisioned

PR #1387, opened 2019-05-06
Changing True to False for Username property in Endpoint class

kylemurph:
Changing behavior to that of Cloudformation: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dms-endpoint.html#cfn-dms-endpoint-username

Given how some Endpoint engine types do not work with the Username property set, this will be more consistent with what's expected.

Related to: https://github.com/cloudtools/troposphere/issues/1019


markpeek:
Thanks!

PR #1388, opened 2019-05-08
Fixed Schedule as basestring on AWS Glue Trigger

Webysther:


markpeek:
According to both the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-glue-trigger.html)  and the spec files, the Schedule property in Glue::Trigger is a basestring. The Schedule property class is used with Glue::Crawler.

Webysther:
My bad.

PR #1390, opened 2019-05-08
make DefinitionString and DefinitionBody mutually exclusive

tylersouthwick:
make DefinitionString and DefinitionBody mutually exclusive, but allow no definition

from
https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessapi

```
Swagger specification that describes your API. If neither DefinitionUri
nor DefinitionBody are specified, SAM will generate a DefinitionBody for
you based on your template configuration.
```


markpeek:
Thanks! +1 on adding the test.

PR #1391, opened 2019-05-08
Fix MethodSettings on AWS::Serverless::Api

tylersouthwick:
According to the SAM documentation (and
https://github.com/awslabs/serverless-application-model/blob/master/samtranslator/model/sam_resources.py):
* MethodSetting -> MethodSettings
* this is an array of MethodSetting, not a single MethodSetting


markpeek:
Thanks!

PR #1392, opened 2019-05-09
Adds Tmpfs prop to LinuxParameters

techman83:
Noted the Tmpfs prop was missing. Interestingly all the props for TmpFs are optional in the ECS docs!

### DIFF
--- a/troposphere/ecs.py
+++ b/troposphere/ecs.py
@@ -188,11 +188,20 @@ class KernelCapabilities(AWSProperty):
     }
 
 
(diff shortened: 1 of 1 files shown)


techman83:
Noted that [technically](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-tmpfs.html) it was `Tmpfs` not `TmpFs`.

markpeek:
Thanks!

Issue #1394, opened 2019-05-10
Appmesh Listener doesn't have a HealthCheckPolicy Property


cigan1:
According to Cloudformation documentation VirtualNode Listener has HealthCheck property while the troposphere code contains HealthCheckPolicy for the same. Documentation source: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-appmesh-virtualnode-listener.html

PR #1395, opened 2019-05-10
Fixed issue #1394 wrong appmesh Listener property and #1396 dependson should be a type list and #1397 proxy props should be list

cigan1:


markpeek:
Thanks! Still wish AWS would publish docs/spec files for this addition.

cigan1:
@markpeek Thank you! I've seen those docs with my own eyes. Our TAM had forwarded an email to me when they release support for CFT. Now I can't find them. I've already asked him to investigate. Will get some answers next week. Thanks.

PR #1398, opened 2019-05-14
Added missing container name property

cigan1:
ContainerName is a required property according to this documentation: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ProxyConfiguration.html


cigan1:
@markpeek sorry for submitting those piecemeal... I am adding them as quickly as I am running into them while deploying...

markpeek:
Thanks! And understood about needing to submit piecemeal. Appreciate you getting it working!

PR #1399, opened 2019-05-15
Update region/az information

michael-k:
`ap-northeast-2` (Seoul) has only 2 AZs, see https://aws.amazon.com/about-aws/global-infrastructure/regions_az/ and `aws ec2 describe-availability-zones --region 'ap-northeast-2'`

### DIFF
--- a/troposphere/constants.py
+++ b/troposphere/constants.py
@@ -2,12 +2,16 @@
 # Regions
 #
 
+AP_EAST_1 = 'ap-east-1'
 AP_NORTHEAST_1 = 'ap-northeast-1'
 AP_NORTHEAST_2 = 'ap-northeast-2'
+AP_NORTHEAST_3 = 'ap-northeast-3'
 AP_SOUTHEAST_1 = 'ap-southeast-1'
 AP_SOUTHEAST_2 = 'ap-southeast-2'
 AP_SOUTH_1 = 'ap-south-1'
 CA_CENTRAL_1 = 'ca-central-1'
+CN_NORTH_1 = 'cn-north-1'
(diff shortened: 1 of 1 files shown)


markpeek:
Thank you for the update. However, that command shows your AZ's but not necessarily all the AZ's in that region. For example:
```
$ aws ec2 describe-availability-zones --region 'ap-northeast-2' | jq .AvailabilityZones[].ZoneName
"ap-northeast-2a"
"ap-northeast-2c"
```
There is likely a better source of this information but I'd suggest also checking [here](https://aws.amazon.com/about-aws/global-infrastructure/regions_az/?p=ngi&loc=2).

michael-k:
I've updated the commit to exclude the removal.

---

Hm, I totally missed that it's `2b` that's missing for me as well.  I noticed that the `ZoneId`s I see are `apne2-az1` and `apne2-az3`, but somehow missed the `ZoneName`s.

I've assumed that if they expose only 2 of 3 AZs to a customer, they use `a` and `b` in the name.  But using `aws ec2 describe-availability-zones --region 'us-west-1'`¹ with several accounts, I've only ever seen `usw1-az1` and `usw1-az3`, but all possible `ZoneName`s.

I found no indication that they use `ap-northeast-2b` for any customer, but since it doesnÄt hurt to have one additional constant, I decided to ignore this.

> There is likely a better source of this information but I'd suggest also checking here.

That's the same link I posted. ;)  But it doesn't say anything about the naming of the AZs.


¹ `us-west-1` (Northern California) started with 3 AZs, but was limited to 2 for new customers.

markpeek:
Thanks!

PR #1401, opened 2019-05-17
Extending Firehose with support for DataFormatConversionConfiguration prop

wieshka:
Came across missing properties for Firehose JSON to Parquet or OCR converting. Props added against boto3 1.9.150 documentation.


wieshka:
Feature is not supported via CloudFormation, closing pull.

PR #1404, opened 2019-05-22
Fix Parameters on AWS::Batch::JobDefinition

lkinley-rythmos:
Parameters is not a required property as per https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-batch-jobdefinition.html#cfn-batch-jobdefinition-parameters


markpeek:
Thanks!

PR #1407, opened 2019-05-24
Adding AWS::Transfer resources, per 2019 May 23 update

axelpavageau:


markpeek:
@axelpavageau I'm not sure we want to include the CloudFormationResourceSpecification.json into the repo since that is maintained externally by AWS. I'd be curious about your thoughts since you included it in this PR and #1408.

axelpavageau:
you're right, it doesn't make sense. I'll remove the file from both PRs

markpeek:
Thanks!

PR #1409, opened 2019-05-29
[Carry #1136] Return real booleans in the output

michael-k:
This picks up the work done by @ikben in PR cloudtools/troposphere#1136, addresses the merge conflict by rebasing and fixes the failing tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@ python:
   - "3.5"
   - "3.6"
   - "3.7.3"
+env:
+  global:
+    - TROPO_REAL_BOOL=true
 
 before_install:
   - pip install --upgrade pip setuptools wheel
--- a/examples/EMR_Cluster.py
+++ b/examples/EMR_Cluster.py
@@ -15,14 +15,14 @@ kms_key = 'arn:aws:kms:us-east-1:123456789012:key/1234-1234-1234-1234-1234'
 
 security_configuration = {
     'EncryptionConfiguration': {
-        'EnableInTransitEncryption': 'true',
+        'EnableInTransitEncryption': True,
         'InTransitEncryptionConfiguration': {
             'TLSCertificateConfiguration': {
                 'CertificateProviderType': 'PEM',
                 'S3Object': 's3://MyConfigStore/artifacts/MyCerts.zip'
             }
         },
-        'EnableAtRestEncryption': 'true',
+        'EnableAtRestEncryption': True,
         'AtRestEncryptionConfiguration': {
             'S3EncryptionConfiguration': {
                 'EncryptionMode': 'SSE-KMS',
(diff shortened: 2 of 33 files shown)


michael-k:
Commit  62fbe39 should not have passed the tests.  I've tried that before: https://travis-ci.com/michael-k/troposphere/builds/113547741
Maybe I've pushed the last commit to fast and travis got confused. `¯\_(ツ)_/¯`

ikben:
Feel free to squash all the commits together, there's nothing ground-breaking in mine :)

markpeek:
Waiting on review/approval from @phobologic

phobologic:
This looks good to me from a code perspective - but I'm curious what we should do to make sure folks aren't freaked out if suddenly a bunch of their stacks start updating?

The other question I have is whether or not json's booleans are actually case sensitive?  This page makes me think they are:

https://json-schema.org/understanding-json-schema/reference/boolean.html

michael-k:
> The other question I have is whether or not json's booleans are actually case sensitive?

Yes, they are.

Python:
```python
import json

json.loads("true")  # True
json.loads("True")  # JSONDecodeError
```

Javascript:
```javascript
JSON.parse("true")  # true
JSON.parse("True")  # SyntaxError
```

phobologic:
Ahh, sorry - my bad on the case sensitive question, I had flipped the cases between the python/resulting templates in my head somehow.  That looks good.

I waffle a little on punishing/scaring folks for not using cfn-lint.  Maybe a better way to roll this out would be to have this be sort of "Feature Flagged" with an environment variable or something, and only turn it on if that variable is set - then start using warnings to let folks know that this change is coming.  Then after a verison or two we can make it the default? What are you thoughts?

michael-k:
Any suggestions regarding the naming of such a env var?

phobologic:
I'm open to just about anything - `OUTPUT_REAL_BOOLEANS` maybe?

Also, don't take my suggestion as what *has* to happen :)  Just wanted to get some conversation started around it.  It seems like a nice compromise though, letting folks who want to use real booleans start using it while not penalizing folks who aren't ready for it.

Also curious around thoughts on throwing warnings to let folks know about the change so that we can encourage people to prepare.  @markpeek @michael-k

markpeek:
Likely prefix for uniqueness: `TROPO_REAL_BOOL`? Is just an env var sufficient or do we need to add a function enablement as well? I do wonder if warnings are too invasive and we should communicate via release notes and the email list.

michael-k:
I've added support for the env var `TROPO_REAL_BOOL`.  As you can see in the [`commits`](https://github.com/cloudtools/troposphere/pull/1409/commits) tab,

- tests are passing in commit 020cb83c683106a7a37f95c8eef814c229a4521e (env var not set = booleans as strings) and
- tests are also passing after adjusting the tests and setting the env var to `"true"` in commit 39c71b5294bfb7e6edefdfa9d9dc108d8cf3f8f7.

Is this good enough or does this need tests for the old behavior as well?

phobologic:
This looks great to me.  Thanks @michael-k for seeing this through!

markpeek:
Thank you @michael-k and @ikben!

PR #1412, opened 2019-06-03
Add support for Rules

nielslaukens:
Rules are a feature of AWS Service Catalogue [1] but is undocumented
for pure CloudFormation. However, it does work in vanilla CloudFormation
as well [2].

This adds minimal support for Rules in templates. It provides a bare
dict, and exports it to JSON/YAML.

[1]: https://docs.aws.amazon.com/servicecatalog/latest/adminguide/reference-template_constraint_rules.html
[2]: https://www.cloudar.be/awsblog/undocumented-feature-using-template-constraint-rules-in-cloudformation/


nielslaukens:
I intend to add additional classes & methods to enhance the usability of Rules; but I wanted to check first what your policy for "undocumented" features is?

markpeek:
Interesting. If it works I'm not opposed to it. I'm going to check with some of my AWS contacts about it first though.

rkz:
I confirm the `Rules` statement does work in CloudFormation (used it). Following up on @nielslaukens' proposal, I guess the usage would be:

```python
template = Template()

# Prepare a rule to lock the region in which this template can run
check_region_rule = {
    "CheckRegion": {
        "Assertions": [
            {
                "Assert": Equals(Ref("AWS::Region"), "eu-west-1"),
                "AssertDescription": "This template is only available in eu-west-1"
            }
        ]
    }
}

# Add it to the template
template.rules.append(check_region_rule)
```

Correct?

That does not seem consistent with the way of doing things in `Template` for two reasons:

1. Developer needs to hack into the `.rules` property using `append()`. Would make sense to have a `Template.add_rule()` method instead.
2. The format of the rule itself could be enforced by creating `Rule` and `Assertion` classes, instead of using plain dicts.

Happy to implement an improvement for issue 1, but I think issue 2 could be addressed later depending on the roadmap.

What do you think @markpeek ? Happy to implement that if you're ready to merge soon after.

nielslaukens:
> ```
> # Add it to the template
> template.rules.append(check_region_rule)
> ```
`template.rules` is a `dict`, so you would need to call `.update()` instead of `.append()`. But, yes, this is current usage. 

> That does not seem consistent with the way of doing things in `Template` for two reasons:
> 
> 1. Developer needs to hack into the `.rules` property using `append()`. Would make sense to have a `Template.add_rule()` method instead.

Yes, you are correct. As I mentioned in my original PR: "This adds minimal support for Rules in templates. It provides a bare dict, and exports it to JSON/YAML." This needs more work to provide better usability. I wanted to find out if this would be acceptable before putting in the extra effort.

markpeek:
I checked with a contact on the CloudFormation team. Rules are supported for Service Catalog and not CloudFormation. Having heard that, I will accept this PR (when it's good to go) to implement it for CloudFormation but obviously can't make promises about AWS continuing to allow its use.

I would suggest people wanting this as a supported feature to make it known to AWS through their forums, support channels, and/or your sales reps.

rkz:
Thanks for checking with AWS. I also verified on my side, uploading the following template to CloudFormation:

```json
{
    "Description": "Testing the 'Rules' feature at the template level",
    "Parameters": {
        "Parameter1": {
            "Type": "String"
        },
        "Parameter2": {
            "Type": "String"
        }
    },
    "Resources": {
        "MyInstance": {
            "Properties": {
                "ImageId": "ami-01b282b0f06ba5fd2",
                "InstanceType": "t3.nano"
            },
            "Type": "AWS::EC2::Instance"
        }
    },
    "Rules": {
        "DemoRule": {
            "Assertions": [
                {
                    "Assert": {
                        "Fn::Equals": [
                            {
                                "Ref": "Parameter1"
                            },
                            {
                                "Ref": "Parameter2"
                            }
                        ]
                    },
                    "AssertDescription": "Parameters must match"
                }
            ]
        }
    }
}
```

It does bring validators to the UI, checking that `Parameter1` and `Parameter2` are equal. The Rules feature for CloudFormation is [working but not documented](https://www.cloudar.be/awsblog/undocumented-feature-using-template-constraint-rules-in-cloudformation/).

I'll prepare the PR with docstrings mentioning the feature is undocumented.

rkz:
Final PR ready: https://github.com/cloudtools/troposphere/pull/1446

Thanks @nielslaukens and @markpeek for your reactivity!

markpeek:
Thanks @nielslaukens and @rkz!

PR #1417, opened 2019-06-05
List of tags vs singular tag.

dovy:
Randomly found I can't add a list of tags via troposphere. Unless you know something I don't.  ;)


markpeek:
@dovy you can pass a dict of tag pairs into it.

PR #1422, opened 2019-06-12
Sagemaker coderepository

kayengar:
A new resource type under SageMaker called AWS::SageMaker::CodeRepository was added. Reference: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-sagemaker-coderepository.html


markpeek:
Thanks!

PR #1425, opened 2019-06-13
Change add_description to set_description in all examples

jake-billings:
I'm new to this library, but I noticed I got a deprecation warning when I copied the CloudFront example


markpeek:
Nice! Thanks!

PR #1426, opened 2019-06-17
Add support for httpHeaderConfig

eloymg:
Add de HttpHeaderConfig support for ALB ListenerRule


markpeek:
Thanks.

Note: this change use "str" as the type instead of "basestring" and there were also other Condition configs needed. I'm going to fix these up with a subsequent commit.

PR #1433, opened 2019-06-18
Add transit gateway ID to Route

gsoyka:
Per docs here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-route.html


markpeek:
Thanks! Likely a documentation issue that TransitGatewayId isn't listed as mutually exclusive.

Issue #1333, opened 2019-02-25
dlm interval issue


perigee:
Current on version `2.4.2`,

The DLM  (data lifecycle manager) interval values aer 2,3,4,6,8,12,24, while in code https://github.com/cloudtools/troposphere/blob/master/troposphere/dlm.py#L11, could you please fix this issue, thank you in advance.

markpeek:
This [interval](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-dlm-lifecyclepolicy-createrule.html#cfn-dlm-lifecyclepolicy-createrule-interval) documentation only shows 12 and 24. Could you link documentation for that expanded range of values?

perigee:
sorry for missing the doc, please reference here https://docs.aws.amazon.com/dlm/latest/APIReference/API_CreateRule.html, also it is confirmed on AWS UI, since I can modify the interval value in UI.

PR #1437, opened 2019-06-19
Update DLM rule interval values (#1333)

flou:
See updated documentation https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-dlm-lifecyclepolicy-createrule.html#cfn-dlm-lifecyclepolicy-createrule-interval


markpeek:
Thanks!

PR #1440, opened 2019-06-21
Remove EC2 creation warning

Bheesma:


markpeek:
This example was a direct conversion from an AWS CloudFormation template. Can you let me know why that warning should be removed? Also, the CI check is failing since the example output is also stored down in the tests subdirectory and compared during the tests.

markpeek:
Closing due to no response for feedback.

PR #1441, opened 2019-06-26
Add Tags property to IAMRole definition

jpvowen:
Add the "Tags" property to troposphere's IAM Role definition.
It looks like CloudFormation has started incorporating tags for IAM roles -- the CloudFormation resource specifications for all regions now include a Tags property in the AWS::IAM::Role spec, and stack tags are being applied to generated roles.


markpeek:
Thanks!

PR #1444, opened 2019-06-28
Add Tags to ECR Repository definition

patrickdean:
ECR added support for tags back in December:
https://aws.amazon.com/about-aws/whats-new/2018/12/amazon-ecr-now-allows-repository-tagging/

Let me know if anything looks out of place. Thanks!

### DIFF
--- a/troposphere/ecr.py
+++ b/troposphere/ecr.py
@@ -1,4 +1,4 @@
-from . import AWSObject, AWSProperty
+from . import AWSObject, AWSProperty, Tags
 try:
     from awacs.aws import Policy
     policytypes = (dict, Policy)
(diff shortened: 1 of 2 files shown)


troposphere/ecr.py
```diff
@@ -20,4 +20,5 @@ class Repository(AWSObject):
         'LifecyclePolicy': (LifecyclePolicy, False),
         'RepositoryName': (basestring, False),
         'RepositoryPolicyText': (policytypes, False),
+        'Tags': ((Tags, list), False),
```
markpeek: When adding new Tags there is a preference not to include 'list' which is being deprecated. Can you change this to just be 'Tags'? Likely need to tweak the test as well.

patrickdean: Updated!

markpeek:
Thanks!

PR #1446, opened 2019-06-28
Add support for Rules

rkz:
Rules are a feature of AWS Service Catalogue 1 but is undocumented for pure CloudFormation. However, it does work in vanilla CloudFormation as well 2.

This adds minimal support for Rules in templates with a `Template.add_rule()` function.

A future improvement could bring a `Rule` class to validate the contents of the rule itself.

Originally contributed by @nielslaukens at https://github.com/cloudtools/troposphere/pull/1412


markpeek:
@nielslaukens since this was based on your original PR, can you review and give your +1?

nielslaukens:
I would add a check for duplicate names, [like it's done for `.add_resource()`](https://github.com/cloudtools/troposphere/blob/master/troposphere/__init__.py#L663-L667). Since we're currently don't have a `BaseAWSObject` for a Rule, I've just copied the duplicate checking part.

I would also add a check on the maximum number of rules. I've searched for a documented maximum number of rules, but couldn't find one in the documentation. If I find the time, I'll try to figure out the current actual limit and configure that. Or put in a TODO for later...

I've pulled in your changes, and added both of the above in my branch (#1412)

markpeek:
Merged via the updated #1412.

PR #1448, opened 2019-07-02
Fix tags type for AWS Backup resources

flou:
In the current implementation, the properties RecoveryPointTags and BackupVaultTags are defined as json_checker, which returns a string whereas CloudFormation expects a dict
`BackupVaultTags property should be a key value pair and not a string`

The example template in `tests/examples_output/Backup.template` is actually deploys a BackupPlan and Vault.


markpeek:
Thanks!

PR #1460, opened 2019-07-18
Fix TemporaryPasswordValidityDays type

djmeph:
This fixes an invalid type requirement on the 'TemporaryPasswordValidityDays' parameter of the Cognito UserPool Password Policy Object. Setting this to 'float' will invalidate the field if an integer is used, and a float value causes the stack to fail in Cloudformation with an ambiguous error.


markpeek:
Thanks!

Issue #1466, opened 2019-07-24
ASK definition is incorrect


mansdahlstrom1:
This file https://github.com/cloudtools/troposphere/blob/master/troposphere/ask.py does not align with the cloudformation documentation found here:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ask-skill.html#cfn-ask-skill-vendorid

mansdahlstrom1:
This can be closed now that https://github.com/cloudtools/troposphere/pull/1467 is merged

markpeek:
Thanks for corrections.

PR #1467, opened 2019-07-24
Fix ASK.py

mansdahlstrom1:
Fixes this issue: https://github.com/cloudtools/troposphere/issues/1466


mansdahlstrom1:
I used a forked repo from last year so the commit history is a bit weird. You if use squash-merge if should be fine thou. The commits there are already in master, so no diff in terms of the files.

markpeek:
Thanks!

PR #1470, opened 2019-07-26
Travis CI: Add flake8 which is a superset of pycodestyle and pyflakes

cclauss:
[__flake8__](http://flake8.pycqa.org/) runs all the tests that pycodestyle and pyflakes do but also adds the Fxxx errors.

__basestring__ was removed in Python 3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,6 @@ install:
 before_script:
   - if [[ $TRAVIS_PYTHON_VERSION == 3* ]]; then 2to3 -n -w --no-diffs troposphere scripts/cfn2py scripts/gen.py examples/CustomResource.py tests; fi
   - flake8 --version
-  - pycodestyle --show-source --show-pep8 .
-  - pyflakes .
   - flake8 . --show-source 
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,14 +13,14 @@ before_install:
   - pip install --upgrade pip setuptools wheel
 
 install:
-  - pip install --upgrade pycodestyle
-  - pip install --upgrade pyflakes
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,14 +13,14 @@ before_install:
...
-  - pip install --upgrade pyflakes
+  - pip install --upgrade flake8
 
 before_script:
   - if [[ $TRAVIS_PYTHON_VERSION == 3* ]]; then 2to3 -n -w --no-diffs troposphere scripts/cfn2py scripts/gen.py examples/CustomResource.py tests; fi
-  - pycodestyle --version
+  - flake8 --version
   - pycodestyle --show-source --show-pep8 .
   - pyflakes .
+  - flake8 . --show-source
```
michael-k: The executions of `pycodestyle` and `pyflakes` can be removed, right?

cclauss: Yes.

markpeek:
I've been hesitant about accepting this PR since there are times when I've needed to "pin" a version of pycodestyle due to new rules or other breakage (5548d43addc20ae47328c3cb11a071fb1fa50b10). Is there a similar way to map pycodestyle versions to flake8 versions in the case of needing a rollback?

michael-k:
`pip install --upgrade flake8 "pycodestyle==2.5.0"` is still possible

markpeek:
Thank you for the PR. I'll be pushing a change to Makefile to run flake8 via "make test" to ensure consistency.

cclauss:
I like __make lint__ for things like flake8 and then __make test__ for things like pytest, unittests, doctests, etc.

Issue #1471, opened 2019-07-29
Kinesis Analytics CodeContent property is incorrectly defined


alexeyts:
When you use `CodeContent` property to define a Kinesis Analytics application you get an error when exporting the CloudFormation, because the property is defined incorrectly.

markpeek:
Fixed via merge.

PR #1472, opened 2019-07-29
Issue #1471: fix bug in CodeContent object

alexeyts:
A fix for the issue: https://github.com/cloudtools/troposphere/issues/1471


alexeyts:
@markpeek should be a pretty straightforward change

markpeek:
Ah yes, thanks for the reminder. I saw it and thought about adding a test to validate all of those fields...but then got distracted. Thanks for the change.

PR #1480, opened 2019-08-19
Fixed a typo in the ClientBroker's value

srinivasreddych:
This typo will make CFN error out about invalid tags which is misleading.


markpeek:
Thanks! I'm going to fix up the example output to make the tests pass.

Issue #1481, opened 2019-08-20
Can't tag Glue Crawlers 


seanvedrinellecirc:
Glue.Crawler requires Tags to be entered as type dict instead of type Tag.
Similar issue to https://github.com/cloudtools/troposphere/issues/1478 I suspect.

markpeek:
Thanks!

PR #1482, opened 2019-08-20
updated glue crawler tag attribute to match aws cloudformation doc

seanvedrinellecirc:
As described in this [issue](https://github.com/cloudtools/troposphere/issues/1481) there is an attribute type mismatch that this patch resolves.

### DIFF
--- a/troposphere/glue.py
+++ b/troposphere/glue.py
@@ -163,7 +163,7 @@ class Crawler(AWSObject):
         'Schedule': (Schedule, False),
         'SchemaChangePolicy': (SchemaChangePolicy, False),
         'TablePrefix': (basestring, False),
-        'Tags': (Tags, False),
+        'Tags': (dict, False),
         'Targets': (Targets, True),
     }
 


peteroruba:
This issue impacts my work. Would love to see this merged.

markpeek:
Before I merge this change, can you show me what a working Tags looks like in the context of Glue Crawler? Outside of troposphere, how is the json/yaml different from normal Tags?

seanvedrinellecirc:
Similar situation to https://github.com/cloudtools/troposphere/issues/1045

What troposphere produces (Normal tags):

```
{ 
    "Resources": {
        "GenericCrawler": {
            "Properties": {
                "DatabaseName": "glueDatabase",
                "Role": "arn:aws:iam::12345:role/GlueRole",
                "Tags": [
                    {
                        "Key": "Environment",
                        "Value": "data"
                    },
                    {
                        "Key": "Project",
                        "Value": "tropospherefix"
                    }
                ],
                "Targets": {
                    "S3Targets": [
                        {
                            "Path": "myfancybucket"
                        }
                    ]
                }
            },
            "Type": "AWS::Glue::Crawler"
        }
    }
}
```

Give us the following error: "Property validation failure: [Value of property {/Tags} does not match type {Map}]"

What tropophere should produce:

```
{
    "Resources": {
        "GenericCrawler": {
            "Properties": {
                "DatabaseName": "glueDatabase",
                "Role": "arn:aws:iam::12345:role/GlueRole",
                "Tags": {"Environment" : "data", "Project" : "tropospherefix"},
                "Targets": {
                    "S3Targets": [
                        {
                            "Path": "myfancybucket"
                        }
                    ]
                }
            },
            "Type": "AWS::Glue::Crawler"
        }
    }
}

```

markpeek:
Thank you for the change and the details. Looks like there are Tags in other resources so I'll be fixing those up as well.

seanvedrinellecirc:
Grand, thanks!

PR #1489, opened 2019-09-01
Add threshold metric

ScOut3R:
Add _ThresholdMetricId_ for AWS::CloudWatch::Alarm.

Also according to the CloudFormation specification _Threshold_ is not mandatory.


markpeek:
Thanks!

Issue #1477, opened 2019-08-08
Errors in ec2.BlockDeviceMapping


catcher2383:
I'm trying to remove a device in a lauchtemplate and I noticed these issues :
- "DeviceName" is not mendatory.
- "NoDevice" prop should be a str instead of a dict.

[The AWS Doc](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-blockdevicemapping.html)

PR #1490, opened 2019-09-02
Fix issue  #1477 - Errors in ec2.BlockDeviceMapping

vissree:


markpeek:
Thanks for this PR. The real issue is due to different definitions of BlockDeviceMappings being used. This issue has been resolved via 02032100ec1f364541de06f8ca50ba64c7bdb7a5 which separates out the different BlockDeviceMappings in EC2.

PR #1491, opened 2019-09-05
Fix naming of parameters in FindInMap helper.

Milesjpool:
Parameters should be two-levels of nesting, rather than <key, value>. See link for reference:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-findinmap.html


markpeek:
Thanks!

PR #1493, opened 2019-09-11
Add missing EnableNonSecurity property to Rule

mattjbarlow:
Adds `EnableNonSecurity` which is a non-required property for [Rule](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ssm-patchbaseline-rule.html)


markpeek:
Thanks!

PR #1496, opened 2019-09-25
Add new Properties to SSM::Parameter

ikben:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-parameter.html


markpeek:
Thanks!

PR #1497, opened 2019-09-27
Add Description field to IAM Role

0rax:
IAM Role now support adding a description to them from a CloudFormation template (see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html#cfn-iam-role-description), this make it possible to do so in troposphere.


markpeek:
Thanks!

PR #1500, opened 2019-10-01
added missing CrawlerName field to Glue Action object

timothygrahamm:


troposphere/glue.py
```diff
@@ -481,6 +481,7 @@ class Action(AWSProperty):
     props = {
         'Arguments': (dict, False),
         'JobName': (basestring, False),
+        'CrawlerName': (basestring, False),
```
markpeek: Can you alphabetize and push an update? Thanks!

timothygrahamm: Done :)

timothygrahamm: I also came across some missing properties in the Condition, so I threw them in too.

markpeek:
Thanks!

PR #1501, opened 2019-10-01
Update placement object

cwgem:
The placement object is missing a few fields from the upstream
CloudFormation specification:

* Affinity
* HostId
* Tenancy

This commit adds them to the Placement object

Current Placement object specification: https://github.com/cwgem/troposphere/pull/new/updated_launch_template_properties


markpeek:
Thanks!

PR #1502, opened 2019-10-02
Allow Tags to be given as a list when creating Resource that allows Tags

JmTravelex:
Added: Allows to pass in a CFN format list of Tags and transform it into a Tags() object on instantiation.

Added the exception for basestring as it seemed to be yelling in my import :thinking:

Example of code doing it:
```
QueueA:
  Properties:
    Tags:
      - Key: Name
        Value: abcd
      - Key: VPC
        Value: vpc-id
```
```
#!/usr/bin/env python
try:
    basestring
except NameError:
    basestring = str


from troposphere.sqs import Queue
with  open('test.yml', 'r') as fd:
    content = fd.read()
import yaml
content_dict = yaml.load(content, Loader=yaml.Loader)
queue = Queue('queuea', **content_dict['QueueA']['Properties'])
print(queue.to_dict())
```

If there is a better place to do it, or a nicer way to import a native CFN format into creating the object in full, I would love to see how :)

Many thanks for your consideration. Keep up the awesome work!

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -39,6 +39,11 @@ PARAMETER_TITLE_MAX = 255
 
 valid_names = re.compile(r'^[a-zA-Z0-9]+$')
 
+try:
+    basestring
+except NameError:
+    basestring = str
+
 
 def is_aws_object_subclass(cls):
     is_aws_object = False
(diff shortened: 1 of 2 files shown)


singh1469:
+1

JohnPreston:
No idea why the py2.7 tests fail :/

JmTravelex:
Opened under the wrong user :facepalm:

PR #1503, opened 2019-10-02
Allow Tags to be given as a list when creating Resource that allows Tags

JohnPreston:
Added: Allows to pass in a CFN format list of Tags and transform it into a Tags() object on instantiation.

Added the exception for basestring as it seemed to be yelling in my import thinking

Example of code doing it:

```QueueA:
  Properties:
    Tags:
      - Key: Name
        Value: abcd
      - Key: VPC
        Value: vpc-id
```
```#!/usr/bin/env python
try:
    basestring
except NameError:
    basestring = str


from troposphere.sqs import Queue
with  open('test.yml', 'r') as fd:
    content = fd.read()
import yaml
content_dict = yaml.load(content, Loader=yaml.Loader)
queue = Queue('queuea', **content_dict['QueueA']['Properties'])
print(queue.to_dict())
```
If there is a better place to do it, or a nicer way to import a native CFN format into creating the object in full, I would love to see how :)

Many thanks for your consideration. Keep up the awesome work!
(Reopened from #1502 which was opened with the wrong account :/)

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -39,6 +39,11 @@ PARAMETER_TITLE_MAX = 255
 
 valid_names = re.compile(r'^[a-zA-Z0-9]+$')
 
+try:
+    basestring
+except NameError:
+    basestring = str
+
 
 def is_aws_object_subclass(cls):
     is_aws_object = False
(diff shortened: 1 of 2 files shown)


markpeek:
Thank you for this PR. Not sure about accepting it yet but let me give some feedback:

- There are other forms of Tags() (some newer services use JSON tags) so this is not a generalized form.
- There is a [template generator](https://github.com/cloudtools/troposphere/blob/master/troposphere/template_generator.py) which will read in existing templates and convert which was part of your question. I'm guessing this is what your intended use case is for this PR.
- Likely the Tags portion could be a simple function to parse the string and return a Tags() object. Then just `Tags=TagConvert(yaml_string)` in your object creation.
- The inclusion of basestring isn't needed. The code is written in Python 2 and converted via 2to3 for Python 3 which converts the basestrings. With the EOL of Python 2 I'm intending to switch the code base over to Python 3.

JohnPreston:
Just realized this was still opened. Good to close as this makes perfect sense @markpeek

Issue #1506, opened 2019-10-09
Cognito missing EnabledMfas param


castaway2000:
As within in the docs a parameter called EnabledMfas is present alongside MfaConfiguration in the UserPool CFt object. 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpool.html#cfn-cognito-userpool-enabledmfas

EnabledMfas is missing in the cognito.py file for troposphere UserPool object.
https://github.com/cloudtools/troposphere/blob/f590a86b5fdac3a6becf87ae94b51f3a08bbc9f8/troposphere/cognito.py#L208

PR #1507, opened 2019-10-09
adding EnabledMfas to cognito UserPool Object.

castaway2000:
related to this https://github.com/cloudtools/troposphere/issues/1506


troposphere/cognito.py
```diff
@@ -205,6 +205,7 @@ class UserPool(AWSObject):
         'EmailVerificationMessage': (basestring, False),
         'EmailVerificationSubject': (basestring, False),
         'LambdaConfig': (LambdaConfig, False),
+        'EnabledMfas': (basestring, False),
```
markpeek: Minor fixup, please move this line up one to alphabetize the properties. Thanks!

castaway2000: updated. thanks.

markpeek:
Thanks!

markpeek:
@castaway2000 I looked at #1508 and the new resources available. I was going to add them via my generator script. However, when looking at EnabledMfas, I missed that it is a list of strings. Can you confirm your usage of it and then I can correct with my other commit.

castaway2000:
@markpeek my implementation is direct from the CFT docs where it is a list of strings. 

Userpool doc: 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpool.html

direct doc:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpool.html#cfn-cognito-userpool-enabledmfas

i am building with the latest revision of troposphere but edited and ran the CFT confirming the current implementation to be working. then copied the code over directly for this PR. 

let me know if you need any additional info.

markpeek:
OK, then instead of this `'EnabledMfas': (basestring, False),` it should be implemented `'EnabledMfas': ([basestring], False),`. Note the brackets that imply a list of strings.

castaway2000:
@markpeek as soon as I realized it i made a PR for it. https://github.com/cloudtools/troposphere/pull/1511

PR #1509, opened 2019-10-10
cognito is missing UserPoolResourceServer

castaway2000:
cognito.py is missing some critical components that are in the CFT docs but not in cognito. this PR is to add in pool resource servers.


markpeek:
Thanks!

PR #1511, opened 2019-10-14
uncaught syntax error on EnabledMfa's

castaway2000:
missed that EnabledMfa's needs to be a list of strings.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpool.html#cfn-cognito-userpool-enabledmfas


castaway2000:
@markpeek thanks for the questions you asked on my last PR it made me realize i missed the list of strings. it was uncaught because of programmer error.  here is the correct implementation.

markpeek:
Thanks!

PR #1514, opened 2019-10-25
Added missing rds scaling configuration capacity

davidgourde:
[The documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-dbcluster-scalingconfiguration.html) states that the value "1" is a valid value, which was missing.

> Valid capacity values are 1, 2, 4, 8, 16, 32, 64, 128, and 256


markpeek:
Thanks!

PR #1517, opened 2019-11-04
Adding missing property for FindingPublishing

oliver-bowman:
Add property listed here in AWS documentation. : 

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-guardduty-detector.html#cfn-guardduty-detector-findingpublishingfrequency

### DIFF
--- a/troposphere/guardduty.py
+++ b/troposphere/guardduty.py
@@ -12,6 +12,7 @@ class Detector(AWSObject):
 
     props = {
         'Enable': (boolean, True),
+        'FindingPublishingFrequency': ([basestring], False),
     }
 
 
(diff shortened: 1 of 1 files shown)


troposphere/guardduty.py
```diff
@@ -12,6 +12,7 @@ class Detector(AWSObject):
 
     props = {
         'Enable': (boolean, True),
+        'FindingPublishingFrequency': ([basestring], False),
```
markpeek: Looking at the docs, this likely should be `basestring` (string) instead of `[basestring]` (list of strings). Could you confirm?

oliver-bowman: Yes, you are correct, it was a mistake on my part. I have corrected that now.

markpeek:
Thanks!

PR #1521, opened 2019-11-14
Support for API Gateway SecurityPolicy

sulich:
Added in SecurityPolicy as per:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-domainname.html


markpeek:
Thanks!

PR #1524, opened 2019-11-19
Add WAF Action Type validator

marinpurgar:
Adding the WAS Action Type validator for WAF and WAFRegional.


markpeek:
Thanks! Appreciate adding the test!

PR #1533, opened 2019-12-04
Add ProvisionedConcurrencyConfiguration for Lambda alias and version

oussemos:
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-alias.html

- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-version.html


markpeek:
Thanks!

PR #1536, opened 2019-12-11
Add CloudWatch Alarm TreatMissingData validator

marinpurgar:
Add the CloudWatch Alarm ThreatMissindData Property validator.
Add the Unit Test for validate_threat_missing_data.
Add Unit Test for validate_units.


troposphere/cloudwatch.py
```diff
@@ -26,6 +29,15 @@ def validate_unit(unit):
     return unit
 
 
+def validate_treat_missing_data(value):
+    """Validate TreatMissingData"""
+
+    if value not in VALID_TREAT_MISSING_DATA_TYPES:
+        raise ValueError("Alarm TreatMissingData must be one of: %s" %
+                         ", ".join(VALID_TREAT_MISSING_DATA_TYPES))
+        return value
```
markpeek: This is likely an incorrect indentation of the return. I like that you added tests but for the positive test cases you should validate the return value matches the unit passed in.

markpeek:
Thanks!

PR #1537, opened 2019-12-11
Adding update policy that allows for in place upgrade of ES cluster

danielxoi:
As per https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-upgradeelasticsearchdomain

This should allow for in place version upgrade of Elasticsearch clusters


markpeek:
Thanks!

PR #1549, opened 2019-12-31
Added Serverless::Application and Serverless ApplicationLocation

dovy:
I found these were missing, so I implemented them. I hope they find their way into the core so I can stop using a shim.  ;)

### DIFF
--- a/troposphere/serverless.py
+++ b/troposphere/serverless.py
@@ -39,7 +39,7 @@ class DeadLetterQueue(AWSProperty):
     def validate(self):
         valid_types = ['SQS', 'SNS']
         if ('Type' in self.properties and
-            self.properties['Type'] not in valid_types):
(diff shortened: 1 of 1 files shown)


dovy:
*sigh* Not sure what you want to do here. Actually this is SAML macros being thrown into CloudFormation and they have no docs to show how to do it the vanilla CloudFormation way.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-secretsmanager-rotationschedule.html

But unless you append the transform structure https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/transform-section-structure.html  to your template, it will fail. Gah.

markpeek:
Thanks! Sorry for the delay. Also, I'll be fixing up a couple of things like added blank lines and removing the validate.

PR #1553, opened 2020-01-06
Break out NoDevice property validation (Fixes #1551)

### LINKED ISSUE #1551: LaunchTemplateData BlockDeviceMappings requires different data type for NoDevice than EC2 instance
The BlockDeviceMapping object for an EC2 instance requires that `NoDevice` resolve to a NoDevice object, or a simple dictionary in terms of object validation:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-blockdev-mapping.html

However, in the case of LaunchTemplateData for launch templates, BlockDeviceMapping requires that `NoDevice` resolve to a string type:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-blockdevicemapping.html

The current setup works for EC2 instances but not Launch Templates, due to both mapping to the same object.

cwgem:
* NoDevice is a string for launch template block device mapping and a null object for EC2 block device mapping
* This addresses the issue by breaking out block device mapping to two separate objects. The EC2 version will keep the original name for backwards compatibility purposes and the launch template version will take up a new name

### DIFF
--- a/troposphere/ec2.py
+++ b/troposphere/ec2.py
@@ -138,8 +138,17 @@ class BlockDeviceMapping(AWSProperty):
     props = {
         'DeviceName': (basestring, True),
         'Ebs': (EBSBlockDevice, False),  # Conditional
-        'NoDevice': (dict, False),
         'VirtualName': (basestring, False),  # Conditional
+        'NoDevice': (dict, False)
+    }
+
+
+class LaunchTemplateBlockDeviceMapping(AWSProperty):
+    props = {
+        'DeviceName': (basestring, True),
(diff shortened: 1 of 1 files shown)


cwgem:
Working on fixing the testing issues

cwgem:
The testing issue have been addressed The original solution was using a shared dictionary between two classes, but after consideration of the script which auto generations object skeletons I went with the more verbose method which duplicates the shared properties between the two device block mapping objects.

markpeek:
Thanks! Agree with you on implementation.

PR #1560, opened 2020-01-17
Strategy in AWS::EC2::PlacementGroup is not required

kichik:
Both the [documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-placementgroup.html) and the spec say it's not required. `gen.py` also outputs it as not required.


markpeek:
Thanks!

PR #1575, opened 2020-01-31
NamespaceId is not strictly required

JohnPreston:
NamespaceId is not strictly required. And NamespaceId is only required once, either there, or in the Service object.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-servicediscovery-service-dnsconfig.html

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-servicediscovery-service.html#cfn-servicediscovery-service-namespaceid


markpeek:
Thanks!

JohnPreston:
Note for pending next release: Setting Ref('AWS::NoValue') works around the problem ;)

Issue #1574, opened 2020-01-30
Fix EKS NodeGroup Parameter Definitions


wesleywh:
When defining a NodeGroup (EKS) these two parameters need to be updated:

`Labels` and `Tags`

In `eks.py` it should be modified to fit:
```
class Nodegroup(AWSObject):
    resource_type = "AWS::EKS::Nodegroup"

    props = {
        'AmiType': (basestring, False),
        'ClusterName': (basestring, True),
        'DiskSize': (double, False),
        'ForceUpdateEnabled': (boolean, False),
        'InstanceTypes': ([basestring], False),
        'Labels': (dict, False),
        'NodegroupName': (basestring, False),
        'NodeRole': (basestring, True),
        'ReleaseVersion': (basestring, False),
        'RemoteAccess': (RemoteAccess, False),
        'ScalingConfig': (ScalingConfig, False),
        'Subnets': ([basestring], False),
        'Tags': (dict, False),
        'Version': (basestring, False),
    }
```

The change is:
Labels = basestring -> dict
Tags = Tags -> dict

I was unable to deploy with the previous settings because I kept getting errors from cloudformation. After the change I was able to deploy correctly.

Here is an example of the deploy:
```
template.add_resource(Nodegroup(
        "NodeGroup",
        AmiType=Ref("AmiType"),
        ...
        ...
        Labels=json.loads(sceptre_user_data.get("Labels")) or Ref("AWS::NoValue"),
        Tags={ 
           "MyKey": "MyValue",
           "MySecondKey": "MySecond Value"
        }
))
```

PatMyron:
```
          "PrimitiveType": "Json",
```
for `AWS::EKS::Nodegroup.Labels` and `AWS::EKS::Nodegroup.Tags` in the [Resource Specification](https://d1uauaxba7bl26.cloudfront.net/latest/gzip/CloudFormationResourceSpecification.json)

---

https://github.com/cloudtools/troposphere/blob/74a59ccafa37c7933379d1e9e3da4206e149a731/troposphere/eks.py#L66-L84

---

`Labels` seems like it needs to change type, but I'm not sure how the `Tags` type works @markpeek

markpeek:
[...]

PR #1576, opened 2020-02-03
Nodegroup tags type

dpoirier:
Fix a couple of types in the NodeGroup, per https://github.com/cloudtools/troposphere/issues/1574


markpeek:
Thanks!

PR #446, opened 2016-04-03
Ignore If expression during validation of ASG

ubergeek42:
Fixes #445.

`If` should be a valid expression for MaxCount on an autoscale group(In addition to the already existing `FindInMap` and `Ref` expressions).

PR #1577, opened 2020-02-04
Ignore If expression in validation of AutoScalingRollingUpdate min instances

effoT:
Ignore If expression in validation of AutoScalingRollingUpdate min instances

Same thing as in https://github.com/cloudtools/troposphere/pull/446, but only for min instances instead of max instances.


markpeek:
Thanks!

PR #1585, opened 2020-02-17
adding new Transfer properties, per 2020 Feb 13 update

axelpavageau:


otanner:
Thanks for updating these. Should `VpcEndpointId` and `VpcId` be not required though as they are valid only if the `EndpointType` is set to a certain value and cannot both be set at the same time?

axelpavageau:
you're right, that was a mistake. Thanks for catching it !

markpeek:
Thanks! And shoutout to @otanner for the review!

PR #1595, opened 2020-02-20
fixing property according to the documentation's example

### LINKED ISSUE #1592: wafv2.py IPSet - Error with Addresses property
First of all, thanks to everyone who maintains this repo - Troposphere is a Great tool!  

In the master branch, I came across a problem with wafv2.py when creating an IPSet.  It generates the Addresses property with the IPAddresses property which includes a list of CIDRs; however, it should produce only the list of CIDRs as values of Addresses (without IPAddresses).

Current (incorrect):
        Addresses:
          IPAddresses:
          - 1.2.3.0/24
          - 2.3.0.0/16

Correct:
        Addresses:
          - 1.2.3.0/24
          - 2.3.0.0/16

axelpavageau:
fixes #1592

### DIFF
--- a/troposphere/wafv2.py
+++ b/troposphere/wafv2.py
@@ -410,17 +410,11 @@ class WebACL(AWSObject):
     }
 
 
-class IPAddresses(AWSProperty):
-    props = {
-        'IPAddresses': ([basestring], False)
-    }
-
-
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks!

PR #1596, opened 2020-02-21
adding UsernameConfiguration prop to UserPool, per Feb 20 2020 update

axelpavageau:


markpeek:
Thanks! I needed to merge due to previous changes so verifying would be good.

Issue #1598, opened 2020-02-24
wafv2.py - Issue with WebACL Rules attribute and ManagedRuleGroupStatement


johnsarg:
1) When using the WebACL, the only way I could get it to work was by changing the 'Rules' attribute from '(Rules, False)' to '([Rule], False)'.

2) When adding a rule to WebACL.Rules[], and adding a statement of type ManagedRuleGroupStatement, Cloudformation seems to require an OverrideAction; though, an Action is optional.  I attempted to assign an OverrideAction of None=wafv2.NoneAction(), but without success.  I modified the 'None' prop of OverrideAction to 'none', and Cloudformation was happy.  Is there a better way to handle a prop which has the same name as a Python reserved word?   Thanks for your help.  JS

**Modified wafv2.py**
class OverrideAction(AWSProperty):
    props = {
        'Count': (CountAction, False),
        'none': (NoneAction, False),
    }

**Example**
                    OverrideAction=wafv2.OverrideAction(
                        none=wafv2.NoneAction(),
                    ),

axelpavageau:
Hi @johnsarg 

1) the current implementation follows the spec https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-wafv2-webacl.html#cfn-wafv2-webacl-rules, but the example contradicts it, and matches your finding https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-wafv2-webacl.html#aws-resource-wafv2-webacl--examples .
I'm pushing a PR changing troposphere's implementation. Can you open a bug with cloudformation ?

axelpavageau:
regarding 2), I see that we have an issue as [AWS::WAFv2::WebACL Rule](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-wafv2-webacl-rule.html) and [AWS::WAFv2::RuleGroup Rule](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-wafv2-rulegroup-rule.html) have different properties. However currently they're the same troposphere object.
I'll update the PR tomorrow to create different objects. Unfortunately we'll have to diverge from cloudformation's naming.

johnsarg:
@axelpavageau - Yes; I've opened a support ticket with AWS re: the documentation discrepancy.  Thanks for your help.

markpeek:
[...]

PR #1599, opened 2020-02-24
Fixing a missimplementation of rules, caused by a bug in the document…

axelpavageau:
…ation


axelpavageau:
aims to fix #1598

markpeek:
Thanks! Appreciate you deciphering the docs to figure out the different Rules being used.

Issue #1600, opened 2020-02-26
Issue with Aurora RDS VALID_SCALING_CONFIGURATION_CAPACITIES


andrewthetechie:
https://github.com/cloudtools/troposphere/blob/master/troposphere/rds.py#L24

These configurations are only valid for Aurora-mysql. Per https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-dbcluster-scalingconfiguration.html#cfn-rds-dbcluster-scalingconfiguration-maxcapacity, the valid Max capacites are:

For Aurora MySQL, valid capacity values are 1, 2, 4, 8, 16, 32, 64, 128, and 256.

For Aurora PostgreSQL, valid capacity values are 2, 4, 8, 16, 32, 64, 192, and 384.

PR #1602, opened 2020-02-27
fix: include valid postgres capacity configurations

andrewthetechie:
Fixes https://github.com/cloudtools/troposphere/issues/1600
We are using troposphere and stacker to spin up Aurora Serverless
clusters using Postgresql and received errors that our MaxCapacity
arguments to ScalingConfiguration were failing validation even though
AWS documentation showed them to be correct.

This adds the valid options for postgres as well as mysql


markpeek:
Thanks!

PR #1604, opened 2020-03-03
adding misc FSX properties, per Feb 27 2020 update

axelpavageau:


mcrozes:
Hey Axel -
I was about to add support for this feature as well, thanks for doing it.

Hey @markpeek  - FSx for lustre v2 is out and merging this PR would be extremely useful. Do you see any challenge with it?
Per https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-fsx-filesystem-lustreconfiguration.html, Amazon has added `DeploymentType` and `PerUnitStorageThroughput`. The latter is required when `DeploymentType` is set to `PERSISTENT_1`

markpeek:
Thanks!

PR #1607, opened 2020-03-05
Solve problems with attributes in WebAcl and ByteMatchStatement

eloymg:
I change this attributes for a direct use of a array of elements for solve errors in cloudformation. Now cloudformation says "Internal error" without this modifications. I think this problem is in other attributes, but I haven't tested ( With the bug of "internal error" is very difficult to debug ).


markpeek:
Thank you for the PR. Part of this was fixed via #1599 and a commit I just pushed due to more changes that were needed.

PR #1611, opened 2020-03-11
adding AWS::CodeStarConnections::Connection, per Marche 5 2020 update

axelpavageau:


markpeek:
Thanks! Will do a couple fixups in a follow up commit.

PR #1617, opened 2020-03-26
Fix several problems in wafv2

marksidell:
- Several classes were subclasses of AWSObject when they should be  subclasses of AWSProperty.

- Several classes have arguments that are lists. These args were defined via intermediate classes that in turn defined the lists. This was incorrect. The classes should accept the lists directly.

  I fixed the list arguments for these classes:

    GeoMatchStatement
    AndStatementOne
    OrStatementOne
    AndStatementTwo
    OrStatementTwo

  I remove these unnecessary intermediate list classes:

    CountryCodes
    StatementTwos
    StatementThrees

### DIFF
--- a/troposphere/wafv2.py
+++ b/troposphere/wafv2.py
@@ -90,31 +90,31 @@ class SingleQueryArgument(AWSProperty):
     }
 
 
-class Body(AWSObject):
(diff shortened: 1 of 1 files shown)


markpeek:
Thank you!

PR #1619, opened 2020-03-28
IoT Analytics fixes and additions

kteague:
A number of fixes for IoT Analytics: Datastore AWSProperty name duplicated/shadowed the Datastore AWSObject name; RetentionPeriod class was listed twice, added ChannelStorage and DatastoreStorage AWSProperties.

Adds an example IoTAnalytics.py template.

I think this brings IoTAnalytics up to sync with CloudFormation - I've started using it to make some IoTAnalytics resources.

Also added IotAnalytics and StepFunctions to iot.Action property for iot.py

### DIFF
--- a/examples/IoTAnalytics.py
+++ b/examples/IoTAnalytics.py
@@ -1,0 +1,126 @@
+from troposphere import Template
+from troposphere import Tags
+from troposphere.iotanalytics import (
+    Channel,
+    ChannelStorage,
+    CustomerManagedS3,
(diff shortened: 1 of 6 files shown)


kteague:
Changed Attributes  prop for AddAttributes from a json_checker to a dict type. It's the same dict format as RDS.DBParameterGroup - the CloudFormation won't provision as a JSON string.

markpeek:
Thank you!

PR #1620, opened 2020-04-01
add secrets manager type to codebuild environment variable types

mng12689:
AWS CodeBuild added support for the SECRETS_MANAGER environment variable type. troposphere should be able to support it as well.


markpeek:
Thanks!

PR #1623, opened 2020-04-08
Add athena workgroup and associated resources

gsoyka:
Adds Athena WorkGroup and child resources based on documentation here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-athena-workgroup.html


markpeek:
Thank you for the change. Duplicate of #1609 which was already merged.

Issue #1624, opened 2020-04-13
apigateway usageplan ThrottleSettings


iainlbc:
The value for "Throttle" within "UsagePlan" is represented in cloudformation as:
```
Throttle:
    "<resource Path>":  //(for example "/{item}/GET")
         BurstLimit: 10
         RateLimit: 5
```

But troposphere is expecting only a map with BurstLimit and RateLimit, with no containing resource path:
```
class UsagePlan(AWSObject):
    resource_type = "AWS::ApiGateway::UsagePlan"

    props = {
        "ApiStages": ([ApiStage], False),
        "Description": (basestring, False),
        "Quota": (QuotaSettings, False),
        "Tags": (Tags, False),
        "Throttle": (ThrottleSettings, False),
        "UsagePlanName": (basestring, False),
    }
...

class ThrottleSettings(AWSProperty):
    props = {
        "BurstLimit": (positive_integer, False),
        "RateLimit": (positive_integer, False),
    }
```
https://github.com/cloudtools/troposphere/blob/master/troposphere/apigateway.py#L416

Their docs don't explain it well, but the above is "straight from the horses mouth" (the rep on the AWS end of my support case).
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigateway-usageplan-apistage.html

marinpurgar:
Current implementation looks proper:
https://github.com/awsdocs/aws-cloudformation-user-guide/blob/06f1a219aa82f2dfc76cf7eec1ce93384ee6b977/doc_source/aws-properties-apigateway-usageplan-throttlesettings.md

marinpurgar:
I would say the wrong implementation is in the ApiStage

https://github.com/cloudtools/troposphere/blob/2dc788dbc89c15ce5984f9c40b143494336a2348/troposphere/apigateway.py#L404

There the "Throttle" property should be "Map containing method-level throttling information for API stage in a usage plan."

This is the implementation in AWS CDK:

https://github.com/aws/aws-cdk/blob/cb30d3d27cdc54b9b6eefe3ea5a58218085fff1a/packages/%40aws-cdk/aws-apigateway/lib/usage-plan.ts#L258

iainlbc:
Yep, came to same conclusion

markpeek:
Thanks @iainlbc and @marinpurgar. #1625 merged.

PR #1625, opened 2020-04-14
Update ApiStage to expect dict for Throttle

iainlbc:
regarding #1624

Allow usagePlan ApiStage's to pass a dict for Throttle


markpeek:
Thanks!

PR #1650, opened 2020-04-14
Adding misc AWS::ApiGatewayV2::Integration properties, per March 26, 2020 update

axelpavageau:
implements #1637


markpeek:
Thanks!

Issue #1655, opened 2020-04-16
Fix type of ApiGateway::RestApi FailOnWarnings property


bradarndt:
Change from string to bool

PR #1656, opened 2020-04-16
Fix FailOnWarnings datatype

bradarndt:
Change the datatype of ApiGateway::RestApi FailOnWarnings property from string to bool

implements #1655


markpeek:
For some reason this branch had conflicts (not sure why). Made the change to master to use a 'boolean'. Thanks for the issue/PR.

PR #1657, opened 2020-04-17
Update SAM Schedule event source spec

Anheurystics:
As per this document: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-schedule.html


markpeek:
Thanks!

Issue #1677, opened 2020-05-12
wafv2.py RegularExpressionList should not take a RegularExpressionList object


jonmoss-konsentus:
When trying to create a Regular Expression matcher in WafV2 using troposphere you have to add a  RegularExpressionList object to a RegularExpressionList parameter.

Looking at the CFN documentation (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-wafv2-regexpatternset.html#cfn-wafv2-regexpatternset-regularexpressionlist)

I think that RegularExpressionList should just take a list of strings (Regex validated).

Currently you would have to write the following python code
```python
cfn.add_resource(
    RegexPatternSet(
        "fooRegex",
        Name="match_foo",
        Description="regex to match /foo",
        RegularExpressionList=RegularExpressionList(
            RegularExpressionList=[Regex(RegexString="^\/foo.*")]
        ),
        Scope="REGIONAL",
    )
)
```

Which outputs the following CFN
```yaml
      fooRegex:
        Properties:
          Description: regex to match /foo
          Name: match_foo
          RegularExpressionList:
            RegularExpressionList:
              - RegexString: ^\/foo.*
          Scope: REGIONAL
        Type: AWS::WAFv2::RegexPatternSet
```

If this you deploy this CFN to AWS it fails with Internal Failure

PR #1678, opened 2020-05-12
RegularExpressionList fix

jonmoss-konsentus:
Potential fix for issue https://github.com/cloudtools/troposphere/issues/1677

Change RegularExpressionList to take a list of Regex and set required True as per the documentation - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-wafv2-regexpatternset.html#cfn-wafv2-regexpatternset-regularexpressionlist

### DIFF
--- a/troposphere/wafv2.py
+++ b/troposphere/wafv2.py
@@ -393,19 +393,13 @@ class IPSet(AWSObject):
     }
 
 
-class Regex(AWSProperty):
-    props = {
-        'RegexString': (basestring, False)
-    }
-
-
 class RegexPatternSet(AWSObject):
     resource_type = "AWS::WAFv2::RegexPatternSet"
 
     props = {
         'Description': (basestring, False),
         'Name': (basestring, True),
-        'RegularExpressionList': ([Regex], True),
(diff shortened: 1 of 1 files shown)


troposphere/wafv2.py
```diff
@@ -399,19 +399,13 @@ class Regex(AWSProperty):
...
-
-
 class RegexPatternSet(AWSObject):
     resource_type = "AWS::WAFv2::RegexPatternSet"
 
     props = {
         'Description': (basestring, False),
         'Name': (basestring, True),
-        'RegularExpressionList': (RegularExpressionList, False),
+        'RegularExpressionList': ([Regex], True),
```
markpeek, line 408: The docs specify "List of String". Should this be "[basestring]" and remove "Regex"?

jonmoss-konsentus: Good point, I mistakenly assumed that the Regex object had some form of validation added. I had verified that it doesn't. The PR has been amended, thanks for reviewing.

markpeek:
Thanks!

PR #1683, opened 2020-05-13
Adding missing AWS::ApiGatewayV2::VpcLink object

### LINKED ISSUE #1682: ApiGatewayV2 is missing VpcLink
We appear to be unable to create an HTTP Vpc Link, as the ApiGatewayV2 Vpc Link is not yet implemented.

From the AWS documentation:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-vpclink.html

From the Troposphere codebase, VpcLink is missing.

https://github.com/cloudtools/troposphere/blob/master/troposphere/apigatewayv2.py

Please note that this is not the same thing as the ApiGateway VpcLink, as that does not support Http Vpc Links.

axelpavageau:
fixes #1682

### DIFF
--- a/troposphere/apigatewayv2.py
+++ b/troposphere/apigatewayv2.py
@@ -333,3 +333,14 @@ class Stage(AWSObject):
         'StageVariables': (dict, False),
         'Tags': (dict, False),
     }
+
+
+class VpcLink(AWSObject):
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks!

PR #1684, opened 2020-05-13
Update instance types in constants

michael-k:
Sources:
- https://aws.amazon.com/ec2/instance-types/
- https://aws.amazon.com/ec2/instance-types/m6/ (for `m6g.metal`)


markpeek:
Thanks!

PR #1703, opened 2020-05-25
Fixing misc AWS::ImageBuilder properties

### LINKED ISSUE #1700: AWS::ImageBuilder::* some wrong data types and attribut missing
In imagebuilder.py (2.6.1 release).

* In AWS::ImageBuilder::*
"Tags" are "json_checker" but should be dict
When encoded in the structure it becomes a string (validator.py, ligne 258,` json.dumps(prop)`) which is creating an issue with CloudFormation that expect a struct like : `"Tags" : {Key : Value, ...}`
* AWS::ImageBuilder::DistributionConfiguration::Distribution
"AmiDistributionConfiguration" is "json_checker" but should be dict.
For the same as above "Tags" 
* In AWS::ImageBuilder::Component
"Data" is missing. And should be "json_checker" in that case.
* In AWS::ImageBuilder::ImageRecipe::InstanceBlockDeviceMapping
"NoDevice" is boolean but should be a string

axelpavageau:
Fixes #1700

### DIFF
--- a/troposphere/imagebuilder.py
+++ b/troposphere/imagebuilder.py
@@ -4,8 +4,8 @@
 # See LICENSE file for full license.
 
 from . import AWSObject, AWSProperty
-from .validators import (integer, boolean, json_checker,
-                         component_platforms, imagepipeline_status,
+from .validators import (integer, boolean, component_platforms,
+                         imagepipeline_status,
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks!

Issue #1724, opened 2020-06-10
AWS Glue Security Configuration EncryptionConfiguration S3Encryptions bug


khutch5:
There is an error in the AWS Cloudformation Documentation on how to create S3 Encryptions for Glue Security configuration [S3Encryptions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-glue-securityconfiguration-s3encryptions.html) The correct values are on a page for S3 Encryption [S3Encryption](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-glue-securityconfiguration-s3encryption.html) 

Because of this documentation issue it appears that the S3Encryptions class in glue.py has been built with no properties. I was using this project to generate Cloudformation for a Glue job that needed to write to an S3 bucket that required SSE-KMS. The only way I was able to do that was to modify the glue.py code for S3Encryptions like so:
`class S3Encryptions(AWSProperty):
    props = {
        'KmsKeyArn': (str, False),
        'S3EncryptionMode': (str, False),
    }`
And to modify the EncryptionConfiguration for S3Encryptions code to:
`'S3Encryptions': ((S3Encryptions,list), False)`

I had to do this because Cloudformation only allows an S3Encryptions object (S3Encryption is invalid) and it must be an array. I have added the correct properties and modified the EncryptionConfiguration class to generate an array for S3Encryptions. I have tested the output of this to deploy a new AWS Glue job and it was successful.

PR #1725, opened 2020-06-10
Modifying glue.py to generate valid CloudFormation output for Glue Jo…

khutch5:
Modifying glue.py to generate valid CloudFormation output for Glue Jobs that need to work with an encrypted S3 Bucket. #1724


markpeek:
Thanks! Appreciate the reworked changes.

PR #1747, opened 2020-07-21
Allow List<Number> param to have MinValue, MaxValue

nigelis:
AWS CloudFormation validates the List<Number> parameter value as numbers.

See:
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html


nigelis:
Hi, is there anything wrong about the PR?

PR #1749, opened 2020-07-23
Add AWS::Serverless::Api's Domain

michael-k:
See https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-api.html#sam-api-domain


markpeek:
Thanks!

PR #1758, opened 2020-08-03
AppMesh Gateway support

JohnPreston:
Untested yet, adding support for AWS AppMesh Gateway:
* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appmesh-virtualgateway.html
* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appmesh-gatewayroute.html

Unfortunately, the Listerners, Match and actions for HTTP and gRPC are not exactly the same as the ones for the Nodes/Router so had to re-implement these properties on their own. Re-used existing AWSProperty whenever possible.

### DIFF
--- a/troposphere/appmesh.py
+++ b/troposphere/appmesh.py
@@ -13,7 +13,6 @@ from troposphere import Tags
 from .validators import boolean
 from .validators import integer
 
-
 VALID_LISTENERTLS_MODE = ('STRICT', 'PERMISSIVE', 'DISABLED')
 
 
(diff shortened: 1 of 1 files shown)


JohnPreston:
@markpeek Hello again :)
Going to test this today (by actually deploying it to CFN).
Otherwise seems to render fine

markpeek:
Thanks!

PR #1770, opened 2020-08-22
Added SSM Parameter examples

0xack13:
SSM Parameter usage examples with values that has inline python function from a standard-library.


markpeek:
Thanks!

PR #1778, opened 2020-08-30
Support for OpenApiVersion in serverless.Api

2start:
The parameter was missing.
Reference: [Cloudformation documentation](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-api.html)


markpeek:
Thanks!

Issue #1784, opened 2020-09-14
Type mismatch of Tags for WorkGroup in Athena


isambitd:
According to the aws cloudformation docs - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-athena-workgroup.html

Tags should be of type `Tags` but in athena.py it is dict()

markpeek:
This was fixed via your PR #1785. Thanks!

PR #1785, opened 2020-09-14
Fixing tags data type for the issue #1784

isambitd:
According to the aws cloudformation docs for athena workgroup, they are expecting an array of key-value pair of type `Tags`


markpeek:
Thanks. I am going to remove the list as that is only for backward compatibility with some early usage of Tags.

PR #1789, opened 2020-09-25
Changhan/owner assignment

chjong:


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* chris@clicktherapeutics.com mng@clicktherapeutics.com chjong@clicktherapeutics.com matthew@clicktherapeutics.com
```
PatMyron, line 1:
```suggestion
```

chjong:
this is a mistake

troposphere/codebuild.py
```diff
@@ -111,6 +111,7 @@ def validate(self):
             valid_types = [
                 'PARAMETER_STORE',
                 'PLAINTEXT',
+                'SECRETS_MANAGER'
```
PatMyron, line 114: Ah, just realized this is already merged in:

https://github.com/cloudtools/troposphere/blob/1ca9e8c0ef00fa5cf333aae038c7f362855e6cfb/troposphere/codebuild.py#L114

nevermind then

PR #1790, opened 2020-09-28
Run tests against Python 3.9

michael-k:
3.9.0 final is [expected](https://www.python.org/dev/peps/pep-0596/#schedule) on Monday, 2020-10-05.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
   - "3.6"
   - "3.7.3"
   - "3.8"
+  - "3.9-dev"
 env:
   global:
     - TROPO_REAL_BOOL=true
--- a/setup.py
+++ b/setup.py
@@ -67,6 +67,7 @@ setup(
         "Operating System :: POSIX :: Linux",
 
         "Programming Language :: Python",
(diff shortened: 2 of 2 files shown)


markpeek:
Thanks!

PatMyron:
@markpeek @michael-k Travis doesn't seem to be running at all for PRs

michael-k:
I can see some builds here: https://travis-ci.org/github/cloudtools/troposphere/pull_requests

But I'm not a big fan of Travis CI any more since they've let go many of their staff a while ago.  Should I prepare a PR to switch to GitHub Actions?  I've had great experiences with that so far.

#1814 probably errored because the PR was merged before it was able to start?

PatMyron:
@michael-k interesting, so Travis does seem to be running for PRs, but the results don't seem to show in PRs themselves

PatMyron:
[results showing in PRs as of this week](https://github.com/cloudtools/troposphere/pull/1844#issuecomment-738909401)

michael-k:
Should we switch to GitHub Actions?  So far I've only had positive experiences with it.

Issue #1792, opened 2020-09-30
EndpointConfiguration for a Serverless API is incorrectly set to String


chrishunter-konsentus:
The documentation for the Serverless API states that the EndpointConfiguration can have multiple values assigned:

Serverless API AWS Documentation:
https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-api.html

EndpointConfiguration Layout:
https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-api-endpointconfiguration.html

Currently troposphere has the EndpointConfiguration set as a `basestring`, when this should be a `dict`.

chrishunter-konsentus:
Related PR

https://github.com/cloudtools/troposphere/pull/1793

chrishunter-konsentus:
Merged PR.

PR #1793, opened 2020-09-30
Added Types to EndpointConfiguration

chrishunter-konsentus:
Serverless APIs can have Endpoint Configurations set that are more complex than a standard String


chrishunter-konsentus:
Related Issue

https://github.com/cloudtools/troposphere/issues/1792

markpeek:
Thanks!

PR #1794, opened 2020-10-02
Fix target type validation

### LINKED ISSUE #1539: Unit Test for TargetGroup Target Type seems to prevent using a parameter to specify the value
TargetGroup(AWSObject)

    def validate(self):
        one_of(self.__class__.__name__,
               self.properties,
               'TargetType',
               [
                   None,
                   TARGET_TYPE_INSTANCE,
                   TARGET_TYPE_IP,
                   TARGET_TYPE_LAMBDA
               ])

I have to hardcode the value of instance, ip, or lambda into my code. Any reason why this is forced particularly? I'd like to reuse the same template and in some environments I want a different target type.

alexeyts:
Allow TargetType to be Ref items, such as `Ref('AWS::NoValue')`, as well as parameters.

This will not allow using `None`, but one would argue `Ref('AWS::NoValue')` would be a better choice.

Also today, when trying to use `None` (at least in python 2.7), you would get the error:
`TargetType is <type 'NoneType'>, expected <type 'basestring'>`

Fixes #1539

### DIFF
--- a/troposphere/elasticloadbalancingv2.py
+++ b/troposphere/elasticloadbalancingv2.py
@@ -281,6 +281,16 @@ TARGET_TYPE_IP = 'ip'
 TARGET_TYPE_LAMBDA = 'lambda'
 
 
+def validate_target_type(target_type):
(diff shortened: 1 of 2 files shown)


markpeek:
Thanks!

PR #1796, opened 2020-10-03
Cloudfront cache and origin policy

mello7tre:
Added support for Cloudfront CachePolicy and OriginRequestPolicy with proper validators.

Tested creation with a template containing both CachePolicy and OriginRequestPolicy.

Note:
You cannot set MaxTLL = MinTLL = DefaultTTL = 0 in CachePolicyConfig.
If you do this CloudFormation give error ("Invalid request provided: AWS::CloudFront::CachePolicy").

By WebGui if you set all to zero you are no more able to configure ParametersInCacheKeyAndForwardedToOrigin's parameters as Cookie, Headers and QueryStrings.
This is probably the root cause of the error.
(ParametersInCacheKeyAndForwardedToOrigin is always required, so omitting it is not a solution)

In that case just set MaxTLL to one (maybe AWS will fix this in the future).

### DIFF
--- a/troposphere/cloudfront.py
+++ b/troposphere/cloudfront.py
@@ -319,7 +319,7 @@ class CachePolicyConfig(AWSProperty):
         'Comment': (basestring, False),
         'DefaultTTL': (integer, False),
         'MaxTTL': (integer, False),
-        'MinTTL': (integer, False),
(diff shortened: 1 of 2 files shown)


markpeek:
Thanks!

PR #1830, opened 2020-11-09
AWS::KinesisFirehose::DeliveryStream.Tags

Kadambari87:
Kinesis Firehose Delivery Stream is missing tags option. When a stream is created, consumer would like to add some tagging to the resource.

### DIFF
--- a/troposphere/firehose.py
+++ b/troposphere/firehose.py
@@ -3,7 +3,7 @@
 #
 # See LICENSE file for full license.
 
-from . import AWSObject, AWSProperty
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks!

Kadambari87:
@markpeek Is this change released? If so, how can I update?

PatMyron:
> @markpeek Is this change released? If so, how can I update?

Not released to PyPI yet, but it is possible to `pip install` directly from git repos if you need an immediate workaround:
```bash
pip install git+https://github.com/cloudtools/troposphere --upgrade
```

Issue #1831, opened 2020-11-13
AuthorizationConfig property on ecs.EFSVolumeConfiguation class needs required value set - not part of release 2.6.3


svsShiva:
Although this issue (https://github.com/cloudtools/troposphere/issues/1806) is resolved with this PR (https://github.com/cloudtools/troposphere/pull/1808), I am still facing the same issue. 

Seems like release 2.6.3 (https://pypi.org/project/troposphere/) does not include the fix.

Correct me if I am wrong - Thanks

PatMyron:
Not released to PyPI yet, but it is possible to `pip install` directly from git repos if you need an immediate workaround:
```bash
pip install git+https://github.com/cloudtools/troposphere --upgrade
```

svsShiva:
Thanks for the response. However, after installing it from GitHub source, I am seeing below issues.
 
E0401: Unable to import 'awacs.sts' (import-error)
E0401: Unable to import 'awacs.aws' (import-error)
 
Can anyone help me out with this?

markpeek:
This was included in the 2.6.4 release.

Issue #1806, opened 2020-10-13
AuthorizationConfig property on ecs.EFSVolumeConfiguation class needs required value set


avosper-intellaegis:
Required value is not set which results in a "TypeError: cannot unpack non-iterable type object" being thrown when attempting to create a new ecs.EFSVolumeConfiguration object.

markpeek:
Thank you for the issue.

PR #1832, opened 2020-11-13
Release 2.6.4

svsShiva:
Release new version with fix for issue regarding efs volume (https://github.com/cloudtools/troposphere/issues/1806). 

fixes https://github.com/cloudtools/troposphere/issues/1831

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -13,7 +13,7 @@ import types
 
 from . import validators
 
-__version__ = "2.6.3"
+__version__ = "2.6.4"
 
 # constants for DeletionPolicy and UpdateReplacePolicy
 Delete = 'Delete'


zbrookle:
@svsShiva Is this release planned to occur soon? There are some updates that my team needs to awslambda from back in November, so it would be great to have those in released version of the package

JohnPreston:
Hello @markpeek @PatMyron 
Any updates on when we can get 2.6.4 release in pypi ? Seems like a number of projects are waiting for that release to make their own.
Given this is a patch release (unless 2.7.0 or 3.0 is more imminent) can we get this out sooner than later?
Many thanks!

sir-sigurd:
2.6.4 was released:
https://github.com/cloudtools/troposphere/releases/tag/2.6.4
https://pypi.org/project/troposphere/2.6.4/

markpeek:
As noted, 2.6.4 was released on 2021-03-08. Look for upcoming releases:
- Will do a catchup release 2.7.0 which will be the final Python 2.x release
- Then switch the code to Python 3.x only for a 3.0.0 release.

PR #1836, opened 2020-11-18
Update apievent

gsoyka:
Update Serverless ApiEvent based on latest docs: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-api.html#sam-function-api-requestmodel


markpeek:
Thanks!

PR #1846, opened 2020-12-12
AWS::Lambda::Function support for container image deployment package

ITProKyle:
This has not been added to [CloudFormation documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html) yet but, the options are present in the [API documentation](https://docs.aws.amazon.com/lambda/latest/dg/API_CreateFunction.html#API_CreateFunction_RequestSyntax. When provided in a CloudFormation template in one of the regions that supports this feature it is working.

## Example

```yaml
Resources:
  TestFunction:
    Type: AWS::Lambda::Function
    Properties:
      Code:
        ImageUri: !Ref ImageUri
      Description: This is just a test.
      PackageType: Image
      Role: !Ref Role
```

### DIFF
--- a/tests/test_awslambda.py
+++ b/tests/test_awslambda.py
@@ -1,9 +1,10 @@
 import unittest
 from troposphere import GetAtt, Template, Join, Ref
-from troposphere.awslambda import Code, Function, Environment
+from troposphere.awslambda import Code, Function, Environment, ImageConfig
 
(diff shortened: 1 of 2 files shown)


markpeek:
Nice! Thanks!

PR #1848, opened 2020-12-16
AWS::CloudFront::Distribution.Origin.OriginShield

jerry153fish:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-origin.html#cfn-cloudfront-distribution-origin-originshield

Not add the validate for shield region as all other regions in other stacks not using validation.


markpeek:
Thanks!

PR #1850, opened 2020-12-22
Add ReplicaModifications of s3

nigelis:
Add ReplicaModifications, which is introduced in https://aws.amazon.com/about-aws/whats-new/2020/12/amazon-s3-replication-adds-support-two-way-replication/.

Reference:

- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-replicamodifications.html
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-sourceselectioncriteria.html

### DIFF
--- a/troposphere/s3.py
+++ b/troposphere/s3.py
@@ -295,6 +295,12 @@ class ReplicationConfigurationRulesDestination(AWSProperty):
     }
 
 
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks!

Issue #1855, opened 2021-01-06
Support MWAA::Environment resource


ghost:
The AWS::MWAA::Environment resource creates an Amazon MWAA environment with Apache Airflow.  Troposphere doesn't appear to support yet?

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-mwaa-environment.html

adam133:
Came here looking to note this issue as well. It's going to slow me down a bit on a project I'm working on.

adam133:
I created a PR for it... My first time contributing here so I'm not 100% sure it's correct, but hopefully speeds up the process for getting support for it.
https://github.com/cloudtools/troposphere/pull/1858/commits/eb61cf59130e4134a947e6466fea870691e3f437

markpeek:
Fixed via above PR.

PR #1858, opened 2021-01-16
AWS::MWAA Adding for managed airflow

adam133:
For this issue: https://github.com/cloudtools/troposphere/issues/1855
My first time contributing to this library, so I'm guessing I'm missing some things. But hopefully someone can use this as a starting point or at least check that I added things correctly!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -235,6 +235,7 @@ Currently supported AWS resource types
 - `AWS::MediaLive`_
 - `AWS::MediaStore`_
 - `AWS::MSK`_
+- `AWS::MWAA`_
 - `AWS::Neptune`_
 - `AWS::NetworkManager`_
 - `AWS::OpsWorks`_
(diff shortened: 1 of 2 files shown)


adam133:
Created to close this issue: https://github.com/cloudtools/troposphere/issues/1855

adam133:
Needs some work yet - I'm missing some required parameters. Update incoming.

adam133:
Just wanted to share that using this PR locally I was able to create a cloudformation stack with a MWAA environment, so it seems like it works. Might be good to add more validation when I get time though. For example, the source bucket name has to start with `airflow-`, and there is a minimum of two subnets for the security group.

markpeek:
Since this change was written for Python3 (str instead of basestring), I'll merge this in as part of the 3.0.0 release.

markpeek:
Thanks!

PR #1864, opened 2021-02-06
docs: use Template.set_metadata instead of add_metadata

mloberg:
`add_metadata` is deprecated, switch to `set_metadata` instead


markpeek:
Thanks!

PR #1873, opened 2021-03-01
Support Throughput for gp3 ebs volumes

mtougeron:
This allows setting the `Throughput` field for the AWS `gp3` ebs volumes when creating an `Ec2EBSBlockDevice`.

### DIFF
--- a/troposphere/ec2.py
+++ b/troposphere/ec2.py
@@ -159,6 +159,7 @@ class EBSBlockDevice(AWSProperty):
         'Encrypted': (boolean, False),
         'KmsKeyId': (basestring, False),
         'Iops': (integer, False),  # Conditional
+        'Throughput': (integer, False),  # Conditional
         'SnapshotId': (basestring, False),  # Conditional
         'VolumeSize': (integer, False),  # Conditional
         'VolumeType': (basestring, False),


markpeek:
I'm not seeing this in the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-blockdev-template.html) or in the specification files. Could you let me know where you saw this property documented?

mtougeron:
It's unfortunately not available for all EBS resources but it is for some like [`LaunchTemplateEbsBlockDevice`](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_LaunchTemplateEbsBlockDevice.html) which use the same `EBSBlockDevice` other EC2 resources. Like `Iops` being conditional based on the volume type I thought it would be fine to add `Throughput` as conditional as well.

Thoughts?

markpeek:
Got it. Thanks for the quick reply. I'll merge it now. I'll have to think about breaking these out separately in a future major version (since it will be a breaking change). Any thoughts on this?

mtougeron:
> Got it. Thanks for the quick reply. I'll merge it now. I'll have to think about breaking these out separately in a future major version (since it will be a breaking change). Any thoughts on this?

Yeah, I was a little torn on this myself. I'm going to be opening a ticket with AWS for a bug report that their API doesn't allow setting that value at instance creation. :(

mtougeron:
thanks for merging the PR!

Asura9:
This not applies to AWS::EC2::Volume, it affects only AWS::EC2::LaunchTemplate.

PR #1874, opened 2021-03-03
Fix autoscaling.Tags to use boolean instead of str

marinpurgar:
In the https://github.com/cloudtools/troposphere/pull/1409  all of the boolean parameters were converted from str to boolean when TROPO_REAL_BOOL=true,
The autoscaling.Tags was always converting the boolean parameter into str.


markpeek:
Thanks!

PR #1886, opened 2021-03-30
lambda memory can be configured in 1 MB increments now

sir-sigurd:
https://aws.amazon.com/blogs/aws/new-for-aws-lambda-functions-with-up-to-10-gb-of-memory-and-6-vcpus/


troposphere/awslambda.py
```diff
@@ -5,11 +5,6 @@
 
 MINIMUM_MEMORY = 128
 MAXIMUM_MEMORY = 10240
-MEMORY_INCREMENT = 64
-MEMORY_VALUES = [
```
sir-sigurd, line 9: I'm not sure if these are from public API and they should be kept.

markpeek:
Thanks!

PR #1887, opened 2021-04-01
Update fis.py

vasinov:
Even though the CF docs say [1] that FIS resource keys are in camelCase they are actually in PascalCase (just like any other standard AWS resource). It's easily testable by submitting a CF template. I'm contacting AWS to notify them about this discrepancy.

1. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-fis-experimenttemplate-experimenttemplateaction.html


markpeek:
Thanks! Yes, the previous spec file must have had the camelCase but is now fixed.

PR #1895, opened 2021-04-27
Create Export instances for Output.Export in cfn2py

### LINKED ISSUE #1878: TypeError for Export dict generated by cfn2py
While using the `cfn2py` script to generate code from a CFN template, it generated:

```python
NATGatewayID = t.add_output(
    Output(
        "NATGatewayID",
        Description="NATGateway ID",
        Value=Ref(NATGateway),
        Export={"Name": Join("-", [Ref("AWS::StackName"), "nat"])},
    )
)
```

This generated a `TypeError`:
```
TypeError: <class 'troposphere.Output'>: NATGatewayID.Export is <class 'dict'>, expected <class 'troposphere.Export'>
```

It works OK with a manual edit to use `troposphere.Export` like:

```python
NATGatewayID = t.add_output(
    Output(
        "NATGatewayID",
        Description="NATGateway ID",
        Value=Ref(NATGateway),
        Export=Export(Join("-", [Ref("AWS::StackName"), "nat"])),
    )
)
```

Another anomaly in a round-trip test is that troposphere `t.to_json()` output seems to output boolean values as strings:

```
$ diff /tmp/template-cfn.json /tmp/template-py.json

243c243
<         "MapPublicIpOnLaunch": false,
---
>         "MapPublicIpOnLaunch": "false",

$ grep 'MapPublicIpOnLaunch' /tmp/template-cfn.json 
        "MapPublicIpOnLaunch": false,
        "MapPublicIpOnLaunch": false,
        "MapPublicIpOnLaunch": true,
$ grep 'MapPublicIpOnLaunch' /tmp/template-py.json 
        "MapPublicIpOnLaunch": "false",
        "MapPublicIpOnLaunch": "false",
        "MapPublicIpOnLaunch": "true",
$ grep 'MapPublicIpOnLaunch' template_vpc.py
    MapPublicIpOnLaunch=True,
    MapPublicIpOnLaunch=False,
    MapPublicIpOnLaunch=False,
```

The `"false"` value is wrong, it should be a JSON `false` value.  The python code seems to have boolean values, e.g.

```python
publicSubnetA = t.add_resource(Subnet(
    "publicSubnetA",
    VpcId=Ref(VPC),
    AvailabilityZone=Select(0, GetAZs("")),
    CidrBlock="172.31.10.0/24",
    MapPublicIpOnLaunch=True,
    Tags=Tags(
        Name=Sub("${AWS::StackName}-public-subnet-a"),
    ),
))
```

To check in pdb:
```python
(Pdb) template_py = json.loads(t.to_json(indent=0, sort_keys=True))
(Pdb) template_py['Resources']['privateSubnetA']['Properties']['MapPublicIpOnLaunch']
'false'
(Pdb) t.to_dict()['Resources']['privateSubnetA']['Properties']['MapPublicIpOnLaunch']
'false'
```

Why does the `t.to_dict()` and `t.to_json()` data contain strings instead of JSON booleans?  It seems like `encode_to_dict(obj)` should be replaced with just a `json.loads(json.dumps(obj))` and let the `json` lib take care of all the necessary python/JSON compatibility and encodings.  Or using marshmallow or pydantic models in general should take care of all the schema mappings and serializations.  It might also be easier to use botocore service descriptions or other AWS json payloads to auto-generate json-schema and models.  It's not quite the same thing as CFN templates, but botocore has service API descriptions in e.g. `lib/python3.7/site-packages/botocore/data/cloudformation/2010-05-15/service-2.json`; see also
- https://pydantic-docs.helpmanual.io/datamodel_code_generator/
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html
- - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html

Are there any round-trip unit tests on the `cfn2py` script?

## versions

```
$ python --version
Python 3.7.10

$ poetry show troposphere
name         : troposphere
version      : 2.6.4
description  : AWS CloudFormation creation library

dependencies
 - awacs >=0.8
 - cfn_flip >=1.0.2
```

aknauss-dod:
Fixes #1878 in part.

### DIFF
--- a/scripts/cfn2py
+++ b/scripts/cfn2py
@@ -354,7 +354,7 @@ def do_outputs(d):
             if isinstance(pv, str):
                 print('    {}="{}",'.format(pk, pv))
             elif pk == 'Export':
-                print('    {}={}(name={})'.format(pk, pk, output_value(pv)))
+                print('    {}={}("{}")'.format(pk, pk, json.loads(output_value(pv))['Name']))
             else:
                 print('    {}={},'.format(pk, output_value(pv)))
         print("))")
--- a/scripts/cfn2py
+++ b/scripts/cfn2py
@@ -56,7 +56,7 @@ def do_header(d):
     print('from troposphere import Base64, Select, FindInMap, GetAtt')
     print('from troposphere import GetAZs, Join, Output, If, And, Not')
     print('from troposphere import Or, Equals, Condition')
-    print('from troposphere import Parameter, Ref, Tags, Template')
+    print('from troposphere import Parameter, Ref, Tags, Template, Export')
     print('from troposphere.cloudformation import Init')
     print('from troposphere.cloudfront import Distribution')
     print('from troposphere.cloudfront import DistributionConfig')
@@ -353,6 +353,8 @@ def do_outputs(d):
         for pk, pv in v.items():
             if isinstance(pv, str):
                 print('    {}="{}",'.format(pk, pv))
+            elif pk == 'Export':
+                print('    {}={}(name={})'.format(pk, pk, output_value(pv)))
             else:
                 print('    {}={},'.format(pk, output_value(pv)))
         print("))")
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks!

PR #1896, opened 2021-04-29
ec2 volume throughput

MarshallMoberg:
noticed the [throughput](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-ebs-volume.html#cfn-ec2-ebs-volume-throughput) property is missing from the AWS::EC2::Volume resource object.

couldn't find any existing unit tests for this resource.


markpeek:
Thanks!

PR #1897, opened 2021-05-05
Transit-Gateway MulticastSupport

tnielsen2:
Add `MulticastSupport` attribute to `TransitGateway` that is currently supported by CloudFormation. Tested and validated as working.


markpeek:
Thanks!

PR #1898, opened 2021-05-06
Add helpers.userdata.from_file_sub()

noamtamim:
Imports userdata from a file, using `Sub` for replacing inline variables such as `${AWS::Region}`.

Useful with a userdata script that looks like this:
```bash
#!/bin/bash -ex

echo Region: ${AWS::Region}
...
```

Unlike `from_file()`, doesn't do any text processing.


markpeek:
Thanks!

PR #1905, opened 2021-05-24
Add constants for RDS instance types: R6G

david-octopart:
https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Concepts.DBInstanceClass.html#Concepts.DBInstanceClass.SupportAurora


markpeek:
Thanks!

PR #1906, opened 2021-05-24
Add fargate ephemeral storage property

gsoyka:
Adds ephemeral storage property for use with Fargate as documented here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-ephemeralstorage.html


markpeek:
Thanks!

PR #1917, opened 2021-06-14
Add ConnectivityType property for NatGateway

reaperharvest:
Newly released property for NatGateway to remove the internet gateway requirement when using NAT Gateway for internal natting. See updated Cloudformation docs:https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-natgateway.html

### DIFF
--- a/troposphere/ec2.py
+++ b/troposphere/ec2.py
@@ -164,7 +164,8 @@ class NatGateway(AWSObject):
     resource_type = "AWS::EC2::NatGateway"
 
     props = {
-        "AllocationId": (str, True),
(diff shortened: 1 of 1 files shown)


reaperharvest:
If someone from this project could review the proposed changes, it is a fairly simple update to account for a new property in Cloudformation for NAT Gateway. I need this update to continue implementation of a project.

markpeek:
Thanks!

reaperharvest:
@markpeek When do merges to master typically make their way into a new release?

PR #1921, opened 2021-06-24
Add new sns event parameters

gsoyka:
Add parameters as documented here: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-sns.html


markpeek:
Thanks!

PR #1922, opened 2021-06-26
(troposphere.serverless) add support for container-based Serverless::Function and add missing props

scrthq:
Heyo -- This is similar to #1846, except this is adding container-based lambda support for `Serverless::Function` as well. I've also filled in some properties I found missing based on the [current spec of the `AWS::Serverless::Function` resource](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html).

Please let me know if there are any adjustments needed!

### DIFF
--- a/troposphere/serverless.py
+++ b/troposphere/serverless.py
@@ -5,18 +5,21 @@
 
 import types
 
-from . import AWSObject, AWSProperty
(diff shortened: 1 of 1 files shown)


markpeek:
Manually committed. Fixed black/lint formatting issue. Thanks!

PR #1925, opened 2021-07-05
fix(AWS::EKS::Nodegroup.Taints): use the correct key for taints 

richardkeit:
Correct key is not `Name` it is `Key`
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-eks-nodegroup-taint.html


markpeek:
Doh! Nice catch. Thanks!

jerry153fish:
Thanks very much.

Issue #1926, opened 2021-07-06
Missing cfn-flip requirement in 3.0.0


anne-k:
The 3.0.0 release no longer has cfn-flip listed as dependency which leads to an error on startup if installed in a clean environment (ModuleNotFoundError: No module named 'cfn_flip'). 
Looks like it happened with the change to setup.cfg; the old setup.py adds the contents of requirements.txt as dependency but setup.cfg doesn't.

jerry153fish:
hi @anne-k, I had a similar situation, ending up put cfn_flip in the project requirements.txt.

povesma:
Also impacts Zappa: https://github.com/zappa/Zappa/issues/998

markpeek:
Thank you for the issue. Fixed via #1927 and released 3.0.1 to address.

PR #1927, opened 2021-07-06
Include cfn_flip in setup.cfg

jerry153fish:

PR #1928, opened 2021-07-06
Add `cfn_flip` to install_requires again

michael-k:
Fixes cloudtools/troposphere#1926

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,6 +30,7 @@ test_suite = tests
 tests_require = awacs>=2.0.0
 include_package_data = true
 install_requires =
+    cfn_flip >= 1.0.2
     typing-extensions >= 3.7.4.3; python_version < "3.8"
 packages =
     troposphere


michael-k:
We could also make it an optional dependency.

1. Add `yaml = cfn_flip>=1.0.2` to `[options.extras_require]` in `setup.cfg`
2. Add a try-except around [`import cfn_flip`](https://github.com/cloudtools/troposphere/blob/a3a87a258a3f46371b1be54340f236d30e5d06db/troposphere/__init__.py#L11).
3. Raise an exception in [`def to_yaml()`](https://github.com/cloudtools/troposphere/blob/a3a87a258a3f46371b1be54340f236d30e5d06db/troposphere/__init__.py#L773-L776) that explains that `troposphere[yaml]` should be installed for YAML support.

But I think YAML is widely used and it's worth always installing it.  Any other opinions?  Any data on JSON vs. YAML for CloudFormation?

markpeek:
Fixed via #1927, added a test to catch in "make release-test", and released 3.0.1. Thanks for the fix though. yaml seems popular enough to make it a mandatory requirement.

michael-k:
Didn't see that one as it didn't link the issue :shrug:

jerry153fish:
Sorry about that forget to link the issue as I had the fixes then I replied in the issue.

PR #1930, opened 2021-07-08
[batch] Update ContainerProperties properties

kumapo:
A property is necessary (but not required) when using Batch on Fargate:

`ContainerProperties.NetworkConfiguration`:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-jobdefinition-containerproperties.html#cfn-batch-jobdefinition-containerproperties-networkconfiguration

### DIFF
--- a/troposphere/batch.py
+++ b/troposphere/batch.py
@@ -106,6 +106,13 @@ class MountPoints(AWSProperty):
     }
 
 
+class NetworkConfiguration(AWSProperty):
+
+    props = {
+        "AssignPublicIp": (str, False),
+    }
+
+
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks @kumapo. Can you remove the p3s/gen.py change from this PR? That directory got committed by mistake and will be removed. The correct version is in scripts/gen.py.

kumapo:
@markpeek Thank you, I didn't notice about that. I've update the PR.
If you find something not so good in it, please let me know.

markpeek:
Thanks!

PR #1931, opened 2021-07-12
Add KmsKeyId Attribute to LogGroup

cartermeyers:
This attribute is supported by CloudFormation: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-loggroup.html#aws-resource-logs-loggroup-syntax


markpeek:
Thanks!

Issue #1933, opened 2021-07-13
networkfirewall.FirewallPolicy doesn't accurately reflect CFN object structure


avosper-intellaegis:
AWS CFN docs:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-networkfirewall-firewallpolicy.html#cfn-networkfirewall-firewallpolicy-description

FirewallPolicy definition in [troposphere.networkfirewall.py](https://github.com/cloudtools/troposphere/blob/main/troposphere/networkfirewall.py#L76-L83)

should be: https://github.com/avosper-intellaegis/troposphere/commit/a785d9008f8d53b42ff58145df525f9754724310

Using as is results in `botocore.exceptions CreateStack operation "Invalid template resource property"` error

PR #1934, opened 2021-07-13
Fix/networkfirewall

avosper-intellaegis:
Addresses https://github.com/cloudtools/troposphere/issues/1933


avosper-intellaegis:
This doesn't fix everything, top level FirewallPolicy needs to be promoted to full AWSObject, same underlying issue also exists with networkfirewall.RuleGroup.

Am closing and will re-submit with a more complete solution.

markpeek:
@avosper-intellaegis thanks for the initial PR. Typically how I've handled having the same name for a Resource and Property is to postfix "Property" to the name. Such as, in this case, using "FirewallPolicyProperty" and keeping "FirewallPolicy" for the resource.

PR #1936, opened 2021-07-18
AWS::Logs::ResourcePolicy

jerry153fish:


troposphere/logs.py
```diff
@@ -1,6 +1,21 @@
 from . import AWSObject, AWSProperty
 from .constants import LOGS_ALLOWED_RETENTION_DAYS as RETENTION_DAYS
-from .validators import integer_list_item
+from .validators import integer_list_item, json_checker
+
+
+def validate_resource_policy(policy_document):
```
markpeek, line 6: This validation should take into account the other policy document types coming from awacs. This is imported via https://github.com/cloudtools/troposphere/blob/main/troposphere/compat.py
Note: the length checks won't work with some of these awacs types and additional checking needs to be done around the json_checker return types.

jerry153fish: hi @markpeek So the plan will be

1. build a more general policy document validator
2. cover the types of dict, Policy, PolicyDocumen, and string
3. add tests cover all above

Any other places need to be considered here?

markpeek: That looks like the right list of items. Thanks.

jerry153fish:
@markpeek I can not reproduce the flake errors on my mac and ubuntu locally, do you know what might happen?

markpeek:
Excellent. Thank you for the commit.

jerry153fish:
Thanks my pleasure.

markpeek:
Oh good catch. Fixed via a7716408e1dc3693ec4157ea170eb6213ef85b36

```
    Fix duplicate AWS::Logs::ResourcePolicy

    The newly added LogResourcePolicy already existed as ResourcePolicy
    in Logs. Prefer the older one as it aligns with the AWS naming.
    This does keep the newly added validation and tests.
```

jerry153fish:
thx, I missed that.

PR #1938, opened 2021-07-21
Fix CanarySettings PercentTraffic definition

### LINKED ISSUE #1937: Apigateway CanarySettings PercentTraffic definition is wrong
I attempted to use PercentTraffic and it gave an error that it expected a list of doubles:

`TypeError: <class 'troposphere.apigateway.CanarySetting'>: None.PercentTraffic is <class 'float'>, expected [<function double at 0x10e783200>]`

I wrapped the double in a list, and cloudformation gave an error that it must be a string:

`Value of property PercentTraffic must be of type String`.

After cloning and modifying the definition from `"PercentTraffic": ([double], False),` to `"PercentTraffic": (double, False),` my cloudformation stack creates without any issues.

gcochard:
Fixes #1937

### DIFF
--- a/troposphere/apigateway.py
+++ b/troposphere/apigateway.py
@@ -78,7 +78,7 @@ class CanarySetting(AWSProperty):
 
     props = {
         "DeploymentId": (str, False),
-        "PercentTraffic": ([double], False),
+        "PercentTraffic": (double, False),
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks!

PR #1940, opened 2021-07-27
Enable MSK IAM Role based authentication

Markitox:
- Enable Iam authentication configuration
- Make Scram not required since now there is support for Iam

CloudFormation doc:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-msk-cluster-sasl.html

### DIFF
--- a/troposphere/msk.py
+++ b/troposphere/msk.py
@@ -39,9 +39,16 @@ class Scram(AWSProperty):
     }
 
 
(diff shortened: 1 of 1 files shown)


markpeek:
Thanks! I manually merged this one to fix the black formatting and alphabetizing the prop list.

PR #1943, opened 2021-08-11
Allow LaunchTemplateSpecification in LaunchTemplateOverrides

cetanu:
As per AWS document:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-autoscaling-autoscalinggroup-launchtemplateoverrides.html

### DIFF
--- a/troposphere/autoscaling.py
+++ b/troposphere/autoscaling.py
@@ -148,6 +148,7 @@ class InstancesDistribution(AWSProperty):
 class LaunchTemplateOverrides(AWSProperty):
     props = {
         "InstanceType": (str, False),
+        "LaunchTemplateSpecification": (LaunchTemplateSpecification, False),
         "WeightedCapacity": (str, False),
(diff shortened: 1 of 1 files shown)


troposphere/autoscaling.py
```diff
@@ -149,6 +149,7 @@ class LaunchTemplateOverrides(AWSProperty):
     props = {
         "InstanceType": (str, False),
         "WeightedCapacity": (str, False),
+        "LaunchTemplateSpecification": (LaunchTemplateSpecification, True),
```
markpeek, line 152: Thank you for the commit. Could you alphabetize this by moving it up a line? And I believe it should be "not required" so it should be False instead of True.

cetanu: Oops, sorry about that! I've fixed it up now.

markpeek:
Thanks!

Issue #1787, opened 2020-09-18
Support for Globals Section in SAM Templates


omijh:
https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-template-anatomy.html 

Globals Section from SAM template anatomy is missing in the troposphere.
we need this feature in order to update OpenApiVersion Version & Function timeout for all serverless functions in the template

markpeek:
Fixed via f5851870b26a6eef3f0a6beb0bef1db2deb705e8

PR #1950, opened 2021-08-31
Support Globals section for serverless

mariaines:
fixes https://github.com/cloudtools/troposphere/issues/1787

### DIFF
--- a/tests/test_template.py
+++ b/tests/test_template.py
@@ -1,3 +1,4 @@
+from troposphere.serverless import Globals
 import unittest
 
 from troposphere import (
@@ -36,6 +37,18 @@ class TestInitArguments(unittest.TestCase):
         template = Template()
         template.set_transform(transform)
         self.assertEqual(template.transform, transform)
+
+    def test_globals(self):
+        template = Template()
+        globals = Globals()
+        with self.assertRaises(ValueError):
+            template.set_globals(globals)
+        transform = "AWS::Serverless-2016-10-31"
+        template.set_transform(transform)
+        template.set_globals(globals)
+        self.assertEqual(template.globals, globals)
+        with self.assertRaises(ValueError):
+            template.set_transform("other_transform")
 
 
 class TestValidate(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


scrthq:
Thank you for this PR! Watching this for completion so we can leverage Globals to work around a long-standing issue with `Serverless::Api` and extra stage generation (https://github.com/aws/serverless-application-model/issues/191)

tests/test_template.py
```diff
@@ -37,6 +38,18 @@ def test_transform(self):
         template.set_transform(transform)
         self.assertEqual(template.transform, transform)
 
+    def test_globals(self):
+        template = Template()
+        globals = Globals()
+        with self.assertRaises(ValueError):
+            template.set_globals(globals)
+        transform = "AWS::Serverless-2016-10-31"
```
markpeek, line 46: Was this intentional to be an explicit string rather than importing SERVERLESS_TRANSFORM?

troposphere/serverless.py
```diff
@@ -615,3 +618,96 @@ def validate(self):
...
+    }
+
+
+class Globals(GlobalsHelperFn):
+    """Supported Globals properties.
+
+    See: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-template-anatomy-globals.html
+    """
+    props = {
+        "Function": (FunctionGlobals, False),
```
markpeek, line 709: Could you alphabetize these properties?

markpeek (changes requested):
Thank you for adding this change and appreciate the tests added as well! A couple of additional general comments:
- Running `make fix-black fix-isort test lint` makes needed formatting changes and lint is showing a couple of errors. (Note: likely need to fix our PR tests to catch these at submit time).
- Would it make sense to add an example as well? examples directory along with output in tests/examples_output

markpeek:
Fixed the above review feedback and pushed via f5851870b26a6eef3f0a6beb0bef1db2deb705e8. Thanks for adding the feature.

PR #1953, opened 2021-09-13
Update firehose.py parameter type validation

Kadambari87:
Added all processor types for validation of processor type. As per the https://docs.aws.amazon.com/firehose/latest/APIReference/API_Processor.html, the valid parameter types are RecordDeAggregation | Lambda | MetadataExtraction | AppendDelimiterToRecord which should be part of the validation.


markpeek:
Thanks!

PR #1954, opened 2021-09-22
Enable support for gp3 EBS volume types in Autoscaling Groups

hunt3ri:
Hi, the EBSBlockDevice properties class used by Autoscaling Groups does not support the creation of gp3 EBS volume types, because it doesn't currently handle the required **Throughput** parameter as described here [CloudFormation Docs](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-launchconfig-blockdev-template.html)

This PR adds the missing Throughput param and adds a test to confirm class can be instantiated correctly for gp3 volumes

Please let me know if any further updates or changes are needed.  This is a blocking issue currently on our environment.

Thanks
Iain

### DIFF
--- a/tests/test_asg.py
+++ b/tests/test_asg.py
@@ -1,7 +1,7 @@
 import unittest
 
 from troposphere import If, Ref
-from troposphere.autoscaling import AutoScalingGroup
+from troposphere.autoscaling import AutoScalingGroup, EBSBlockDevice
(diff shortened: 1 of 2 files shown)


hunt3ri:
I need to temporarily close this PR, while I wait for further support from AWS

PR #1957, opened 2021-09-29
fix: creating specific AWS::MediaPackage::OriginEndpoint properties

axelpavageau:
The goal of this PR is to fix a mix up between AWS::MediaPackage::OriginEndpoint & AWS::MediaPackage::PackagingConfiguration properties. 

They share the same name, but have different properties.

For instance : 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-mediapackage-originendpoint-cmafpackage.html vs https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-mediapackage-packagingconfiguration-cmafpackage.html


markpeek:
Thank you for sorting this out and adding an example.

PR #1958, opened 2021-10-05
AWS Backup: Add EnableContinuousBackup boolean to BackupRuleResourceType

tnielsen2:
Adds the boolean attribute `EnableContinuousBackup`. This was applied and tested against CF in a new rule.


markpeek:
Thanks!

PR #1960, opened 2021-10-07
AWS::OpenSearchService::Domain.EngineVersion

jerry153fish:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-opensearchservice-domain.html

### DIFF
--- a/tests/test_opensearchservice.py
+++ b/tests/test_opensearchservice.py
@@ -1,15 +1,15 @@
 import unittest
 
-import troposphere.elasticsearch as es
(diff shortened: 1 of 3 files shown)


markpeek:
Thanks for making this change but I think I'm going to split this out into a separate file. I doubt AWS will make further improvements to Elasticsearch and OpenSearchService is their replacement. Might was well get this in its own namespace sooner than later.

jerry153fish:
Got it. Do you want me to separate it? Or do you already have the codes?

PR #1962, opened 2021-10-18
Update collections

blaisethompson:


markpeek:
@blaisethompson as you may have found a fix for the Python 3.10 collection deprecation was fixed via #1961 and release in troposphere 3.1.0 over the weekend.

PR #1970, opened 2021-11-05
Added Cloudfront Response Header changes per Nov 4 updates.

glitchindustries:
Adding Changes per https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudfront-responseheaderspolicy.html

### DIFF
--- a/troposphere/cloudfront.py
+++ b/troposphere/cloudfront.py
@@ -3,7 +3,6 @@
 #
 # See LICENSE file for full license.
 
-from re import M
 from . import AWSObject, AWSProperty, Tags
 from .validators import (
     boolean,
@@ -12,7 +11,6 @@ from .validators import (
     cloudfront_cache_query_string_behavior,
     cloudfront_event_type,
     cloudfront_forward_type,
-    cloudfront_frame_option,
     cloudfront_origin_request_cookie_behavior,
     cloudfront_origin_request_header_behavior,
     cloudfront_origin_request_query_string_behavior,
--- a/troposphere/cloudfront.py
+++ b/troposphere/cloudfront.py
@@ -3,6 +3,7 @@
 #
 # See LICENSE file for full license.
 
+from re import M
 from . import AWSObject, AWSProperty, Tags
 from .validators import (
     boolean,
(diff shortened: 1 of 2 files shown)


ntwaddell (approved):

troposphere/cloudfront.py
```diff
@@ -16,6 +16,9 @@
     cloudfront_origin_request_query_string_behavior,
     cloudfront_restriction_type,
     cloudfront_viewer_protocol_policy,
+    cloudfront_access_control_allow_methods,
```
markpeek, line 19: These should be alphabetized in the list.

troposphere/cloudfront.py
```diff
@@ -401,6 +404,130 @@ class OriginRequestPolicy(AWSObject):
     }
 
 
+class AccessControlAllowHeaders(AWSProperty):
+    props = {
+        "Items": [(str, True)]
```
markpeek, line 409: The form `[(str, True)]` should be `([str], True)`. In other words, the tuple is (property_type, required_bool) and the type is an array of str `[str]`.

troposphere/cloudfront.py
```diff
@@ -401,6 +404,130 @@ class OriginRequestPolicy(AWSObject):
...
+    }
+
+
+class AccessControlAllowMethods(AWSProperty):
+    props = {
+        "Items": [(cloudfront_access_control_allow_methods, True)]
```
markpeek, line 415: Same change needed.

troposphere/cloudfront.py
```diff
@@ -401,6 +404,130 @@ class OriginRequestPolicy(AWSObject):
...
+    }
+
+
+class AccessControlAllowOrigins(AWSProperty):
+    props = {
+        "Items": [(str, True)]
```
markpeek, line 421: Same change needed.

troposphere/cloudfront.py
```diff
@@ -401,6 +404,130 @@ class OriginRequestPolicy(AWSObject):
...
+    }
+
+
+class AccessControlExposeHeaders(AWSProperty):
+    props = {
+        "Items": [(str, True)]
```
markpeek, line 427: Same change needed.

markpeek (changes requested):
Thank you for making this change. There needs to be a couple of easy tweaks done. Also, it would be good to run `make fix-black fix-isort test lint` just to verify the change before pushing it again. Thanks!

markpeek (approved):
Looks great!

ntwaddell:
Woohoo!

markpeek:
Thanks!

PR #1971, opened 2021-11-08
Add Architectures to AWS::Serverless::Function

ds-mn:
Per the [SAM Developer Guide](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html#sam-function-architectures)


markpeek:
Thanks!

PR #1979, opened 2021-11-29
Updated elasticache.py, ec2.py

ds-mn:
Added User, UserGroup to Elasticache and Ingress/Egress Properties to EC2


markpeek:
The EC2 change looks incomplete as those need to be tied into a resource or property. Also, best to split these changes into individual PRs. Thanks.

markpeek:
ElastiCache is now auto-generated and these changes are included.

PR #1984, opened 2021-12-11
NetworkFirewall - Fixed LoggingConfiguration

ntwaddell:
The logging configuration was not correct per CloudFormation documentation

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-networkfirewall-loggingconfiguration.html

### DIFF
--- a/troposphere/networkfirewall.py
+++ b/troposphere/networkfirewall.py
@@ -114,7 +114,7 @@ class FirewallPolicy(AWSObject):
     }
 
 
-class LogDestinationConfig(AWSProperty):
+class LogDestinationConfigs(AWSProperty):
     props = {
         "LogDestination": (dict, True),
         "LogDestinationType": (str, True),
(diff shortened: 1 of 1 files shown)


ntwaddell:
@markpeek

troposphere/networkfirewall.py
```diff
@@ -114,17 +114,28 @@ class FirewallPolicy(AWSObject):
     }
 
 
-class LogDestinationConfig(AWSProperty):
+class LogDestinationConfigs(AWSProperty):
```
markpeek, line 117: According to the docs, this should remain singular.

troposphere/networkfirewall.py
```diff
@@ -114,17 +114,28 @@ class FirewallPolicy(AWSObject):
...
         "LogDestinationType": (str, True),
         "LogType": (str, True),
     }
 
 
-class LoggingConfiguration(AWSProperty):
+class LoggingConfigurationProperty(AWSProperty):
+
+    props = {
+        "LogDestinationConfigs": ([LogDestinationConfigs], True)
```
markpeek, line 128: This should be use a singular inside of the list.
```
"LogDestinationConfigs": ([LogDestinationConfig], True)
```

markpeek (changes requested):
Thank you for correcting this Object/Property issue. Looking at the docs `LogDestinationConfig` should remain singular.

troposphere/networkfirewall.py
ntwaddell: Ahh, you are correct.  Updated!

markpeek (approved):
LGTM

markpeek:
Thanks!

PR #1988, opened 2021-12-16
Update Kinesis per 2021-12-09

glitchindustries:
Added Kinesis StreamModeDetails per https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-kinesis-stream-streammodedetails.html


markpeek:
Thanks!

PR #1989, opened 2021-12-17
pre-commit checks for black+isort

sirrus233:
Got bitten by this on my last PR - so figured I'd give it a little automation for the next person to come along (especially if that should happen to be me!)

Pretty simple to start, but can be expanded with additional checks down the line if desired.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,10 @@
+repos:
+- repo: https://github.com/psf/black
+  rev: '21.12b0'
+  hooks:
+    - id: black
+
+- repo: https://github.com/timothycrosley/isort
+  rev: '5.10.1'
+  hooks:
+    - id: isort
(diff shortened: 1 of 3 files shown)


sirrus233:
FWIW - pre-commit only runs against files that are changed in the commit. Since there's no source code in this PR, nothing was actually checked when I committed it. Out of curiosity, however, I ran it against the whole repo with `pre-commit run -a`. Looks like there are currently a handful of `black` violations checked in:

```
	modified:   docs/conf.py
	modified:   scripts/cfn
	modified:   scripts/cfn2py
	modified:   troposphere/kinesis.py
```
I figured the change would be easier to review if I left these alone, but if it's helpful I can add a commit with the (automated) fixes to these files.

markpeek:
Nice! Thanks! Also, I've been meaning to update the github action to catch these as well.

PR #1995, opened 2021-12-29
Adding missing package

JohnPreston:
Adding missing package for validators

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,6 +36,7 @@ packages =
     troposphere
     troposphere.openstack
     troposphere.helpers
+    troposphere.validators
 scripts =
     scripts/cfn
     scripts/cfn2py


markpeek:
Good catch. Thanks!

JohnPreston:
@markpeek 
QQ: are you planning to roll out the generated classes as part of the 3.x versions (and therefore do a slow roll out) or have that prepared for a 4.x version ?
Also wondering if there is a release strategy to releases ? i.e. patch every month / minor every quarter ? (something like that?)
Thank you :rocket:

markpeek:
One of the reasons for moving to the code generation is to make troposphere more maintainable and allow for easier releases. Right now it takes a lot of sleuthing to figure out what AWS has changed week-to-week. I have tried to be careful with backward compatibility to allow for a slow rollout via 3.x releases. For 4.x I was thinking of removing most or all the backward compatibility to align with the AWS naming conventions. I'll be pushing a few more changes and then releasing in the next couple of days. But yes, I'd like to get to a more regular cadence for the releases.

JohnPreston:
> One of the reasons for moving to the code generation is to make troposphere more maintainable and allow for easier releases. Right now it takes a lot of sleuthing to figure out what AWS has changed week-to-week. I have tried to be careful with backward compatibility to allow for a slow rollout via 3.x releases. For 4.x I was thinking of removing most or all the backward compatibility to align with the AWS naming conventions. I'll be pushing a few more changes and then releasing in the next couple of days. But yes, I'd like to get to a more regular cadence for the releases.

Excellent. In an attempt at further automation of the rendering process I had created [this](https://pypi.org/project/aws-cfn-resources-schemas/) only to realize after the facts that you use the Specifications files (:bell: //shame on me).

I might take a stab at doing something similar and see to if something similar could do the trick to render the troposphere files. The trick is in the `validators` to be imported correctly. I haven't spent enough time though with the `gen()` method.

PR #2009, opened 2022-02-25
automating maintenance with Github actions

PatMyron:
tested in my fork to successfully reproduce https://github.com/cloudtools/troposphere/commit/227aa64541b3db7500f175a0bd27ce70d47b5547: https://github.com/PatMyron/troposphere/pull/1


markpeek:
Oh cool, thanks for putting this together as I was going to look at implementing something similar over the weekend. Some comments:
- I like that it creates a PR for the changes so we can validate the changes. But given new spec files tend to come out once a week, should this run daily instead of every 5 minutes?
- I was thinking something similar for PRs (such as new validators or other tweaks) to ensure a regen will give a clean git tree (meaning there aren't manual changes being made).
- Should run `make test` and `make lint` (might need flake8 added as a dependency) to verify the code before moving on to the PR stage.

PatMyron:
> new spec files tend to come out once a week

Noting https://github.com/peter-evans/create-pull-request/ is idempotent, so it won't open multiple PRs, and it'll only update the existing PR if there are new changes. But feel free to slow the cadence if desired:
Screen Shot 2022-02-25 at 4 57 07 PM

> I was thinking something similar for PRs (such as new validators or other tweaks) to ensure a regen will give a clean git tree (meaning there aren't manual changes being made)

Good idea, should be able to re-run `make regen` and then run [`git diff --exit-code`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---exit-code). Might open a separate PR for that

> Should run `make test` and `make lint` to verify the code before moving on to the PR stage.

These currently run after the PR stage:
https://github.com/cloudtools/troposphere/blob/227aa64541b3db7500f175a0bd27ce70d47b5547/.github/workflows/tests.yml#L38-L41
Thinking it'd be useful to see the generated PR and failing tests, rather than more silently failing with no PRs being generated

markpeek:
Thanks!

PR #2011, opened 2022-02-28
Expanding one_of validator

JohnPreston:
* Fixed typo for validation function
* Validating primitive types for `one_of` as can't use ``troposphere.AWSHelperFn`` due to cyclic include.

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -1,6 +1,7 @@
 import unittest
 
 from troposphere import NoValue, Parameter, Ref, Tags
+from troposphere import AWS_REGION
 from troposphere.validators import (
     boolean,
     check_required,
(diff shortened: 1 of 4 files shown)


JohnPreston:
@markpeek To add for 4.0 ?
Would have preferred to test against `troposphere.AWSHelperFn` class but then gets cyclic import.

markpeek:
@JohnPreston the cyclic import can be worked around by importing AWSHelperFn within the function itself. There are examples in that same file.
```
def sample():
    from .. import AWSHelperFn
    pass
```

markpeek:
I think the change you might be wanting is something like below. Feel free to make the change in your code, test, and push an update.
```
def one_of(class_name, properties, property, conditionals):
    from .. import AWSHelperFn

    if isinstance(properties.get(property), AWSHelperFn):
        return

    if properties.get(property) not in conditionals:
...
```

JohnPreston:
done @markpeek

markpeek:
I pulled your changes, split the commits into two ([ecs typo](5447948786e92de312d3e8586e0139ccb6f8beda) and [validator fix](75db1661e17e7d6c36c53d269d999774a0a30e45)), tweaked slightly, and pushed manually.

Thanks for the fixes @JohnPreston!

markpeek:
Oops, let's try the link to the changes this way 5447948786e92de312d3e8586e0139ccb6f8beda and 75db1661e17e7d6c36c53d269d999774a0a30e45

JohnPreston:
Thanks @markpeek 
I will split the commits next time :+1:

Issue #2012, opened 2022-03-01
ManagedRuleGroupStatement missing ScopeDownStatement property


linasdsilva:
wafv2.py ManagedRuleGroupStatement seems to be missing the ScopeDownStatement property, i.e. as per https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-wafv2-webacl-managedrulegroupstatement.html definition.

markpeek:
I explicitly removed it from v4.0.0-beta.0 as it was not present in v3.2.2. Looking at it closer, it will likely need to follow the One/Two/Three pattern present in this file to handle the cyclic use of Statement. In other words, this pattern which will break backward compatibility for any usage of the current ManagedRuleGroupStatement:
```
class ManagedRuleGroupStatementOne(AWSProperty):
    props: PropsDictType = {
        ...
        ScopeDownStatement: (StatementTwo, False),
        ...
    }
```
You are likely asking since you need it. Be curious what the community thinks about the backward compatibility issue. At least 4.0.0-beta is intended for some breaking changes.

Note: I've also thought about removing the One/Two/Three workaround and using a validation function to prevent the cyclic behavior.

markpeek:
This was added in the above commit.

PR #2014, opened 2022-03-03
Reimplement WAFv2 Statement validation

markpeek:
A fix for #2012 would introduce another larger set of object copies to handle the ScoreDownStatement use case. Rather than perpetuate that set of changes I am proposing a breaking change to make the classes more uniform. Would be good to get feedback on this approach.

Tagging some committers/users of this service:
@linasdsilva @axelpavageau @marksidell @michael-k @nkrumm @jerry153fish 

The current WAFv2 works around a cyclic issue by using copies of objects
to allow hierarchical use. This change works around this issue by using
property validators instead. This does break backward compatibility with
the following class renames:
  - AndStatementOne, AndStatementTwo => AndStatement
  - NotStatementOne, NotStatementTwo => NotStatement
  - OrStatementOne, OrStatementTwo => OrStatement
  - RateBasedStatementOne, RateBasedStatementTwo => RateBasedStatement
  - StatementOne, StatementTwo, StatementThree => Statement


jerry153fish (approved):
LGTM.

michael-k (approved):
LGTM (but I'm not affected by this)

PR #2016, opened 2022-03-07
Do not merge - Prototype intellisense by emitting function signatures for all classes

markpeek:
Based on the auto-gen work and @ITProKyle typing changes, I put together a quick prototype to allow for some intellisense functionality. This is a WIP.

To try out:
```
- Download the branch
- Install the dev requirements
- make spec
- make regen (note: wafv2 will fail)
- Test with your favorite editor
```

Issues:
- Template has a circular dependency so tests fail. Does anyone even use that argument?
- wafv2 has a "None" property name that doesn't play well with python. Guess no one is using it anyway.
- Doesn't handle multiple validation types where a Union might be needed.
- Likely more...this was a quick test.

Might be of interest to @JohnPreston @michael-k

### DIFF
--- a/scripts/gen.py
+++ b/scripts/gen.py
@@ -410,7 +410,6 @@ class CodeGenerator:
         self.resources: Dict[str, ResourceType] = service.resources
         self.properties: Dict[str, PropertyType] = service.properties
         self.property_validators = service.property_validators
-        self.statement_found = False
 
     def generate(self, file=None) -> str:
         """Generated the troposphere source code."""
@@ -428,6 +427,8 @@ class CodeGenerator:
         if self._walk_for_tags():
             code.append("from . import Tags")
         code.append("from . import PropsDictType")
+        code.append("from . import Template")
+        code.append("from typing import Optional")
 
         if not stub:
             # Output imports for commonly used validators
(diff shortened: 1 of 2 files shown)


scripts/gen.py
```diff
@@ -676,10 +670,13 @@ def _get_type(self, value: Property, stub=False):
 
         if value.type == "List":
             if value.item_type:
-                return "[%s]" % value.item_type
+                if stub:
+                    return "'list[%s]'" % value.item_type
```
michael-k, line 674: `list[…]` instead of `typing.List[…]` is only available in Python 3.9+

michael-k (commented):
Generates eg.

```python
class Group(AWSObject):
    def __init__(
        self,
        title: Optional[str],
        template: Optional[Template] = None,
        validation: bool = True,
        *,
        GroupName: iam_group_name = None,
        ManagedPolicyArns: "list[str]" = None,
        Path: iam_path = None,
        Policies: "list[Policy]" = None,
    ):
        ...
```
but `iam_group_name` and `iam_path` aren't valid types.  It also shouldn't add `= None` for required properties.

Personally I'd prefer switching to [dataclasses](https://docs.python.org/3/library/dataclasses.html) or [pydantic](https://pydantic-docs.helpmanual.io/) instead of generating `__init__` methods.

PR #2022, opened 2022-03-24
Update serverless.py

gsoyka:
Update based on current docs:
https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-api.html
https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-requestmodel.html


markpeek:
Thank you for the serverless update.

PR #2025, opened 2022-03-31
Let type hints show that lists of resources, parameters, and outputs are also valid

michael-k:
See also https://github.com/cloudtools/troposphere/pull/2013#discussion_r838745452


markpeek:
Thanks!

PR #2035, opened 2022-04-06
Add ephemeral storage

gsoyka:
Adds support for Serverless transform EphemeralStorage option per docs: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html


markpeek:
Thanks!

PR #2069, opened 2022-08-07
An example of SNS alert for failed batch job events

rayjan0114:
Extend the official tutorial to be more close to production, the original tutorial is located at:
https://docs.aws.amazon.com/batch/latest/userguide/batch_sns_tutorial.html

This example would send a successful job every 1 minute and failed job every 3 minutes and the failed jobs would invoke the lambda function through SNS.

### DIFF
--- a/examples/BatchEventSnsLambda.py
+++ b/examples/BatchEventSnsLambda.py
@@ -6,8 +6,9 @@
 from awacs.aws import Allow, PolicyDocument, Principal, Statement
 from awacs.batch import SubmitJob
 from awacs.sns import Publish
-from troposphere import GetAtt, Join, Parameter, Ref, Region, Template
-from troposphere.awslambda import Environment, Permission, Function, Code
+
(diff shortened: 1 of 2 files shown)


markpeek:
Thank you for the PR. I already know this will fail the tests as you need to add the output into `tests/examples_output/BatchEventSnsLambda.template` to verify the output when the tests are run. Also, be sure to run `black` and `isort` on the python file. Thanks!

rayjan0114:
OK, got it, I will fix it

rayjan0114:
Thanks for the feedback @markpeek 

I have done the following checks
- [x] deployed the template and checked it as expected
- [x] passed the `python setup.py test`
- [x] `isort examples/BatchEventSnsLambda.py`
- [x] `black examples/BatchEventSnsLambda.py`

Please let me know if there is anything else, thanks

markpeek:
@rayjan0114 Thank you for the PR and making the updates.

PR #2074, opened 2022-09-12
Add a simple test for the new AWS::LanguageExtensions transform

michael-k:
See https://aws.amazon.com/about-aws/whats-new/2022/09/aws-cloudformation-new-language-extensions-transform/

I've take the example from https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-languageextension-transform.html to see if that works.

At least the combination of `AWS::LanguageExtensions` with the `AWS::Serverless` transform and globals needs a bit of work.  The globals checks don't work with multiple transforms.  https://github.com/cloudtools/troposphere/blob/759496c4d2ed3fec6fcb7315b98fa9bf7e2335b2/troposphere/__init__.py#L834-L850

### DIFF
--- a/tests/test_language_extensions.py
+++ b/tests/test_language_extensions.py
@@ -1,0 +1,68 @@
+import unittest
+
+from troposphere import AWSHelperFn, Parameter, Template
+from troposphere.sqs import Queue
+
+
+class TestServerless(unittest.TestCase):
+    def test_transform(self):
+        t = Template()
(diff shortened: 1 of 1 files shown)


michael-k:
The tests passed locally.

The failing spec test is a bit annoying.  I've looked into it, but spent most of my available time understanding the code.

markpeek:
Thanks!

PR #2078, opened 2022-09-29
me-central-1 (UAE)

PatMyron:
https://aws.amazon.com/blogs/aws/now-open-aws-region-in-the-united-arab-emirates-uae/


markpeek:
Thanks!

PatMyron:
Wonder if we want to keep maintaining these region constants manually: https://github.com/terraform-linters/tflint-ruleset-aws/pull/367#issuecomment-1270310875

There have already been a couple new AWS regions launched since this PR and expecting another region soon:

https://aws.amazon.com/blogs/aws/a-new-aws-region-opens-in-switzerland/
https://aws.amazon.com/blogs/aws/now-open-aws-region-in-spain/

---

Only able to find one public import of these region constants from searching on sourcegraph, so I'm wondering if they should be deprecated:
https://github.com/arhs/spikeseed-cloud-labs/blob/master/troposphere/getting_started/constants.py

markpeek:
I believe there is still value in having them included. Let me open an issue to look into auto-generating them to make maintaining easier.

PR #2083, opened 2022-10-18
Allow setting hosted elasticsearch volume_types to gp3

estheruary:
Referenced https://aws.amazon.com/about-aws/whats-new/2022/07/amazon-opensearch-service-support-ebs-gp3-volume-type/


markpeek:
Thanks!

PR #2089, opened 2022-10-25
Run tests against Python 3.11 and add trove classifier

michael-k:
released on 2022-10-24, see https://pythoninsider.blogspot.com/2022/10/python-3110-is-now-available.html


markpeek:
Thanks!

PR #2114, opened 2023-01-11
Be more lenient about types for Export

estheruary:
My use case is something like

```python
  t.add_output([
    Output(
      ...,
      Export=Export(Join(":", ["some", "stuff"]))
    )
  ])
```


markpeek:
Thank you for the PR!

PR #2115, opened 2023-01-17
support for FilterPolicyScope on the sns topic subscription

alapatipavan:
This is will add the new `FilterPolicyScope(string)` which can be used on the sns to filter on message_attributes or message_body.

https://docs.aws.amazon.com/sns/latest/dg/sns-message-filtering.html#sns-message-filtering-scope


markpeek:
Thank you for the PR. This change was incorporated via the spec 104.0.0 update (6cf6ed95).

PR #2124, opened 2023-02-21
Add many missing EC2, RDS and Elasticache instance type constants.

echernyavskiy:


markpeek:
Thank you for making these updates.

PR #2125, opened 2023-02-22
Fix valid values used by validators

michael-k:
Also fixed a broken `.join()` that I found while writing tests.

### DIFF
--- a/troposphere/validators/appconfig.py
+++ b/troposphere/validators/appconfig.py
@@ -9,12 +9,12 @@ def validate_growth_type(growth_type):
     Property: DeploymentStrategy.GrowthType
     """
 
-    VALID_GROWTH_TYPE = "LINEAR"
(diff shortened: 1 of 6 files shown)


troposphere/validators/appconfig.py
```diff
@@ -9,12 +9,12 @@ def validate_growth_type(growth_type):
     Property: DeploymentStrategy.GrowthType
     """
 
-    VALID_GROWTH_TYPE = "LINEAR"
+    VALID_GROWTH_TYPES = ("LINEAR",)
 
-    if growth_type not in VALID_GROWTH_TYPE:
+    if growth_type not in VALID_GROWTH_TYPES:
```
michael-k, line 14: Before `LINEA` was also accepted as valid.

markpeek:
Nice! Thank you!

PR #2126, opened 2023-02-22
Mark constants as final

michael-k:
mypy (don't know about pyright) treats `Delete = "Delete"` as `Delete: str = "Delete"` for backwards compatibility, see https://mypy.readthedocs.io/en/stable/literal_types.html#declaring-literal-variables.

`Delete: Final = "Delete"` is the same as `Delete: Literal["Delete"] = "Delete"`.


michael-k:
No worries.  I need to fix some type errors anyway.

markpeek:
Thanks!

PR #2127, opened 2023-02-23
Run spec checks in their own GitHub Actions job

michael-k:
… and polish the maintenance workflow a bit.

Imho this is better for reviewers (and PR authors):

image

Before:

image


markpeek:
This looks good to get better test results from PRs. I had been thinking to have the maintenance action to commit the spec version used. Then the test action could used the last committed spec version to run tests instead of using the "latest" spec version that usually cause errors.

PR #2137, opened 2023-03-24
update engine types validator for AWS::RDS::DBInstance

jerry153fish:
update the RDS custom types for the validator:  https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html


markpeek:
Thanks!

PR #2142, opened 2023-04-05
Remove storage size validation for gp3 RDS StorageType

dieutoutpuissant:
update validator for gp3 storage type, which isn't based on `Iops` minimum allocated storage anymore

### DIFF
--- a/troposphere/validators/rds.py
+++ b/troposphere/validators/rds.py
@@ -385,7 +385,10 @@ def validate_dbinstance(self) -> None:
 
     allocated_storage = self.properties.get("AllocatedStorage")
     iops = self.properties.get("Iops", None)
-    if iops and not isinstance(iops, AWSHelperFn):
+    if (
+        iops and not isinstance(iops, AWSHelperFn)
+        and storage_type not in ["gp3"]
+    ):
         min_storage_size = 100
         engine = self.properties.get("Engine")
         if not isinstance(engine, AWSHelperFn) and engine.startswith("sqlserver"):


markpeek:
Looking on the [RDS DBInstance](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html) page under StorageType I see:
```
StorageType
Specifies the storage type to be associated with the DB instance.

Valid values: gp2 | gp3 | io1 | standard

The standard value is also known as magnetic.

If you specify io1 or gp3, you must also include a value for the Iops parameter.

Default: io1 if the Iops parameter is specified, otherwise gp2
```

So it seems there is still a requirement to specify Iops for gp3?

dieutoutpuissant:
It is still a requirement to specify Iops for gp3, however the minimum allocated storage constraints don't apply to said storage type anymore according to the [RDS Documentation](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Storage.html#Concepts.Storage.GeneralSSD):
`By using General Purpose SSD gp3 storage volumes, you can customize storage performance independently of storage capacity.`

fpavageau:
Actually, it seems it's even more complicated than that: the same [RDS DBInstance](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html) page under `Iops` says
> **Note**
If you specify io1 for the StorageType property, then you must also specify the Iops property.

It could be seen as a oversight of the gp3 addition, but the [RDS DB instance storage](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Storage.html#Concepts.Storage.GeneralSSD) page under "gp3 storage" has "N/A" as the range of Provisioned IOPS when the storage size is less that 400 GiB for MariaDB, MySQL, and PostgreSQL (200 GiB for Oracle), so maybe it's not a hard requirement. Also, more practically, we tried to update an existing database with 100 GiB of storage from gp2 to gp3, specifying the baseline 3000 for `Iops`, and CloudFormation failed with the following error:
> Resource handler returned message: "You can't specify IOPS or storage throughput for engine postgres and a storage size less than 400. (Service: Rds, Status Code: 400, Request ID: 12345678-1234-1234-1234-123456789012)" (RequestToken: 98765432-9876-9876-9876-987654321098, HandlerErrorCode: InvalidRequest)

It could be seen as a bug in CloudFormation, which could accept the baseline as the only valid value for small storage sizes instead of refusing it altogether, especially given that the threshold is different depending on the database engine.

markpeek:
Thanks!

PR #2145, opened 2023-04-14
Add ApiFunctionAuth serverless class for Api Function events

### LINKED ISSUE #2144: Serverless Api Function events should use ApiFunctionAuth instead of Api Auth
There is currently only one `Auth` class defined in `serverless.py`. This class is used by both the `Api` class as well as `ApiEvent` class, the expected properties of these objects does not however match.

See https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-api-apiauth.html, versus https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-apifunctionauth.html.

Another class should be created to provide support for the `ApiFunctionAuth` entity.

vvangestel:
Api Function Auth follows a different structure from Api Auth and thus should not re-use the same class.
Previously used https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-api-apiauth.html versus intended https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-apifunctionauth.html.

This PR fixes this discrepancy.
Fixes #2144

### DIFF
--- a/troposphere/serverless.py
+++ b/troposphere/serverless.py
@@ -583,6 +583,16 @@ class DynamoDBEvent(AWSObject):
     }
 
 
+class ApiFunctionAuth(AWSProperty):
+    props: PropsDictType = {
+        "ApiKeyRequired": (bool, False),
+        "AuthorizationScopes": (list, False),
+        "Authorizer": (str, False),
+        "InvokeRole": (str, False),
+        "ResourcePolicy": (ResourcePolicyStatement, False),
+    }
+
+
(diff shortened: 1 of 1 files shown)


vvangestel:
Hello, can this PR receive a set of :eyes: 

Thanks :pray:

markpeek:
Thank you for the PR and sorry about the delay. I added more logic via 2c7f04f for backward compatibility, deprecation warning, and tests.

PR #2147, opened 2023-04-17
TypeError() return the expected Class as last exception argument

JohnPreston:
When the validator is used instead of the AWSProperty class (i.e. in WAFv2 for `Statement`), the TypeError() will return the expected AWSProperty SubClass as the last argument. This enables other libraries to identify the expected class and perform manipulations to import the resource property class.

For simpler class types (int, str, etc.) the validators only validate that the input is of that exact type and is simpler to catch.

Went over all the validators, these are the only two with an AWSProperty type validation, for due diligence. Although only need this today with WAFv2


markpeek:
Thank you for the PR and sorry for the delay in merging.

PR #2153, opened 2023-05-05
Remove gp3 from list of storage types that require Iops to be set for RDS

romanglovahs:
Cloudformation does not allow to set Iops for storage GP3 < 400Gib, example error:
Resource handler returned message: "You can't specify IOPS or storage throughput for engine mysql and a storage size less than 400. (Service: Rds, Status Code: 400
For <400Gib it's set automatically to 3000 and >=400 Gib is set to 12000

### DIFF
--- a/troposphere/validators/rds.py
+++ b/troposphere/validators/rds.py
@@ -383,10 +383,10 @@ def validate_dbinstance(self) -> None:
     storage_type = self.properties.get("StorageType", None)
     if (
         storage_type
-        and storage_type in ["io1", "gp3"]
+        and storage_type in ["io1"]
         and "Iops" not in self.properties
     ):
-        raise ValueError("Must specify Iops if using StorageType io1 or gp3")
+        raise ValueError("Must specify Iops if using StorageType io1")
 
     allocated_storage = self.properties.get("AllocatedStorage")
     iops = self.properties.get("Iops", None)


romanglovahs:
@markpeek Hello Mark, can you please look into this

codingtim:
We also just ran into this issue.
However the storage size limit for specifying the Iops is different per DB engine as can be seen in this table: (for example Oracle Iops cannot be specified is storage size < 200)
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Storage.html#gp3-storage
So either the validation can be removed or take DB engine and storage size into consideration.

tnielsen2:
I have a PR in here that adds additional validator logic, but this would also fix our deployment issues we are experiencing as well.

https://github.com/cloudtools/troposphere/pull/2164

tnielsen2:
Tagging Mr @markpeek , Hi Mark is this something we can look at trying to get into the next release or some variation of it?

romanglovahs:
@tnielsen2 Looks like Mark is not interested in this tool, I'd recommend to migrate to something else or fork it for full access

markpeek:
@tnielsen2 was this resolved via 0faace5?

tnielsen2:
> @tnielsen2 was this resolved via 0faace5?

Yep!

markpeek:
Excellent. Thanks!

Issue #2151, opened 2023-04-25
field 'Tags' missing from events.Rule


lreeder-arpio:
Per Boto and Cloudformation documents (https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-tagging.html, or see Boto return types for EventBus and Rule), "In EventBridge, you can assign tags to rule[s] event buses".   In Troposphere, events.EventBus has Tags, but events.Rule does not.

markpeek:
Thanks for the issue and PR.

PR #2158, opened 2023-05-26
Added tags to EventBridge (Events) Rule resource and fixed an error in CONTRIBUTING.md

lreeder-arpio:
Per https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-tagging.html, only two EventBridge resources support tags:  event buses and rules.   The Troposphere EventBus resource had a field for tags, but Rules was missing it, probably because the Cloudformation Resource Specification was missing it.  This is issue https://github.com/cloudtools/troposphere/issues/2151.

This PR adds a jsonpatch to plug Tags for the Events::Rules resource into the resource specification and adds Tags to the Rules class. Also, the example for running `scripts/gen.py` included the cloudformation spec file as an argument, which causes gen.py to treat it as a service name and build a corrupted file.  This patch removes the cloudformation spec file from the example.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,7 +45,7 @@ make spec
 To generate code, the current process is roughly, scan the CloudFormation history to identify changes and then run (using S3 as an example):
 
 ```
-python3 scripts/gen.py s3 CloudFormationResourceSpecification.json > troposphere/s3.py
+python3 scripts/gen.py s3 > troposphere/s3.py
 ```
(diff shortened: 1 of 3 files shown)


markpeek:
Thanks!

markpeek:
Needed to revert this commit since it is not implemented yet and causes rollbacks.

lreeder-arpio:
Yeah, sorry.  I missed that it wasn't supported in CloudFormation.

PR #2159, opened 2023-05-31
Adding support for default values in FindInMap

the-jules:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-findinmap-enhancements.html

### DIFF
--- a/troposphere/__init__.py
+++ b/troposphere/__init__.py
@@ -514,7 +514,7 @@ class FindInMap(AWSHelperFn):
         find_in_map_details = [self.getdata(mapname), toplevelkey, secondlevelkey]
 
         if defaultvalue is not None:
-            find_in_map_details.append(defaultvalue)
+            find_in_map_details.append({'DefaultValue': defaultvalue})
 
         self.data = {"Fn::FindInMap": find_in_map_details}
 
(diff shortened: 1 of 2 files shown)


troposphere/__init__.py
```diff
@@ -509,11 +509,14 @@ def __init__(self, data: Any) -> None:
...
-        self, mapname: object, toplevelkey: object, secondlevelkey: object
+        self, mapname: object, toplevelkey: object, secondlevelkey: object, defaultvalue: Optional[object] = None
     ) -> None:
-        self.data = {
-            "Fn::FindInMap": [self.getdata(mapname), toplevelkey, secondlevelkey]
-        }
+        find_in_map_details = [self.getdata(mapname), toplevelkey, secondlevelkey]
+
+        if defaultvalue is not None:
+            find_in_map_details.append(defaultvalue)
```
markpeek, line 517: Thank you for this change. If I'm reading the documentation correctly, shouldn't this be:
```
    find_in_map_details.append({'DefaultValue': defaultvalue})
```
And corresponding changes in the tests.

the-jules: My original thought was that the fourth item is essentially a dict type which would mean inputting "{'DefaultValue': defaultvalue}" in the template itself, but I suppose this is cleaner, I updated the PR.

markpeek:
Thanks! Merged locally to fix formatting via 7533fbabcf3e51b0ad5785884c2d5508fe301414

PR #2164, opened 2023-07-27
Update rds validator logic

tnielsen2:
Troposphere has a new validator released in 4.3.1 that does not allow omitting the `Iops` attribute when gp3 and small StorageType parameters are passed. 

If we passed `Iops` to an instance under 400GB, with EngineType = mysql, postgres, mariadb; deploying CFN will provide the following deployment error:

```
2023-07-27T13:15:27.261000+00:00 CREATE_FAILED (AWS::RDS::DBInstance) testrds Resource handler returned message: "You can't specify IOPS or storage throughput for engine mysql and a storage size less than 400. (Service: Rds, Status Code: 400, Request ID: omitted)" (RequestToken: omitted, HandlerErrorCode: InvalidRequest)
2023-07-27T13:15:28.148000+00:00 CREATE_FAILED (AWS::IAM::Role) role Resource creation cancelled
2023-07-27T13:15:28.592000+00:00 ROLLBACK_IN_PROGRESS (AWS::CloudFormation::Stack) dev-omitted-rds The following resource(s) failed to create: [role, testrds]. Rollback requested by user.
```

Documentation [here](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Storage.html#gp3-storage) outlines the possible combinations of parameters related to GP3, EngineType, and conditional values/requirements. 

image

Reviewing [CloudFormation documentation here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html#cfn-rds-dbinstance-iops) states that IOPS is only required for `io1`.

> Note
If you specify io1 for the StorageType property, then you must also specify the Iops property.

This PR updates validation logic to allow small storage RDS instances backed by GP3 to be launched with a valid configuration values as per the above linked documentation. 

Edit post submittal:

Summary of refspec ride-along changes/fixes:
- update gen.py not to suffer recursion of EvaluationForm resource infinintely. Used same method as outlined in `scripts/gen.py` for wafv2
- Patch connect resources to change type to fix out of order circular dependency issues
- Migrate FSX fix patch from s3 file to FSX file
- Update S3 patching to accept `Encryption` Type
- Update Sagemaker patches that were removed in latest refspec
- Remove invalid S3 transfer ItemType. 
- Make refspec, and regen

### DIFF
--- a/troposphere/validators/rds.py
+++ b/troposphere/validators/rds.py
@@ -442,8 +442,9 @@ def validate_dbinstance(self) -> None:
             min_storage_size = 20
             min_iops_to_allocated_storage_ratio = 0.5
         else:
-            raise ValueError(f"StorageType gp3 is not supported for engine "
-                             f"{engine}")
+            raise ValueError(
+                f"StorageType gp3 is not supported for engine " f"{engine}"
+            )
         # Validate storage size constraints when under certain sizes
         if (
             not isinstance(allocated_storage, AWSHelperFn)
@@ -456,10 +457,13 @@ def validate_dbinstance(self) -> None:
 
         # Validate storage performance constraints for iops ratios
         if not (
-                not isinstance(allocated_storage, AWSHelperFn)
-                and not isinstance(iops, AWSHelperFn)
-                and (
-                    min_iops_to_allocated_storage_ratio < iops / int(allocated_storage) < max_iops_to_allocated_storage_ratio)
+            not isinstance(allocated_storage, AWSHelperFn)
+            and not isinstance(iops, AWSHelperFn)
+            and (
+                min_iops_to_allocated_storage_ratio
+                < iops / int(allocated_storage)
+                < max_iops_to_allocated_storage_ratio
+            )
         ):
             raise ValueError(
                 "Invalid ratio of Iops to AllocatedStorage. Minimum ratio for "
--- a/troposphere/validators/rds.py
+++ b/troposphere/validators/rds.py
@@ -368,8 +368,10 @@ def validate_dbinstance(self) -> None:
         )
 
     nonetype = type(None)
+    allocated_storage = self.properties.get("AllocatedStorage")
     avail_zone = self.properties.get("AvailabilityZone", None)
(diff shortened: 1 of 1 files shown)


tnielsen2:
As part of this PR, I am attempting to update the code to the latest ref spec, but I am having difficulty getting all CI to pass.

When regen generates the spec, it will sometimes generate the class objects out of order within the Python module. 

[Example](https://github.com/tnielsen2/troposphere/blob/13441c0e11d8f961c0bc9ec7c080ec93d7915229/troposphere/s3.py#L992)

When I fix those errors, the regen portion of the PR fails, but I can get lint to pass. 

The main goal of this PR is to update RDS GP3 validator logic, while attempting to update refspec so CI will pass, but I have been unsuccessful. I seem to be able to only get Lint or Refspec to pass. 

@markpeek What is the course for action for this PR? How important is it to have both CI and Refspec generation pass? I looked into patch scripts to address this issue, but I couldn't seem to figure it out on own. Any insight is appreciated.

tnielsen2:
With the latest commit, at this moment, lint is failing due to class objects not being generated in the correct order, as well as what seems to be circular dependencies in the `connect` module with the introduction of new AWS Properties and nested properties.

What ist he best way for this to be fixed? And if it is a patch script, what operations would I be using to fix these?

tnielsen2:
I have applied the patch changes to get this to pass lint and refspec generation. I have updated the PR with the necessary ride-along refspec updates.

markpeek:
@tnielsen2 thank you for the PR. Some of the change are needed but there is still an issue with regenerating from the spec. I  looked at it this weekend and I think I have a change that should allow this to work better. I'm going to create a branch and start applying the older spec versions into it so we can better reason about the changes. I'll ping back here when I've pushed the new branch out for your review. Thank you for your patience.

tnielsen2:
I have rebased from `main` with upstream changes and this is ready for review after the latest round of discussions/conversation @markpeek 

Thanks for all you do!

markpeek:
Thank you for the PR and getting it rebased.

PR #2166, opened 2023-08-06
Adding Runtime option in JobCommand for updated CloudFormation config

RandomForestGump:
Link:
https://docs.aws.amazon.com/glue/latest/webapi/API_JobCommand.html

Currently Ray runtime option specification is not possible in Troposphere and adding the option can enable the same.


markpeek:
Thank you for the PR. This is an auto-generated file and a recent update from the spec files has included this change.

PR #2171, opened 2023-08-14
Fix RDS Validations

JohnPreston:
The new dbinstance validations do not check properties types resulting in testing a string `oracle` against `Ref` for example.
A lot of the tests should be ignored as soon as the DBClusterIdentifier is set, indicating that this is going to be an instance attached to Aurora DB Cluster.

Exception caught

```
        File "/home/john/.cache/pypoetry/virtualenvs/ecs-composex-6XqxN9Zj-py3.9/lib/python3.9/site-packages/troposphere/validators/rds.py", line 401, in validate_dbinstance
          and "oracle" in engine                                                                          
      TypeError: argument of type 'Ref' is not iterable      
```

### DIFF
--- a/troposphere/validators/rds.py
+++ b/troposphere/validators/rds.py
@@ -367,6 +367,7 @@ def validate_dbinstance(self) -> None:
         )
 
     nonetype = type(None)
+    dbclusteridentifier = self.properties.get("DBClusterIdentifier", None)
     allocated_storage = self.properties.get("AllocatedStorage")
     avail_zone = self.properties.get("AvailabilityZone", None)
     multi_az = self.properties.get("MultiAZ", None)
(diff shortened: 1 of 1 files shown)


markpeek:
@tnielsen2 could you take a look at this PR and approve given your recent PR?
@JohnPreston is there a test that can be added to catch any regressions based on this change?

JohnPreston:
> @tnielsen2 could you take a look at this PR and approve given your recent PR? @JohnPreston is there a test that can be added to catch any regressions based on this change?

Hey @markpeek 
I will have a look and come up with something. I have my project which I created a branch for, to update troposphere version specifically, and it runs lots of tests, the only ones that failed were for RDS, hence how I found this.

I am guessing the validation test should cater for AWS HelperFn

tnielsen2 (approved):

tnielsen2:
This looks good to me. We don't run tests against validator logic, but the thought did cross my mind that having some in place for this would be great since there is a matrix of acceptable combinations with RDS specifically.

markpeek:
Published via [Release 4.4.1](https://github.com/cloudtools/troposphere/releases/tag/4.4.1)

PR #2182, opened 2023-09-18
add support for object comparison

### LINKED ISSUE #1821: add comparison capability to CloudFormation intrinsic function objects
When writing tests for Troposphere templates, I often check that resource properties contain the appropriate value references. Currently, only the `Ref` object has comparison capabilities (`__eq__` only) requiring me to compare attributes/method outputs instead of comparing the objects directly. This functionality should be extended to the remainder of the intrinsic functions.

From briefly looking through all of the subclasses of `AWSHelperFn` , it appears this could be implemented on the base class as all objects will have a `data` attribute that could be assessed or the `.to_dict()` method could be used to compare to raw dictionaries.

ITProKyle:
# Summary

Adds support to compare objects.

# Why This Is Needed

An advantage to using troposphere is being able to unit test various iterations of a dynamic template. However, the lack of support for being able to compare objects with one another makes these test cumbersome to write.

Resolves #1821

# What Changed

## Added

- added `.to_json()` method to base classes
- added `.__eq__()` and `.__ne__()` methods to base classes
  - these use the newly added `.to_json()` method for comparison of same objects created from the same class to ensure properties are sorted in the same order
  - comparing against a dict (or list for `AWSHelperFn` to support `Tags`) uses `.to_dict()` instead of converting both to a JSON string
- added `AWSHelperFn.__hash__()`
  - `tests/test_tags.py:TestTags.test_Unsortable` was failing with `TypeError: unhashable type: 'Sub'` without it

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -5,6 +5,7 @@ from troposphere import (
     AWSObject,
     AWSProperty,
     Cidr,
+    GenericHelperFn,
     If,
     Join,
     NoValue,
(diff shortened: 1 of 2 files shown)


markpeek:
Thank you for adding this functionality.

PR #2190, opened 2023-10-05
Drop support for Python 3.7

michael-k:
Python 3.7 reached its EOL on 2023-06-27, see https://peps.python.org/pep-0537/#lifespan

Python 3.7 was removed from the test matrix in commit bf8730b9a4307ee3e75f05d95c2ccdc23078a6ba.


markpeek:
Thanks!

PR #2191, opened 2023-10-05
Fix Parameters.validate incorrect validate for CommaDelimitedList

### LINKED ISSUE #2180: Incorrect ValueError: AllowedPattern can only be used with parameters of the String type
Example:

```python
    cft.add_parameter(
        Parameter(
            "Foo",
            Type="CommaDelimitedList",
            AllowedPattern="^[A-Z]{2}$",
            Default="",
        )
    )
```

According to the [AWS docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html) this should be allowed:
> AllowedPattern
> A regular expression that represents the patterns to allow for String or CommaDelimitedList types. When applied on a parameter of type String, the pattern must match the entire parameter value provided. When applied to a parameter of type CommaDelimitedList, the pattern must match each value in the list.

jerry153fish:
fix #2180

### DIFF
--- a/tests/test_parameters.py
+++ b/tests/test_parameters.py
@@ -17,5 +17,33 @@ class TestInitArguments(unittest.TestCase):
         self.assertDictEqual(reference.data, {"Ref": "title"})
 
 
+class TestParameterValidator(unittest.TestCase):
+    def test_allowed_pattern_for_number(self):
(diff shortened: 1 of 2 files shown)


markpeek:
Thanks!

PR #2194, opened 2023-10-18
Add support for ManageMasterUserPassword

eofs:
Alternatively, allow the use of `ManageMasterUserPassword` instead of `MasterUserPassword`. Both properties are mutually exclusive and cannot be used simultaneously.

Updated tests to validate `DBInstance` configuration


markpeek:
Thanks! Merged locally to squash and fix formatting via 7090af94180dcbd75b67c9bc84c878339e84ef85

PR #2197, opened 2023-10-25
Fix __eq__ protocol

ndparker:
Hello,

I'm maintaining a deployment tool based on troposphere and the latest update (4.5.0) failed one of my tests. This PR is supposed to fix the issue.

If it is not known how to compare objects, `__eq__` should return `NotImplemented` instead of `False`. That way the right side object of the comparison might possibly take over.
In my specific case the other object is a mock helper, which happily would do the comparison instead (and did until now).

Thank you.

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -2,6 +2,7 @@ import pickle
 import unittest
 
 from troposphere import (
+    AWSHelperFn,
     AWSObject,
     AWSProperty,
     Cidr,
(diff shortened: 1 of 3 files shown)


markpeek:
@ITProKyle could you review and comment/approve based on your previous change?

ITProKyle:
Sounds reasonable enough to me. Given that no tests were modified, the functionality I implement should still function as intended. @ndparker, mind adding tests to cover this use case as well to ensure it continues to function the same for you?

ndparker:
Sure, there you go.

I'm a bit unhappy about the duplication of the helper class. But I couldn't find a shared place in the tests directory.

markpeek:
Thank you for the fix and for adding the tests.

ndparker:
Thanks for merging. And sorry for not running the formatter before...

Issue #2198, opened 2023-10-25
Elasticsearch instance type names have changed


NeilJed:
AWS Comment here: https://docs.aws.amazon.com/opensearch-service/latest/developerguide/rename.html#rename-instances

> Instance types in Amazon OpenSearch Service are now in the format `<type>.<size>.search`—for example, `m6g.large.search` rather than `m6g.large.elasticsearch`. You don't need to take any action. Existing domains will start automatically referring to the new instance types within the API and in the Billing and Cost Management console.

The list of new instance type names are here: https://docs.aws.amazon.com/opensearch-service/latest/developerguide/supported-instance-types.html

While the old names point to the new, would it be more "correct" to use the new names? Maybe in the first instance we can just add the new names to the list in the constants?

markpeek:
@NeilJed can you comment/approve PR #2199? I did shorten the names by using a prefix of "SEARCH_*" instead of "OPENSEARCH_*".

I pulled the instance names from the aws cli so hopefully those are all correct.

markpeek:
@NeilJed thank you for reviewing. I did make the sort order a little nicer and pushed it out in [Release 4.5.1](https://github.com/cloudtools/troposphere/releases/tag/4.5.1)

PR #2199, opened 2023-10-25
Add new OpenSearch instance names to constants.py (Fixes: #2198)

markpeek:


NeilJed (approved):
LGTM. List from the API call matches the list published in the documentation 👍🏻

Issue #2196, opened 2023-10-25
TypeError: unhashable type: 'Parameter'


JohnPreston:
Hello,

This might be something that only affects me, but I have been relying for 3 years now and built my application to rely on the Parameters as keys to `dict`. In the 4.5.0, I am getting the error

`TypeError: unhashable type: 'Parameter'` and basically now, the entire thing is broken.
Seems to me the change is with, but I am not entirely sure.

```
    def __hash__(self) -> int:
        return hash(self.to_json(indent=0))
```

What to do now?

EDIT:

Here is an example among many of how I am able to use it in dict: https://github.com/compose-x/ecs_composex/blob/main/ecs_composex/dynamodb/dynamodb_stack.py#L38

markpeek:
@JohnPreston sorry to hear the latest release caused issues with your application. The likely PR is #2182 but would like to find out some more information.
- What version of Python are you using?
- Assuming it works fine with troposphere 4.4.1?
- Do you have a small test case that would exhibit this issue so it can be easily reproduced to look for a fix?

Adding @ITProKyle for any insights based on the references PR.

markpeek:
Looking at the above code, I believe this is a simple repro that works on 4.4.1 but causes the error on 4.5.0:
```
$ cat param_hash.py
from troposphere import Parameter

param = Parameter("param")
d = {
    param: "foo",
}
$ python param_hash.py
Traceback (most recent call last):
  File "/troposphere/param_hash.py", line 4, in <module>
    d = {
        ^
TypeError: unhashable type: 'Parameter'
```

ITProKyle:
So, it's not caused by my additions of `__hash__` because none of those are for ancestors of `Parameter`. It's actually caused by the addition of `__eq__` without also adding `__hash__`.

This is going to be a little tricky. Starting with [how python defines something as being **hashable**](https://docs.python.org/3.12/glossary.html#term-hashable):

> An object is hashable if it has a hash value which never changes during its lifetime (it needs a `__hash__()` method), and can be compared to other objects (it needs an `__eq__()` method). Hashable objects which compare equal must have the same hash value.
> 
> Hashability makes an object usable as a dictionary key and a set member, because these data structures use the hash value internally.
> 
[...]

PR #2200, opened 2023-10-25
Object equality fixes

markpeek:
- Add __hash__ to the Parameter class
- Disable validation when doing BaseAWSObject object comparison __eq__


JohnPreston:
That seems to address the issues with #2196 @markpeek :+1:

PR #2225, opened 2024-02-13
Fix RDS validator for gp3 iops/allocated_storage ratio check

### LINKED ISSUE #2176: AllocatedStorage can't be a Ref
If I try to use a Parameter for setting the AllocatedStorage of an RDS database I get the following error:
```
  File "/home/cup/.local/share/virtualenvs/system-oD26KLrH/lib/python3.11/site-packages/troposphere/validators/rds.py", line 402, in validate_dbinstance
    and int(allocated_storage) < 400
        ^^^^^^^^^^^^^^^^^^^^^^
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'Ref'
```

Version: 4.4.1

markpeek:
Fixes #2176
The [test case](https://github.com/cloudtools/troposphere/issues/2176#issuecomment-1940482962) in the Issue was [falling through](https://github.com/cloudtools/troposphere/issues/2176#issuecomment-1938252769) to the body of the if statement which was an indication the test was incorrect.

@tnielsen2 @JohnPreston @uldall looking for a quick code review if you have time to check it

### DIFF
--- a/tests/test_rds.py
+++ b/tests/test_rds.py
@@ -237,6 +237,41 @@ class TestRDS(unittest.TestCase):
         )
         i.validate()
 
+    def test_gp3_storage_performance(self):
+        # Using a Ref() will disable the gp3 ratio check
+        db_az = "us-east-1"
+        allocated_storage = Parameter("allocatedstorage", Type="int")
+        i = rds.DBInstance(
+            "Database1",
+            Engine="postgres",
+            EngineVersion="15.3",
+            StorageType="gp3",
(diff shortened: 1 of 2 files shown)


tnielsen2 (approved):
Yes, I can see that this should fix the logic encountered with the Parameter based on my understanding of the first `not` on line 496 reversing the logic of the condition evaluation.

markpeek:
Thank you @tnielsen2 for the code review.

PR #2247, opened 2024-05-28
Update glue.py with SNOWFLAKE validate type

evilensky:
The SNOWFLAKE validation type allows an AWS::Glue::Connection with the ConnectionType of `SNOWFLAKE` to be created.

This is not yet in any documentation nor is the underlying API documented :( 

https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1810

### DIFF
--- a/troposphere/validators/glue.py
+++ b/troposphere/validators/glue.py
@@ -26,6 +26,7 @@ def connection_type_validator(type):
         "MONGODB",
         "NETWORK",
         "SFTP",
+        "SNOWFLAKE",
     ]
     if type not in valid_types:
         raise ValueError("% is not a valid value for ConnectionType" % type)


ewascent:
I think you need to add a target class in [glue.py](https://github.com/cloudtools/troposphere/blob/46d9031b87ddf88bae5c2f4aede4db1bc31f0c09/troposphere/glue.py#L222) (e.g. JDBCTarget or MongoDBTarget) and add that to the list of targets in the [Targets class](https://github.com/cloudtools/troposphere/blob/46d9031b87ddf88bae5c2f4aede4db1bc31f0c09/troposphere/glue.py#L271) dict. 

For this connection I do not think this needs a default port in constants.py. Snowflake has a default port of 443. We both have a default https_port value, and the default of the Snowflake connection properties assume the port is 443 [here](https://github.com/snowflakedb/snowflake-connector-python/blob/5e61c94691c72e6e7e7a7c83849949a188b72172/src/snowflake/connector/connection.py#L1169)

markpeek:
Thank you for the PR. As @ewascent mentions, there are likely more changes needed to fully support Snowflake. With this change were you able to bring up a valid CloudFormation stack with Snowflake?

evilensky:
Thank you for considering the PR.

We aren't currently using [Crawler](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-glue-crawler.html) or [Target](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-glue-crawler-targets.html) so I'm unable to verify at this time. 

and just like the SNOWFLAKE connection type, I cannot find documentation for SNOWFLAKE support in these objects either in Cloudformation or [underlying API documentation](https://docs.aws.amazon.com/glue/latest/dg/aws-glue-api-crawler-crawling.html#aws-glue-api-crawler-crawling-CreateCrawler), only this time there is no [Roadmap card](https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/projects/1?card_filter_query=snowflake#card-90723483) for them.

Right now I'm monkey patching the Validator with this change, and yes, it creates a snowflake `Connection` successfully.

image

There are some interesting internal validators in the Glue API: 
```
SparkProperties.sfUrl: does not match the regex pattern .+[.]snowflakecomputing[.]com (Service: AWSGlue; Status Code: 400; Error Code: InvalidInputException; Request ID: 79f731f2-89f7-4f04-b32f-9eda6b6f3e02; Proxy: null)
```

I'm curious if there is a pattern for troposphere to validate this since the API type if `String` but the string must be of a Json of a particular shape.

markpeek:
I'm going to pull in this change since it doesn't really hurt and your screenshot shows it seems to work.

To your validation question, for validation of interconnected properties you would want to create a "class validator" for the object in validators/glue.py. I'll highlight this easy example of the [EC2 Subnet class validator](https://github.com/cloudtools/troposphere/blob/main/troposphere/validators/ec2.py#L326) here that ensures Ipv6CidrBlock is set if AssignIpv6AddressOnCreation is used in the object.

PR #2248, opened 2024-05-29
Add SnapStart to Serverless Function

Trollderim:
Allowed according to Documentation: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html


markpeek:
Thanks!

PR #2270, opened 2024-09-28
Validate the type for Tags

### LINKED ISSUE #2268: Tags types seems a bit confusing for newcomers
Hello,

Looking at the current [CloudFormation resource provider schemas](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html), tags properties are mainly expressed in 2 different types (`array` or `object`):

AWS::EC2::Instance:
```json
"Tags" : {
  "uniqueItems" : false,
  "description" : "The tags to add to the instance.",
  "insertionOrder" : false,
  "type" : "array",
  "items" : {
    "$ref" : "#/definitions/Tag"
  }
},
```

AWS::Batch::JobDefinition:
```json
"Tags" : {
  "type" : "object"
}
```

The situation is quite confusing for a newcomer, in particular regarding the `Tags` class provided by troposphere.
This class seems to generate the `array` type of tags.

At first I used an instance of the `Tags` class to add tags to a JobDefinition which was bound to fail, is there any way to ensure that troposphere will raise an error in this case?

Are there any best practices or recommendations regarding the usage of the `Tags` class?
Is using a Python dictionary to express the current `object` type of tags the way to go (for instance for a JobDefinition)?

markpeek:
Prior to this change Tags (which is a subclass of AWSHelperFn) would not do validation of the expected type and it prevented validation of Json style Tags needing a dict.

Fixes #2268

### DIFF
--- a/tests/test_serverless.py
+++ b/tests/test_serverless.py
@@ -1,6 +1,6 @@
 import unittest
 
-from troposphere import ImportValue, Parameter, Ref, Sub, Tags, Template
+from troposphere import ImportValue, Parameter, Ref, Sub, Template
 from troposphere.s3 import Filter, Rules, S3Key
 from troposphere.serverless import (
     SERVERLESS_TRANSFORM,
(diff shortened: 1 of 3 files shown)


gautaz (approved):
I have tested the branch with a JobDefinition (my original test case) with both `dict` and `Tags`, everything seems fine and working.
The test passes with `dict` and fails on `Tags` with error `JobDefinition.Tags is <class 'troposphere.Tags'>, expected <class 'dict'>` which is exactly what I expected.
Thank you!

PR #2271, opened 2024-10-02
Allow AWSHelperFn for Tags properties

benbridts:
Instead of checking the property name, we check the type of the value. This means that the Tags class gets checked, but other helper functions are allowed

We were running into this issue, because we have a custom resource that outputs a tag list.

I've also added tests for this case

### DIFF
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -1,6 +1,6 @@
 import unittest
 
-from troposphere import If, Sub, Tag, Tags
+from troposphere import GetAtt, If, Sub, Tag, Tags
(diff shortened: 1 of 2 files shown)


jvdmr (approved):

markpeek:
@benbridts apologies for the regression and thank you for the PR. I will create a new release when I get time this morning.

markpeek:
Released as [4.8.3.](https://github.com/cloudtools/troposphere/releases/tag/4.8.3)

benbridts:
Thanks!

Issue #2279, opened 2025-01-16
EC2 Route error when targeting prefix list ID


gph2584:
Troposphere throws the following error when adding an ec2.Route resource: "[ERROR] Route: one of the following must be specified: DestinationCidrBlock, DestinationIpv6CidrBlock" .
However, cloudformation now supports prefix lists as targets for routes: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-route.html "You must specify either a destination CIDR block or prefix list ID. You must also specify exactly one of the resources as the target"

markpeek:
Completely untested but see if this works:

```
% git diff
diff --git a/troposphere/validators/ec2.py b/troposphere/validators/ec2.py
index 1c9977ec..74a4bdea 100644
--- a/troposphere/validators/ec2.py
+++ b/troposphere/validators/ec2.py
@@ -268,6 +268,7 @@ def validate_route(self):
     cidr_conds = [
         "DestinationCidrBlock",
         "DestinationIpv6CidrBlock",
+        "DestinationPrefixListId",
     ]
     gateway_conds = [
         "CarrierGatewayId",
```

gph2584:
Thanks @markpeek I can confirm this works.

gph2584:
It appears I do not have permission to branch and submit a pull request for this.

markpeek:
The model is fork/PR for changes. But since I have it in my tree already I'll just commit directly.

gph2584:
I figured it out and just opened this: https://github.com/cloudtools/troposphere/pull/2280
Thanks!

markpeek:
Thank you for the issue and PR.

PR #2280, opened 2025-01-16
#2279 Validate prefix list IDs as route destinations

gph2584:


markpeek:
The pre-check tests are showing the need for a trailing comma on that new line. Can you push that change and then I can get it merged.

markpeek:
Thanks! Fixes #2279

PR #2282, opened 2025-01-23
Validate missing AWS Org policy types

alextodicescu:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-organizations-policy.html#cfn-organizations-policy-type


markpeek:
Thanks!

PR #2284, opened 2025-01-30
Updates to 189.0.0 and handles EC2 port range duplicate

wuiscmc:
The latest spec introduced a duplicate: 
[AWS::EC2::VerifiedAccessEndpoint PortRange](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-verifiedaccessendpoint-portrange.html) that conflicts with the existing property [AWS::EC2::NetworkAclEntry PortRange](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-networkaclentry-portrange.html). 

This PR handles that issue.


wuiscmc:
@markpeek exactly, I tried to fix the PortRange duplicate but I haven't mastered the patching system yet. I learnt quite a bit with your solution, thanks!

PR #2288, opened 2025-02-12
Add "Policy-Min-TLS-1-2-PFS-2023-10" as valid value TlsSecurityPolicy

sir-sigurd:
it's not listed in [CloudFormation doc ](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticsearch-domain-domainendpointoptions.html#cfn-elasticsearch-domain-domainendpointoptions-tlssecuritypolicy), but it's mentioned [here](https://docs.aws.amazon.com/securityhub/latest/userguide/es-controls.html#es-8) and I tested that it actually works


markpeek:
Thanks!

PR #2296, opened 2025-03-24
[Cloudwatch Logs] Include new retention-in-days values

amodsachintha:
AWS has added additional values for retention. Currently, troposphere raises a ValueError.

https://docs.aws.amazon.com/cli/latest/reference/logs/put-retention-policy.html#options


troposphere/constants.py
```diff
@@ -1358,8 +1358,13 @@
...
     545,
     731,
+    1096,
     1827,
-    3653,
+    2192,
+    2557,
+    2922,
+    3288,
+    3653
```
markpeek, line 1367: Minor change, can you put a comma at the end of the last entry?

markpeek:
Thanks!

PR #2304, opened 2025-05-08
Update AWS docs urls

### LINKED ISSUE #2303: AWS docs URLs need to be updated
Seems very recent, as earlier today I was still looking at the CloudFormation docs at the old URLs, which now all redirect to the new CloudFormation Template Reference Guide main page.

Try for yourself: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-servicecatalog-cloudformationprovisionedproduct.html now redirects to https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/introduction.html

The page I was looking for can now be found at https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-servicecatalog-cloudformationprovisionedproduct.html

So all that really needs to be done is replace `UserGuide` with `TemplateReference` everywhere.

jvdmr:
Fixes #2303


markpeek:
Thank you for the PR. As the documentation comes from the spec file, I need to wait for this to be fixed upstream by AWS first.

PR #2346, opened 2025-12-03
Fix regression in RDS validator

markpeek:
@yamjoepobuda I think this is more explicit to fix the regression. Can you verify and approve?

### DIFF
--- a/tests/test_rds.py
+++ b/tests/test_rds.py
@@ -271,6 +271,45 @@ class TestRDS(unittest.TestCase):
         ):
             i.to_dict()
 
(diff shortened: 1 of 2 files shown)


yamjoepobuda (approved):
Change is good as-is to resolve the reported issue, but `oracle` may have regressed as well.

markpeek:
Thank you for the review. Good catch on the oracle change. I pushed a new version and added tests as well for both of them. Can you take a look before a merge?

yamjoepobuda (approved):
lgtm, thank you for looking into this quickly!

Issue #2361, opened 2026-06-30
Add support for CpuOptions.NestedVirtualization


mcrozes:
Amazon EC2 Nested Virtualization was announced in February: https://aws.amazon.com/about-aws/whats-new/2026/02/amazon-ec2-nested-virtualization-on-virtual/. 

However, support for the `NestedVirtualization` field in `CpuOptions` property was only added to CloudFormation in early June: https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-ec2-launchtemplate-cpuoptions.html.

I already have a local implementation working and am planning to submit a PR later this week.

mcrozes:
PR merged. Closing issue.

PR #2362, opened 2026-06-30
Add NestedVirtualization on CpuOptions

mcrozes:
**Reference**: https://github.com/cloudtools/troposphere/issues/2361

Note: `CpuOptions` for `AWS::EC2::LaunchTemplate` (https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-ec2-launchtemplate-cpuoptions.html) differs from `CpuOptions` for `AWS::EC2::Instance` (https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-ec2-instance-cpuoptions.html).

Currently, there is only a single `CpuOptions` class defined in `ec2.py`.

- https://github.com/cloudtools/troposphere/blob/main/troposphere/ec2.py#L1177-L1186 defines the current `CpuOptions` property. This implementation follows the `AWS::EC2::LaunchTemplate` schema, as indicated by the presence of the `AmdSevSnp` attribute.

- Both the `AWS::EC2::LaunchTemplate` reference (https://github.com/cloudtools/troposphere/blob/main/troposphere/ec2.py#L1840) and the `AWS::EC2::Instance` reference (https://github.com/cloudtools/troposphere/blob/main/troposphere/ec2.py#L1387) point to this same shared `CpuOptions` class.

Long term, it may make sense to introduce separate `CpuOptions` classes for `AWS::EC2::LaunchTemplate` and `AWS::EC2::Instance`, since their CloudFormation schemas are not identical. 
For now, this PR extends the existing LaunchTemplate-style implementation by adding support for the new `NestedVirtualization` attribute.


mcrozes:
From what I can see, it looks like the specgen output needs to be updated (this is the result of running `make regen` locally):

```text
All done! ✨ 🍰 ✨
66 files reformatted, 522 files left unchanged.
```

(which is also what I am seeing on the failing checks above)

Based on https://github.com/cloudtools/troposphere/pull/2359, this appears to be handled automatically. However, it looks like the GitHub Actions **Automated Maintenance** workflow has been failing since May 28, which might explain why these updates haven't been generated.

Not sure if I am missing anything on my side @markpeek or if it's an error at the repo level?

markpeek:
@mcrozes thank you for the gentle prod. I've been busy on other things and meaning to fix the code generator to better handle dependencies in remaining properties. This has now been fixed. It looks like the changes for 254.0.0 picked up the NestedVirtualization property in CpuOptions.

You should be able to add your troposphere/validators/ec2.py file, run `make regen` to regenerate troposphere/ec2.py, and then run `make test lint` to ensure it passes. Then you should be able to commit the two files and push again.

There has been a long standing issue of resources for the same service reusing the same property names but with differing properties. I've had to fix it in the past with careful renaming via patches. I've had thoughts about emitting resources into separate files which would make the importing more explicit. But it would be a big change (troposphere 5.0.0) requiring rewrite of all troposphere scripts so I haven't pursued it.

For example, instead of:
```
from troposphere.ec2 import CpuOptions
```

it would be:
```
from troposphere.ec2.instance import CpuOptions
```

mcrozes:
Yep, `254.0.0` includes the changes (cool script, by the way!).

I'll leave it up to you whether you want to merge the new revision. It simply checks whether the `NestedVirtualization` value is `enabled` or `disabled`, following the AWS documentation: https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-ec2-launchtemplate-cpuoptions.html.

I'll keep an eye on your next release cycle to see if we can include it in our project. We're a little behind on versions at the moment (https://github.com/awslabs/engineering-development-hub/blob/main/source/soca/cluster_node_bootstrap/templates/linux/soca_python_controller_requirements.txt.j2#L61), so there's no rush, we have a workaround in place for now.

markpeek:
Thanks!

PR #2365, opened 2026-09-17
Support EventBridgeRuleEvent input transformer

reddymadhira111:
Details:
 AWS Does support this in their clouformation document, want to use this feature
 https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-eventbridgerule.html#sam-function-eventbridgerule-inputtransformer


markpeek:
Thanks!

